PR #6, opened 2015-07-22
Changes for enabling SGE execution

pgm:
Here are my changes for using SGE.   (Enabled when you run "wdl run --sge file.wdl")   Feel free to take them if they're useful for anyone. 

I only made sure some simple test workflows ran successfully.  I haven't run anything large so I wouldn't be surprised to discover it may need changes to make it more robust when working with large workflows.  Conceptually, I think it should work though.

At the same time, no pressure to take these changes.  You can always wait until I have had more real-life experience using them and they might mature further later.

### DIFF
--- a/python/examples/1.wdl
+++ b/python/examples/1.wdl
@@ -14,7 +14,7 @@ task grep {
 
 task wc {
   command {
-    wc -l ${sep=' ' File files+} | tail -1 | cut -d' ' -f 2
+    wc -l ${sep=' ' File files+} | tail -1 | tr -s ' ' | cut -d' ' -f 2
   }
   output {
     Int count = read_int(stdout())
--- a/python/wdl/engine.py
+++ b/python/wdl/engine.py
@@ -149,11 +149,57 @@ def engine_functions(execution_context=None):
 
     return get_function
 
+class LocalRunner:
+    def __init__(self):
+        self.pid_to_proc = {}
+        self.pid_to_outputs = {}
+        self.pid_to_cwd = {}
+    
+    # TODO: this state should probably live in execution table, not in runner
+    def get_cwd(self, pid):
+        return self.pid_to_cwd[pid]
+        
+    def run(self, cmd_string, docker, cwd):
+        if docker:
+            cmd_string = 'docker run -v {}:/root -w /root {} bash -c "{}"'.format(cwd, docker, cmd_string)
+
+        # TODO: previously these files were having trailing whitespace stripped when written.  Need to move that to when they're read
+        stdout_path = os.path.join(cwd, 'stdout')
+        stderr_path = os.path.join(cwd, 'stderr')
+        stdout_file = open(stdout_path, 'w')
+        stderr_file = open(stderr_path, 'w')
+            
+        print(colorize(cmd_string, ansi=9))
+        proc = subprocess.Popen(
+            cmd_string,
+            shell=True,
+            universal_newlines=True,
+            stdout=stdout_file,
+            stderr=stderr_file,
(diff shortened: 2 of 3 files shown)


scottfrazer:
Is this ready for review/merge or are you still adding to it?

python/wdl/engine.py
```diff
@@ -1,12 +1,11 @@
 from wdl.binding import *
 from wdl.util import *
-import wdl.parser
 import re
 import subprocess
 import tempfile
 import uuid
-import json
-import sys
+import xml.etree.ElementTree as ETree
```
scottfrazer: Will setup.py need to add `lxml` or something to it for this to work?

pgm: ElementTree is part of the stdlib in python3

https://docs.python.org/3/library/xml.etree.elementtree.html#module-xml.etree.ElementTree

pgm:
Yes, this is ready for review.   I only would have new changes if I
discover bugs while trying to use it.   ​From what I've done so far, it
appears to work.

I haven't thought much about a more comprehensive testing strategy though.

python/wdl/engine.py
scottfrazer: Huh... look at that, I learned something new today!

python/examples/1.wdl
```diff
@@ -14,7 +14,7 @@ task grep {
 
 task wc {
   command {
-    wc -l ${sep=' ' File files+} | tail -1 | cut -d' ' -f 2
+    wc -l ${sep=' ' File files+} | tail -1 | tr -s ' ' | cut -d' ' -f 2
```
scottfrazer: I suspect this is because I'm using the GNU versions of all of these tools and the default on Mac OS X is BSD-style versions of all of these tools.  Fine 95% of the time until you notice whitespace differences that and odd quirks that differ between them.

Surprisingly this is not the first time this has happened this week.

python/wdl/main.py
```diff
@@ -56,11 +59,17 @@ def cli():
 
     if cli.action == 'run':
         inputs = {}
+        run_service_name = "local"
+
+        if cli.sge:
+            run_service_name = "sge"
+
```
scottfrazer: Not that I care that much, but `--backend=[local|sge]` would probably be how I'd set this up.  It's not really an issue until there are more backends implemented

python/wdl/engine.py
```diff
@@ -178,69 +331,87 @@ def __init__(self, workflow, inputs={}):
...
-            close_fds=True,
-            cwd=cwd
-        )
-        stdout, stderr = proc.communicate()
-        print(colorize('rc = {}'.format(proc.returncode), ansi=1 if proc.returncode!=0 else 2))
-        with open(os.path.join(cwd, 'stdout'), 'w') as fp:
-            fp.write(stdout.strip(' \n'))
-        with open(os.path.join(cwd, 'stderr'), 'w') as fp:
-            fp.write(stderr.strip(' \n'))
-        return (proc.pid, proc.returncode, stdout.strip(' \n'), stderr.strip(' \n'))
```
scottfrazer: The polling backoff is a cool feature

scottfrazer:
:+1: This is really cool, thanks for implementing this

python/wdl/main.py
pgm: Good point.  I'll change that.

scottfrazer:
If you add any more SGE related stuff, it'd be nice to change the README.md file too.  It shows the `--help` output right now and that's slightly out of date, and it'd be nice to have a little section on how to use SGE support.

PR #9, opened 2015-07-28
Update SPEC.md

geoffjentry:


SPEC.md
```diff
@@ -908,7 +908,7 @@ workflow wf {
 $scatter = 'scatter' '(' $identifier 'in' $expression ')' '{' $workflow_element* '}'
 ```
 
-A "scatter" clause defines that the everything in the body can be run in parallel.  The clause in parenthesis declares which collection to scatter over and what to call each element.
+A "scatter" clause defines that everything in the body can be run in parallel.  The clause in parenthesis declares which collection to scatter over and what to call each element.
```
mcovarr: parenthesis -> parentheses

PR #18, opened 2015-12-30
initial copying of cromwell's scala wdl bindings

geoffjentry:
Mostly a copy of cromwell.binding over to the WDL repo surrounded by an sbt project to build a jar and with engine-specific stuff removed, specifically:

* RuntimeAttributes has been massively slimmed down as all but memory units is engine dependent
* References to Backend and BackendType have been removed
* Hashing, which was introduced by call caching was removed as it is the responsibility of an engine
* IoInterface was removed as that's the responsibility of engines
* Removed a variety of things (largely in the package object) which weren't actually used by cromwell.binding
* Discarded Cromwell specific fields from Call (docker, failOnStderr, continueOnReturnCode)

In terms of code review, this will *not* be a review of the copied code and I will not handle such requests. Interested parties can file a PR after this is merged in.

Things to review:

* The sbt project structure
* The READMEs (WDL and wdl-scala)
* Modified code - see above for pointers to key spots
* Things which should be removed. I have a feeling there's more stuff which belongs in the realm of engines lurking in here


geoffjentry:
@cjllanwarne You seem to be online now and then so tagging you as well

kcibul:
For the runtime attributes, what about cores/cpus?  Or another way... what makes memory non-backend specific?  They seem to both be backend specific (for example, the local backend doesn't do anything with the memory parameter, right?).  But perhaps the guidance as to what is in the WDL spec versus backend specific would be "if its reasonable to expect that a backend would make use of it" they would be part of the spec (so both cores and memory would fit that definition I think).

geoffjentry:
@kcibul They're engine specific, not backend. None of that is specified by the WDL spec, which states that runtime attributes are generally arbitrary key/value pairs which may or may not be interpreted by the engine. It specifies that 'docker' and 'memory' are known keys and has verbiage on allowed values for 'memory'.

Personally I'd not have even gone that far in regards to the spec as I think it sets a bad precedent leading to more and more stuff being added as known key/values and would argue against adding any more. However this branch is validating against what the spec, well, specs :)

So for instance the Cromwell README still goes into a good amount of depth as to what runtime attributes are recognized, what the values allowed are and which backends support what.

geoffjentry:
@kcibul Thinking about it a bit more, I could be convinced that some/many/all of the k/v pairs we came up with for cromwell could be standardized into the WDL spec, although obviously the spec would declare these to all be fully optional for an engine (or backend, in the case of cromwell) to support.

kcibul:
Maybe this should be under the category of "it's a move, not a refactor" at the moment.  I see what you mean, and agree about not wanting bloat the runtime section.  The only competing concern is the aspirational portability goal that the WDL written for one engine has a hope of being run another engine, which may require 'standardizing' the things needed to make that possible (like docker image).  It's a careful balance though.  Certainly doesn't hold up this PR, but maybe something to talk through after the break (history aside, guidelines about what part of the runtime section should be in the WDL spec)

geoffjentry:
@kcibul Yup, all of that was the basis of my inner struggle. Was also thinking that this would all be a good candidate for post-transition PR, there are already a few things like that I know will need to be done - e.g. this PR doesn't address the command line calls in cromwell. So yeah, agreed - let's just follow the spec for now and clean up once the dust settles

scala/wdl_scala/build.sbt
```diff
@@ -0,0 +1,66 @@
+import sbt.Keys._
+import sbtassembly.MergeStrategy
+
+name := "wdl-scala"
+
+version := "0.01"
```
kshakir: Why `01` and not `1`?

scala/wdl_scala/project/plugins.sbt
```diff
@@ -0,0 +1,9 @@
+addSbtPlugin("com.typesafe.sbt" % "sbt-git" % "0.6.4")
+
+addSbtPlugin("com.github.gseitz" % "sbt-release" % "0.8.3")
+
+addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.14.0")
```
kshakir: Bump this to `0.14.1` and delete the file `scala/wdl_scala/project/assembly.sbt`

mcovarr:
Possibly unrelated to this PR:

It agree it feels weird to have `memory` be the only resource specifier afforded special status by the spec.  I'm also not too enthused with the current "silently ignore unsupported and possibly unrecognized runtime attributes" behavior (I believe if a runtime attribute key is unrecognized by any backend it will be logged, but that may not be visible to the user).  The combination of these two allows results in the local backend silently ignoring `memory` specifications even though it seems reasonable to expect Cromwell's Docker support to honor it via `--memory`.

Just something to think about after the dust settles from this refactor.

README.md
```diff
@@ -88,7 +101,405 @@ workflow wf {
 }
 ```
 
-Architecture
-------------
+# Getting Started with WDL
+
+We'l use Cromwell to run these examples but you can use any WDL engine of your choice. If you don't already have a reference to the Cromwell JAR file, one can be [downloaded](https://github.com/broadinstitute/cromwell/releases) or [built manually](#building)
+
```
mcovarr: The `#building` link looks like C/P from the Cromwell README and doesn't resolve in this README.

README.md
```diff
@@ -88,7 +101,405 @@ workflow wf {
...
+workflow test {
+  call hello
+}
+```
+
+Generate a template `hello.json` file with the `inputs` subcommand:
+
+```
+$ java -jar cromwell.jar inputs hello.wdl
```
mcovarr: Do these Cromwell command line usages really belong in this README?

mcovarr:
`.gitignore` entries for the SBT `project` and `target` directories would be nice.

scala/wdl_scala/src/main/java/wdl_scala/parser/MemorySize.java
```diff
@@ -0,0 +1,35 @@
+package wdl_scala.parser;
```
mcovarr: Must we have an underscore in the package component name?

scala/wdl_scala/src/main/scala/wdl_scala/RuntimeAttributes.scala
```diff
@@ -0,0 +1,76 @@
...
+    * Checks that a provided memory value is valid, if not throws an IllegalArgumentException
+    */
+  private def validateMemory(value: String): Unit = {
+    validateMemStringInGb (value) match {
+      case Failure(f) => throw new IllegalArgumentException("Invalid memory value specified:\n" + f.toList.mkString("\n"))
+      case Success(_) => ()
+    }
+  }
+
+  private def validateMemStringInGb(mem: String): ValidationNel[String, Double] = {
```
mcovarr: looks like `ErrorOr` should find its way over to this repo, maybe even applied to a feline library

mcovarr:
Shouldn't this be set up to build in Travis now?

mcovarr:
What of the Cromwell side of this to show the refactor is workable?

scala/wdl_scala/src/main/scala/wdl_scala/ContinueOnReturnCode.scala
```diff
@@ -0,0 +1,31 @@
+package wdl_scala
```
mcovarr: Is this file really meant to be here?

README.md
geoffjentry: For now, yes. Note his is only the library portion and not he clp. That'll be later

scala/wdl_scala/src/main/scala/wdl_scala/RuntimeAttributes.scala
geoffjentry: *points to the statement in the PR about people being free to submit PRs* :)

geoffjentry:
Re Cromwell, that's in progress. There's probably something which doesn't work here but we can fix it but I wanted this in place first

geoffjentry:
Re your runtime attr supporting comment, that's 100% unrelated to this PR, eg from wdls perspective there's no such thing as a local backend

scala/wdl_scala/build.sbt
geoffjentry: *shrug* I think I meant to say 0.0.1

README.md
geoffjentry: There's a reason it looks to be C/P :)

scala/wdl_scala/src/main/scala/wdl_scala/ContinueOnReturnCode.scala
geoffjentry: Nope

geoffjentry:
@mcovarr re Travis I don't have the access to set it up for this repo, it'll have to be done separately

geoffjentry:
@kshakir @mcovarr I think this should cover all of the review comments (except for a couple of Miguel's which are in the "won't fix" category)

README.md
```diff
@@ -88,7 +101,405 @@ workflow wf {
 }
 ```
 
-Architecture
-------------
+# Getting Started with WDL
+
+We'l use Cromwell to run these examples but you can use any WDL engine of your choice. If you don't already have a reference to the Cromwell JAR file, one can be [downloaded](https://github.com/broadinstitute/cromwell/releases)
```
kshakir: "We'll" is misspelled.

README.md
```diff
@@ -88,7 +101,405 @@ workflow wf {
...
+Generate a template `hello.json` file with the `inputs` subcommand:
+
+```
+$ java -jar cromwell.jar inputs hello.wdl
+{
+  "test.hello.name": "String"
+}
+```
+
+WDL has a concept of fully-qualified names.  In the above output, `test.hello.name` is a fully-qualified name which shoul\
```
kshakir: The slashes at the EOL are visible in the [README](https://github.com/broadinstitute/wdl/tree/1e14be65402b2b43e7b6d1bff0ae7100521f85a1#hello-world-wdl)

scala/wdlscala/README.md
```diff
@@ -0,0 +1,60 @@
+# Scala API Usage
+
+The main entry point into the parser is the `WdlNamespace` object.  A WDL file is considered a namespace, and other namespaces can be included by using the `import` statement (but only with an `as` clause).
+
+```scala
+import java.io.File
+import cromwell.parser.BackendType
+import cromwell.binding.NamespaceWithWorkflow
```
kshakir: Sorry, disappearing offline again, but is this package correct in this example?

README.md
geoffjentry: Crap, thought I got all of those. Was an artifact from how I did the C/P

scala/wdlscala/README.md
geoffjentry: Nope

geoffjentry: And BackendType doesn't even exist anymore, sheesh

geoffjentry:
@kshakir @mcovarr - Responded to Khalid's most recent comments. I was hoping to get this merged in over the weekend so we can swarm on the next phases come monday morning. So if you all get a chance can you see if this passes muster? This (and Cromwell) will be iterated on a bit over the next week or two so it's not intended to be perfect, just usable.

kshakir:
Gotta run, and not sure how much more time I'll have this weekend. I looked over build.sbt and lightly scanned for standout issues, so willing to give this a LGTM

scala/wdlscala/src/main/scala/wdlscala/Workflow.scala
```diff
@@ -0,0 +1,174 @@
...
+      outputs
+    } else {
+      (outputs ++ inputs) filter {
+        case PotentialReportableSymbol(fqn, wdlType, wildcardsAllowed) =>
+          workflowOutputDecls.isEmpty || workflowOutputDecls.exists(_.outputMatchesDeclaration(fqn, wildcardsAllowed))
+      }
+    }
+
+    val x = filtered map { case PotentialReportableSymbol(fqn, value, wildcardAllowed) => ReportableSymbol(fqn, value) }
+    x
```
mcovarr: :do_not_litter:

mcovarr:
I applaud seizing the opportunity to fix whitespace ala

```scala
case f:Foo => ...
```
to 
```scala
case f: Foo => ...
```

but IDE search points out quite a few more remaining

mcovarr:
urg, `wdl_scala` -> `wdlscala` was not quite what I had in mind.  :smile:  Does the `scala` part really need to be in the name?  That feels weird and redundant.

`wdl/scala/wdlscala/src/main/scala/wdlscala`

mcovarr:
:+1:

geoffjentry:
Re wdlscala - what do you call the next language and how do you differentiate it?

scala/wdlscala/src/main/scala/wdlscala/Workflow.scala
geoffjentry: This might cause me to overrule my ban on fixing things I didn't change. I curse you for bringing this to my attention

geoffjentry:
The Python one is pywdl, to be uniform this should be scawdl or scalawdl or something I suppose

geoffjentry:
Hmmm my inner ska fan would enjoy ScaWdl

geoffjentry:
Thinking more, something like "swdl" or "wdls' would both differentiate it a la "pywdl" and fit a more standard scala/java naming convention for projects of the "it's Foo for scala/java"

scala/wdlscala/src/main/scala/wdlscala/Workflow.scala
geoffjentry: Decided I'd allow this one since I probably did it in the first place :)

geoffjentry:
Ok, merging this in. There were a variety of good suggestions for general cleanups, once this is in I consider it fair game if people want to be good samaritans and submit PRs to cleanse some of the sins of the past. 

@mcovarr I went with wdl4s, it had a nice ring to it for me in terms of fitting common patterns and helping to differentiate from a future project e.g. kotlin bindings.

PR #19, opened 2016-01-05
Updating specification

scottfrazer:


SPEC.md
```diff
@@ -847,21 +847,77 @@ Any `${identifier}` inside of a string literal must be replaced with the value o
...
+    python script.py
+  }
+  runtime {
+    docker: ["ubuntu:latest", "broadinstitute/scala-baseimage"]
+  }
+}
+```
+
+The value for the `docker` runtime attribute in this case is an array of values.  The parser should accept this.  Some engines might interpret it as an "either this image or that image" or it could reject it outright.
+
```
mcovarr: it could => could

SPEC.md
```diff
@@ -847,21 +847,77 @@ Any `${identifier}` inside of a string literal must be replaced with the value o
...
+  command {
+    python script.py
+  }
+  runtime {
+    docker: ["ubuntu:latest", "broadinstitute/scala-baseimage"]
+  }
+}
+```
+
+The value for the `docker` runtime attribute in this case is an array of values.  The parser should accept this.  Some engines might interpret it as an "either this image or that image" or could reject it outright.
```
geoffjentry: I'm merging this as-is but for the future IMO it should be clear that engines also have the freedom to ignore keys/values/etc as they see fit, not just reject

PR #32, opened 2016-03-23
Add sub function to Spec

Horneth:


SPEC.md
```diff
@@ -2114,6 +2115,42 @@ task example {
...
+
+## String sub(String, String, String)
+
+Given 3 String parameters `input`, `pattern`, `replace`, this function will replace any occurence matching `pattern` in `input` by `replace`.
+`pattern` is expected to be a [regular expression](https://en.wikipedia.org/wiki/Regular_expression). There are different implemetations of regular expression evaluators, different sets of rules etc... This choice is left to the implementor.
+
+Example 1:
+
+```wdl
+  String chocolike = "I like chocolate when it's late"
```
scottfrazer: :-D

SPEC.md
```diff
@@ -86,6 +86,7 @@
   * [File write_objects(Array\[Object\])](#file-write_objectsarrayobject)
   * [File write_json(mixed)](#file-write_jsonmixed)
   * [File size(File, \[String\])](#float-sizefile-string)
+  * [File sub(String, String, String)](#string-substring-string-string)
```
scottfrazer: link is not correct, should be `#string-sub-string-string-string` I think

scottfrazer: I don't really advertise this much, but I wrote a tool called `mdtoc` which is useful for stuff like this

```
$ pip install mdtoc
$ mdtoc SPEC.md
```

Seems that the `size()` function link might be wrong too?

scottfrazer: bah, I guess I was wrong, but the two headers are still not quite right

```diff
-  * [File size(File, \[String\])](#float-sizefile-string)
-  * [File sub(String, String, String)](#string-substring-string-string)
+  * [Float size(File, \[String\])](#float-sizefile-string)
+  * [String sub(String, String, String)](#string-substring-string-string)
```

scottfrazer:
Besides the table of contents changes, :+1:

SPEC.md
Horneth: Oh nice catch. For the link the way I do it is I create the section, generate an html preview and then use whatever link the anchor points to.

SPEC.md
```diff
@@ -2114,6 +2115,42 @@ task example {
 Supported units are KiloByte ("K", "KB"), MegaByte ("M", "MB"), GigaByte ("G", "GB"), TeraByte ("T", "TB") as well as their [binary version](https://en.wikipedia.org/wiki/Binary_prefix) "Ki" ("KiB"), "Mi" ("MiB"), "Gi" ("GiB"), "Ti" ("TiB").
 Default unit is Bytes ("B").
 
+
+## String sub(String, String, String)
+
+Given 3 String parameters `input`, `pattern`, `replace`, this function will replace any occurence matching `pattern` in `input` by `replace`.
+`pattern` is expected to be a [regular expression](https://en.wikipedia.org/wiki/Regular_expression). There are different implemetations of regular expression evaluators, different sets of rules etc... This choice is left to the implementor.
```
mcovarr: implemetations => implementations

That spelling of `implementor` may be a [passing fad](https://books.google.com/ngrams/graph?content=implementer%2Cimplementor&year_start=1800&year_end=2000&corpus=15&smoothing=3&share=&direct_url=t1%3B%2Cimplementer%3B%2Cc0%3B.t1%3B%2Cimplementor%3B%2Cc0)?

Do we really want to leave regexes so nebulously specified?

SPEC.md
```diff
@@ -2114,6 +2115,42 @@ task example {
 Supported units are KiloByte ("K", "KB"), MegaByte ("M", "MB"), GigaByte ("G", "GB"), TeraByte ("T", "TB") as well as their [binary version](https://en.wikipedia.org/wiki/Binary_prefix) "Ki" ("KiB"), "Mi" ("MiB"), "Gi" ("GiB"), "Ti" ("TiB").
 Default unit is Bytes ("B").
 
+
+## String sub(String, String, String)
+
+Given 3 String parameters `input`, `pattern`, `replace`, this function will replace any occurence matching `pattern` in `input` by `replace`.
```
mcovarr: occurence => occurrence

SPEC.md
```diff
@@ -2114,6 +2115,42 @@ task example {
...
+```wdl
+  String chocolike = "I like chocolate when it's late"
+
+  String chocolove = sub(chocolike, "like", "love") # I love chocolate when it's late
+  String chocoearly = sub(chocolike, "late", "early") # I like chocoearly when it's early
+  String chocolate = sub(chocolike, "late$", "early") # I like chocolate when it's early
+}
+```
+
+The sub function will also accept `input` and `replace` parameters that can be coereced to a String (eg File). This can be useful to swap the extension of a filename for example
```
mcovarr: eg => e.g.

mcovarr: coereced => coerced

SPEC.md
cjllanwarne: I guess the problem is, the Python  equivalent of WDL will have a different regex evaluator?

Horneth: If not, how much more specific do we want to be ? I personally am no regex expert and I don't have any really strong opinion about which standard we should encourage/enforce :/

mcovarr: Yeah I get that we'll have different impls under the hood, I just worry about setting up incompatibility issues.  Maybe not a big deal

mcovarr:
:+1:

PR #37, opened 2016-04-22
Point at the official website

geoffjentry:


geoffjentry:
@scottfrazer @knoblett can you two check this out?

scottfrazer:
uhh I didn't think we were gutting the whole entire thing...  those are good examples which just need a little bit of updating!  It'd be a shame to lose them entirely... and also the website doesn't (yet) entirely cover these examples.  For example, I was unable to find an example that used a `read_int` function

If we're going to remove them from this README, can we make sure all concepts are exercised in these examples are also on the web site in some form?

vdauwera:
Oof that does seem a bit extreme. I would recommend copying that content to a separate document that is labeled as a developer-oriented resource.

geoffjentry:
@scottfrazer The gutting part is what I was asking about earlier, not updating them. I think the less overlap we have with the website the better - @vdauwera @knoblett can we get richer examples on the website at some point in the not too distant future?

vdauwera:
Certainly we can plan to add examples to cover all of this on the website -- how about creating issues to that effect in this repo?

geoffjentry:
@vdauwera All of it or just the stuff which doesn't overlap with the site? The fact that we have this issue here demonstrates the pain of having the same effective stuff in multiple places

geoffjentry:
@vdauwera my last comment was in response to your comment prior to the issue filing btw.

vdauwera:
@geoffjentry Yep, got that. I'm reluctant to get rid of any of that content right away because I wonder if it might be useful as an orthogonal approach aimed at developers. I'd like to keep it in the back pocket until we've had a chance to get more feedback on how different types of people feel about the website content and presentation.

vdauwera:
So I guess I'm saying, keep the whole thing on the side for now, with the possibility of removing it entirely a few months down the road.

scottfrazer:
Well sorry I misunderstood you... I don't think we should remove these until the web site covers these concepts sufficiently... I know that people have used and liked these examples!  It'd be a shame to remove them with only a promise that eventually they'll be added to the web site again... I'll very gladly fix these docs this weekend if that's what it'll take

geoffjentry:
@vdauwera I was hoping to avoid having to actually fix what's broken in there, doh ;) I'll close this for now, may or may not get around to fixing the examples & moving them

vdauwera:
Well it would still be good to add the reference to the website (the first paragraph you added) and maybe state explicitly that there are these two types of docs -- say "read on for a developer-oriented quick-start doc, or head over to the website for more hand-holding"

PR #42, opened 2016-07-14
Update PairedSingleSampleWf and add JSON of inputs

vdauwera:
- Generate public WDL from finalized private WDL
- Replace 160624 WDL with the new WDL (move that one to the archive)
- Add JSON of example inputs with dummy filepaths (to be replaced with bucket addresses once the Verily bucket is up)
- Amend README to reflect presence of inputs JSON


vdauwera:
The WDL was tested already so the only thing to review is the README.

vdauwera:
As discussed with @geoffjentry, I'm taking on stewardship of the scripts portion of the wdl repo, so review by red team is no longer necessary for this kind of PR.

PR #59, opened 2016-09-16
PublicPairedSingleSampleWf: create reference cache from fasta.

mbookman:
Avoids fetching from EBI servers during "samtools index <cram>"
by creating the local reference cache from the input fasta.
The fetch from EBI can hang "samtools index" as described here:
  https://github.com/samtools/htslib/issues/412
The approach added here is that recommended by:
  http://www.htslib.org/workflow/#the-refpath-and-refcache
except without the fallback to EBI.


vivster7:
@vdauwera, @sooheelee, is comms in charge of maintaining this repo? Or is it red team?

vdauwera:
Yep, this is on us.

vivster7:
I actually made the same change in #60, generated it from a new private wdl. If you can review that, we can close this PR.

vdauwera:
implemented in #60

PR #60, opened 2016-09-27
Generate new public single sample gotc WDL

vivster7:
Same changes as seen in #59, except generated from private wdl.

Also includes a docker version bump.


vivster7:
@sooheelee, @vdauwera is comms in charge of this repo? if so, we should probably add you guys to the pull approve.

geoffjentry:
@vivster7 They own the repo now although I'm sure you could get one of us to opine if in a time crunch

vivster7:
@geoffjentry, I think comms should look at the PR, thanks though.

vdauwera:
Will take a look tomorrow, thanks

sooheelee:
Just a question @vivster7. For the changes to the code to be effective, does one need to use the updated Docker? Or will the new code changes work fine in the original Docker?

vivster7:
The original Docker would work as well for these specific changes. The docker change bumped the GATK version from 3.6-2-g69af359 to 3.6-44-ge7d1cd2 -- it should have no effect on the public wdl since we GATK36 is not used in the public WDL. 

Unsure why your diff is showing a change for `--read_filter`. Did you run `diff PublicPairedSingleSampleWf_160927.wdl archive/PublicPairedSingleSampleWf_160720.wdl`?

sooheelee:
@vivster7 I just used `diff` on my terminal. Hmm. 

Another question, I'm using `docker inspect` and see we reverted to a prior minor release of Java:

```
<                 "JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64/jre",
<                 "JAVA_VERSION=8u91",
<                 "JAVA_DEBIAN_VERSION=8u91-b14-1~bpo8+1",
---
>                 "JAVA_VERSION=8u66",
>                 "JAVA_DEBIAN_VERSION=8u66-b17-1~bpo8+1",
```
Is this intentional? If so, I'm curious to know why, e.g. if it impacts GATK tool performance.

Other changes include changing Docker version from 1.8.1 to 1.8.2, and as you say, upping our GATK36 version.

vivster7:
The Java and Docker version changes were unintentional, and I believe not harmful as well (its currently the Docker image we use to run production).

vivster7:
Would like to get this closed out. @vdauwera, do you also want to review this?

vdauwera:
Yes, will do in the next hour.

vdauwera:
@vivster7 Sorry for the delay in reviewing this. Just a couple of nitpicks. 

1. For all files (not just the wdl), rather than rename the original files to the new timestamp, we prefer to move the originals to the [archive](https://github.com/broadinstitute/wdl/tree/develop/scripts/broad_pipelines/archive) directory, and create new files with the new timestamp. That allows git-naive users to easily find older versions (eg to replicate someone else's findings) without having to look at the git history. The older wdl versions in the archive don't have all the associated files because we hadn't generated them for those wdls, but we do want to do this for all the files going forward.

2. Did you generate the new public WDL from the private one using Kate's script, or did you apply the same modifications independently to both? There may have been a few other tweaks to the private WDL that we didn't yet copy over (because they were too minor to warrant it) and I'd like to make sure that the private and public don't accumulate any divergences (which is why we have Kate's script).

vivster7:
1. The inputs and options files are the exact same, so I'll just copy them over with the old name so they can be tied to the old wdl

2. I used Kate's script to generate this wdl

scripts/broad_pipelines/PublicPairedSingleSampleWf_160927.wdl
```diff
@@ -0,0 +1,771 @@
...
+  Int disk_size
+  Int preemptible_tries
+
+  command <<<
+      samtools view -C -T ${ref_fasta} ${input_bam} | \
+      tee ${output_basename}.cram | \
+      md5sum > ${output_basename}.cram.md5
+
+      # Create REF_CACHE. Used when indexing a CRAM
+      seq_cache_populate.pl -root ./ref/cache ${ref_fasta}
```
vdauwera: Where does seq_cache_populate.pl live? Is it in the docker image?

vdauwera:
Great, thanks @vivster7. Just one more question related to the seq_cache_populate.pl script.

vdauwera:
Great,  all good then :+1:

PR #63, opened 2016-10-04
Create GATK wrapper which includes a workflow for each WDL

knoblett:
After quite a few iterations, this ended up being the simplest solution-- to add a workflow-writing section to the original gatkToWdlWrapper.py in order for it to be used on Firecloud. (Firecloud requires a workflow section in order for the files to be uploaded.)

Is it worthwhile to post the WDL files generated using this script? Or just contact the Firecloud team and get those uploaded to their repository?


vdauwera:
Wouldn't hurt to have single-tool WDLs for all the tools in wrappers/gatk/. However I wouldn't name the script (or the workflows) Firecloud anything since they're not really specific to firecloud; they can be used on any platform.

knoblett:
Files added and updated. Back to you @vdauwera

vdauwera:
OK, a few nitpicks/suggestions:

1. Would it make sense perhaps to rename WDLTasks_3.6 to GATKTasks_3.6, for consistency?
2. The comment block in a workflow wdl (as opposed to a task wdl) should reflect that it's a workflow, not a standalone task. 
3. The README needs to be updated to reflect the directory structure + difference between task and workflow WDLs.

vdauwera:
@knoblett Quick ping -- is this still on your radar? No rush, just checking that you're not waiting on me here.

knoblett:
It was on hold until Firecloud had implemented arrays, since this PR was for that. Alternatively, if you'd like I can go through it now and merge. I can't remember if I had applied your previously suggested changes or not before this got put on the backburner, but I can check and apply them if needed.

vdauwera:
Oh I see, I had forgotten about that. Well, this could actually be useful beyond just FireCloud, so let's go ahead and finalize it (when you have time away from the workshop of course).

vdauwera:
*bump*

We'll also want to generate a set for 3.7

vdauwera:
Given the move to openWDL, this should live in the GATK codebase instead of the WDL repo... and should be built as part of the release process.

PR #72, opened 2016-11-17
Subworkflow WDL Doc

Horneth:


geoffjentry:
Making sure @mcovarr is on the assignee list as our resident grammar guru

SPEC.md
```diff
@@ -1237,14 +1237,75 @@ Conditionals only execute the body if the expression evaluates to true
...
+  output {
+    String t_out = t.out
+    String u_out = u.out
+    String input_as_output = w_input
+    String previous_output = u_out
+  }
+}
+```
+
+Note that they can't reference task inputs. However this can be achieved by declaring the desired task input as an output of this task.
```
geoffjentry: I don't understand what this means

SPEC.md
```diff
@@ -1274,6 +1335,54 @@ workflow wf {
 
 In this example, the fully-qualified names that would be exposed as workflow outputs would be `wf.task1.results`, `wf.altname.value`.
 
+# Sub Workflows
```
geoffjentry: I feel like this should just be part of the *call* block above

SPEC.md
```diff
@@ -1274,6 +1335,54 @@ workflow wf {
...
+
+Workflows can also be called inside of workflows, like a regular call.
```
geoffjentry: I think this phrasing is an artifact of this being a spiffy new feature. I'd phrase this as if it *is* a regular call, just a workflow instead of a task

SPEC.md
Horneth: It means with this task t 

```
task t {
   String i
   command {...}
   output {
      String o = i
   }
}
```
you can't do

```
workflow w {
  call t
  output {
     String wf_o = t.i
  }
}
```

but you can do 

```
workflow w {
  call t
  output {
     String wf_o = t.o
  }
}
```

mcovarr:
:+1:

geoffjentry:
👍

PR #75, opened 2016-12-02
Update import statement docs

ruchim:
These changes will be implemented in Crowell 0.23

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -90,8 +90,8 @@
   * [String sub(String, String, String)](#string-substring-string-string)
   * [Array\[Int\] range(Int)](#arrayint-rangeint)
   * [Array\[Array\[X\]\] transpose(Array\[Array\[X\]\])](#arrayarrayx-transposearrayarrayx)
-  * [Pair(X,Y) zip(X,Y) (#WdlPair-zip)
-  * [Pair(X,Y) cross(X,Y) (#WdlPair-cross)
+  * [Pair(X,Y) zip(X,Y)](#pairxy-zipxy)
+  * [Pair(X,Y) cross(X,Y)](#pairxy-crossxy)
 * [Data Types & Serialization](#data-types--serialization)
   * [Serialization of Task Inputs](#serialization-of-task-inputs)
     * [Primitive Types](#primitive-types)
@@ -589,7 +589,8 @@ $import = 'import' $ws+ $string ($ws+ 'as' $ws+ $identifier)?
 
 The import statement specifies that `$string` which is to be interpted as a URI which points to a WDL file.  The engine is responsible for resolving the URI and downloading the contents.  The contents of the document in each URI must be WDL source code.
 
-Every imported WDL file requires a namespace and it can be specified using an identifier (via the `as $identifer` syntax). If no namespace identifier is explicitly specified, then the default namespace is the filename of the imported WDL file. For all imported WDL files, the tasks and workflows imported from that file will only be accessible through that assigned [namespace](#namespaces)
+Every imported WDL file requires a namespace which can be specified using an identifier (via the `as $identifier` syntax). If you do not explicitly specify a namespace identifier then the default namespace is the filename of the imported WDL, minus the .wdl extension.
+For all imported WDL files, the tasks and workflows imported from that file will only be accessible through that assigned [namespace](#namespaces)
 
 ```wdl
 import "http://example.com/lib/analysis_tasks" as analysis
@@ -1389,7 +1390,7 @@ In this example, the fully-qualified names that would be exposed as workflow out
 
 # Namespaces
 
-Import statements can be used to pull in tasks/workflows from other locations as well as create namespaces.  In the simplest case, an import statement adds the tasks/workflows that are imported into the specified namespace.  For example:
+Import statements can be used to pull in tasks/workflows from other locations as well as to create namespaces.  In the simplest case, an import statement adds the tasks/workflows that are imported into the specified namespace.  For example:
 
 tasks.wdl
 ```
@@ -1411,7 +1412,7 @@ workflow wf {
 }
 ```
 
-Tasks `x` and `y` are inside the namespace `pyTasks`, and which is different from the 'wf' namespace belonging to the primary workflow.  However, if no namespace is specified for tasks.wdl:
+Tasks `x` and `y` are inside the namespace `pyTasks`, which is different from the `wf` namespace belonging to the primary workflow.  However, if no namespace is specified for tasks.wdl:
 
 workflow.wdl
 ```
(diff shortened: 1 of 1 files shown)


SPEC.md
```diff
@@ -587,17 +587,18 @@ $import = 'import' $ws+ $string ($ws+ 'as' $ws+ $identifier)?
 
 The import statement specifies that `$string` which is to be interpted as a URI which points to a WDL file.  The engine is responsible for resolving the URI and downloading the contents.  The contents of the document in each URI must be WDL source code.
 
-If a namespace identifier (via the `as $identifer` syntax) is specified, then all the tasks and workflows imported will only be accessible through that [namespace](#namespaces).  If no namespace identifier is specified, then all tasks and workflows from the URI are imported into the current namespace.
+Every imported WDL file requires a namespace and it can be specified using an identifier (via the `as $identifer` syntax). If no namespace identifier is explicitly specified, then the default namespace is the filename of the imported WDL file. For all imported WDL files, the tasks and workflows imported from that file will only be accessible through that assigned [namespace](#namespaces)
```
mcovarr: `...then the default namespace is the filename of the imported WDL file.`

I assume this means the base name minus the likely ".wdl" file extension?

SPEC.md
```diff
@@ -1400,29 +1401,29 @@ task y {
...
+  call tasks.x
+  call tasks.y
 }
 ```
 
-Now everything inside of `tasks.wdl` must be accessed through the namespace `ns`.
+Now everything inside of `tasks.wdl` must be accessed through the default namespace `tasks`.
 
-Each namespace contains: namespaces, tasks, and workflows.  The names of these needs to be unique within that namespace.  For example, there cannot be a task named `foo` and also a namespace named `foo`.  Also there can't be a task and a workflow with the same names, or two workflows with the same name.
+Each namespace contains: namespaces, tasks, and workflows.  The names of these needs to be unique within that namespace.  For example, there cannot be a workflow named `foo` and also an imported namespace named `foo`.  Also there can't be a workflow and a task in that workflow with the same names, or import two workflows with the same namespace. However, you can import two workflows with different namespace names that share all the same tasks. For example, you can import namespace 'foo' and 'bar', both of which contain a task 'baz', and you can call 'foo.baz' and 'bar.baz' inside the same primary workflow.
```
mcovarr: Each namespace contains: namespaces, tasks, and workflows. => Each namespace may contain namespaces, tasks, and at most one workflow.

(Little shaky on the "at most one workflow" bit)

mcovarr: The names of these needs => The names of the contained namespaces, tasks, and workflows need

mcovarr: `or import two workflows with the same namespace`

I'm not 100% sure what that's trying to say

SPEC.md
```diff
@@ -1400,29 +1401,29 @@ task y {
...
 workflow wf {
-  call x
-  call y
+  call pyTasks.x
+  call pyTasks.y
 }
 ```
 
-Tasks `x` and `y` are in the same namespace as workflow `wf` is.  However, if workflow.wdl could put all of those tasks behind a namespace:
+Tasks `x` and `y` are inside the namespace `pyTasks`, and which is different from the 'wf' namespace belonging to the primary workflow.  However, if no namespace is specified for tasks.wdl:
```
mcovarr: and which => which

mcovarr: you've single quoted wf rather than the customary backticks

SPEC.md
mcovarr: and it can be => which can be

SPEC.md
```diff
@@ -1386,7 +1387,7 @@ In this example, the fully-qualified names that would be exposed as workflow out
 
 # Namespaces
 
-Import statements can be used to pull in tasks/workflows from other locations as well as create namespaces.  In the simplest case, an import statement adds the tasks/workflows that are imported into the current namespace.  For example:
+Import statements can be used to pull in tasks/workflows from other locations as well as create namespaces.  In the simplest case, an import statement adds the tasks/workflows that are imported into the specified namespace.  For example:
```
mcovarr: as create => as to create

SPEC.md
ruchim: On your first 1 comment, I agree that it only contains 1 primary/main/to-be-executed workflow.

On your third comment -- what if I rephrase it to "or import two namespaces with the same names" ??

SPEC.md
ruchim: I dont explicitly mention the minus fie extension piece--but hopefully the example provided would clear that point.

SPEC.md
katevoss: For Miguel's third comment, I'd suggest "or two imported namespaces with the same names". 

TOL something like:
For example, there cannot be two imported workflows named `foo` in the same namespace, nor a workflow and a namespace both named `foo` in the same namespace. Similarly there cannot be a task `foo` in a workflow also named`foo`.

Question: is a workflow/namespace/task **in** a namespace, **has** a namespace, **named** a namespace, or something else? Using one verb consistently might help users conceptualize what the namespace is (and isn't).

SPEC.md
```diff
@@ -2260,17 +2255,52 @@ Example 2:
...
+
   output {
     File outputFile = ${output_file_name}
   }
 }
 ```
 
+## Array[Int] range(Int)
+
+Creates an array of integers of length equal to the range argument. For example `range(3)` provides the array: `(0, 1, 2)`.
```
katevoss: TOL: as I mentioned to @ruchim, the previous library sections have descriptions that follow this formula: "Given X, this returns Y". I'd suggest following this model, but it's not critical.

SPEC.md
```diff
@@ -90,6 +90,8 @@
   * [String sub(String, String, String)](#string-substring-string-string)
   * [Array\[Int\] range(Int)](#arrayint-rangeint)
   * [Array\[Array\[X\]\] transpose(Array\[Array\[X\]\])](#arrayarrayx-transposearrayarrayx)
+  * [Pair(X,Y) zip(X,Y) (#WdlPair-zip)
+  * [Pair(X,Y) cross(X,Y) (#WdlPair-cross)
```
ruchim: Thanks Kate!

SPEC.md
ruchim: I visualize the workflow+task to be members inside of a namespace. A namespace can also contain other namespaces (which consequently can have their own workflows + tasks). But like you said, we need more consistency throughout this doc on that note :)

ruchim:
@mcovarr @katevoss can you please review your requested changes and let me know when this is ready for your thumbs up, thanks :)

SPEC.md
```diff
@@ -587,17 +589,19 @@ $import = 'import' $ws+ $string ($ws+ 'as' $ws+ $identifier)?
 
 The import statement specifies that `$string` which is to be interpted as a URI which points to a WDL file.  The engine is responsible for resolving the URI and downloading the contents.  The contents of the document in each URI must be WDL source code.
 
-If a namespace identifier (via the `as $identifer` syntax) is specified, then all the tasks and workflows imported will only be accessible through that [namespace](#namespaces).  If no namespace identifier is specified, then all tasks and workflows from the URI are imported into the current namespace.
+Every imported WDL file requires a namespace which can be specified using an identifier (via the `as $identifier` syntax). If you do not explicitly specify a namespace identifier then the default namespace is the filename of the imported WDL, minus the .wdl extension.
+For all imported WDL files, the tasks and workflows imported from that file will only be accessible through that assigned [namespace](#namespaces)
```
katevoss: Missing a period after `[namespace](#namespaces)`.

SPEC.md
```diff
@@ -1400,29 +1404,30 @@ task y {
...
 
-Now everything inside of `tasks.wdl` must be accessed through the namespace `ns`.
+Now everything inside of `tasks.wdl` must be accessed through the default namespace `tasks`.
 
-Each namespace contains: namespaces, tasks, and workflows.  The names of these needs to be unique within that namespace.  For example, there cannot be a task named `foo` and also a namespace named `foo`.  Also there can't be a task and a workflow with the same names, or two workflows with the same name.
+Each namespace may contain namespaces, tasks, and at most one workflow.  The names of the contained namespaces, tasks, and workflow need to be unique within that namespace. For example, one cannot import two workflows while they have the same namespace identifier. Additionally, a workflow and a namespace both named `foo` cannot exist inside a common namespace. Similarly there cannot be a task `foo` in a workflow also named `foo`.
+However, you can import two workflows with different namespace identifiers that have identically named tasks. For example, you can import namespaces 'foo' and 'bar', both of which contain a task 'baz', and you can call 'foo.baz' and 'bar.baz' from the same primary workflow.
```
mcovarr: should these single quotes be backticks?

mcovarr:
:+1:

PR #77, opened 2017-01-03
Documenting conditionals

cjllanwarne:
@katevoss not sure how you want to include this in your refactoring, but I wanted to at least provide the content for the conditionals section.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -1352,13 +1352,55 @@ Loops are distinct from scatter clauses because the body of a while loop needs t
 
 ### Conditionals
 
-:pig2: Coming soon in [Cromwell](https://github.com/broadinstitute/cromwell)
+:pig2: Available in [Cromwell](https://github.com/broadinstitute/cromwell) version 24 and higher
 
 ```
 $conditional = 'if' '(' $expression ')' '{' $workflow_element* '}'
 ```
 
-Conditionals only execute the body if the expression evaluates to true
+Conditionals only execute the body if the expression evaluates to true.
+
+* When a call's output is referenced outside the same containing `if` it will need to be handled as an optional type. E.g.
+```
+workflow foo {
+  # Call 'x', producing a Boolean output:
+  call x
+  Boolean x_out = x.out
+
+  # Call 'y', producing an Int output, in a conditional block:
+  if (x_out) {
+    call y
+    Int y_out = y.out
+  }
+  
+  # Outside the if block, we have to handle this output as optional:
+  Int? y_out_maybe = y.out
+
+  # Call 'z' which takes an optional Int input:
+  call z { input: optional_int = y_out_maybe }
+}
+```
+* Optional types can be coalesced by using the `select_all` and `select_first` array functions:
+```
+workflow foo {
+  Array[Int] scatter_range = [1, 2, 3, 4, 5]
+  scatter (i in scatter_range) {
+    call x { input: i = i }
+    if (x.validOutput) {
+      Int x_out = x.out
+    }
+  }
+
+  # Because it was declared inside the scatter and the if-block, the type of x_out is different here:
+  Array[Int?] x_out_maybes = x_out
+
+  # We can select only the valid elements with select_all:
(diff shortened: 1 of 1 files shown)


katevoss:
@knoblett here is the pull for review.

SPEC.md
```diff
@@ -1352,13 +1352,55 @@ Loops are distinct from scatter clauses because the body of a while loop needs t
...
+  if (x_out) {
+    call y
+    Int y_out = y.out
+  }
+  
+  # Outside the if block, we have to handle this output as optional:
+  Int? y_out_maybe = y.out
+
+  # Call 'z' which takes an optional Int input:
+  call z { input: optional_int = y_out_maybe }
```
knoblett: Do you need to explicitly define it as an optional `Int`? Or would you be able to feed `y.out` directly into task z? I would assume that z.optional_int is an optional variable.

```
call z { input: optional_int = y.out }
```

SPEC.md
```diff
@@ -1352,13 +1352,55 @@ Loops are distinct from scatter clauses because the body of a while loop needs t
...
+}
+```
+* Optional types can be coalesced by using the `select_all` and `select_first` array functions:
+```
+workflow foo {
+  Array[Int] scatter_range = [1, 2, 3, 4, 5]
+  scatter (i in scatter_range) {
+    call x { input: i = i }
+    if (x.validOutput) {
```
knoblett: I assume `x.validOutput` is a boolean generated in the task which checks if the output of x is valid for some reason. Along that line of thought though, will there be or could there be a function that checks if the output was generated or not? Essentially something like `variableName.exists()` would return true if `variableName` was output by the task.

SPEC.md
```diff
@@ -1352,13 +1352,55 @@ Loops are distinct from scatter clauses because the body of a while loop needs t
...
+      Int x_out = x.out
+    }
+  }
+
+  # Because it was declared inside the scatter and the if-block, the type of x_out is different here:
+  Array[Int?] x_out_maybes = x_out
+
+  # We can select only the valid elements with select_all:
+  Array[Int] x_out_valids = select_all(x_out_maybes)
```
knoblett: Doesn't `x_out` only contain valid elements already? It only took on a value if `x.validOutput` returned true.

SPEC.md
cjllanwarne: Correct! I only called it out in the separate line to show that the type has changed

SPEC.md
cjllanwarne: There are a lot of potential helper functions around optionals which don't exist yet, e.g.
```
  hasValue: (optional: X?) => Boolean
  getOrElse: (optional: X?, otherwise: X) => X
```

We didn't add them yet but definitely they're (1) possible and (2) easy to add and (3) would be great issues/tickets/requests. I think we just wanted to see what people really wanted before implementing everything possible all at once.

SPEC.md
cjllanwarne: Good point! In this example every element of `x_out_maybes` would be a real value, based on our knowledge of what `x.validOutput` means. But WDL won't be able to infer that in the general case so it still assigns `x_out_maybes` as an `Array[Int?]`. To get an `Array[Int]` from that we still need to use `select_all`.

Perhaps this is a slightly contrived/confusing example?

cjllanwarne: Oh wait, actually I think it is slightly different that what you're expecting. 

If the `x.validOutput` is true, then that entry in `x_out_maybes` is a value. If `x.validOutput` is false, there is still an entry in `x_out_maybes` for that, but it's an empty optional value. So this `select_all` is necessary after all to filter out the empties. Does that make sense?

SPEC.md
knoblett: Makes sense then :)

SPEC.md
knoblett: Good point!

SPEC.md
knoblett: Makes sense!

knoblett:
This should be ready to merge now, I believe.

PR #80, opened 2017-01-07
Add script to generate subsets of scatter intervals from master list

vdauwera:
Since Cromwell/WDL scatter-gather doesn't allow dynamic apportioning of intervals, we produce subsetted scatter intervals lists to scatter a fixed amount of ways. 

We offer such a set of intervals lists (50-ways) for human WGS against hg38, but we and others need to be able to quickly generate sets for other references, datatypes or of different sizes.


vdauwera:
This is ready for review. There's only one page of actual code (in create_scatter_intervals.py); the rest are demo files showing example input and expected outputs for the usage command.

PR #82, opened 2017-01-08
WDLs for trivial bam manipulation

vdauwera:
- Grab headers from a list of SAM/BAMs (with caveat and a better way to do it on GCP with gsutil)
- Validate a list of SAM/BAMs in SUMMARY mode

### DIFF
--- a/scripts/broad_dsde_workflows/ExtractSamHeadersWf_170107.inputs.json
+++ b/scripts/broad_dsde_workflows/ExtractSamHeadersWf_170107.inputs.json
@@ -1,0 +1,10 @@
+{
+  "ExtractSamHeadersWf.bam_list": [ 
+  	"gs://genomics-public-data/test-data/dna/wgs/hiseq2500/NA12878/H06HDADXX130110.1.ATCACGAT.20k_reads.bam",
+  	"gs://genomics-public-data/test-data/dna/wgs/hiseq2500/NA12878/H06HDADXX130110.2.ATCACGAT.20k_reads.bam",
+  	"gs://genomics-public-data/test-data/dna/wgs/hiseq2500/NA12878/H06JUADXX130110.1.ATCACGAT.20k_reads.bam"
+  ],
+
+  "ExtractSamHeadersWf.ExtractSAMHeader.mem_size": "1 GB",
+  "ExtractSamHeadersWf.ExtractSAMHeader.disk_size": 200
+}
--- a/scripts/broad_dsde_workflows/ExtractSamHeadersWf_170107.wdl
+++ b/scripts/broad_dsde_workflows/ExtractSamHeadersWf_170107.wdl
@@ -1,0 +1,84 @@
+## Copyright Broad Institute, 2017
+## 
+## This WDL extracts the headers from a list of SAM/BAMs
+##
+## Requirements/expectations :
+## - List of valid SAM or BAM files
+##
+## Outputs: 
+## - Set of .txt files containing the header, one per input file
+##
+## Cromwell version support 
+## - Successfully tested on v24
+## - Does not work on versions < v23 due to output syntax
+##
+## Runtime parameters are optimized for Broad's Google Cloud Platform implementation. 
+## For program versions, see docker containers. 
+##
+## Note that this is a really dumb way to get a SAM/BAM header when running on GCP 
+## because it will require localizing the entire SAM/BAM file, which can take a while. 
+## A much better way is to use gsutil like this:
+##
+##     gsutil cp gs://bucket/path/your.bam - | samtools view -H -
+##
+## And yes there's a weird trailing dash (-) there at the end; it's not a typo, leave 
+## it in. 
+##
+## If you're running locally or on a cloud platform that doesn't require localizing 
+## files then it should be fine. 
+##
+## LICENSING : 
+## This script is released under the WDL source code license (BSD-3) (see LICENSE in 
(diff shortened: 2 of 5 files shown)


vdauwera:
Note that this builds on https://github.com/broadinstitute/wdl/pull/81

Will run full tests before requesting review.

vdauwera:
- [ ] Test ExtractSamHeadersWf: fa16e964-ef39-4cbe-ba10-f104fc81c538
- [ ] Test ValidateBamsWf: 0fbe8061-a1e9-4684-9d83-3b057badaebc

vdauwera:
Both tests pass; this is ready to review.

scripts/broad_supported/ExtractSamHeadersWf_170107.wdl
```diff
@@ -0,0 +1,80 @@
...
+    call ExtractSAMHeader {
+      input:
+        bam_file = input_bam,
+        output_basename = sub(sub(input_bam, sub_strip_path, ""), sub_strip_suffix, "") + ".header"
+    }
+  }
+
+  # Outputs that will be retained when execution is complete
+  output {
+    ExtractSAMHeader.*
```
cjllanwarne: The new style syntax would be `File output_bam = ExtractSAMHeader.output_bam`. That'd make it much easier to compose this WDL into other WDLs in the future

scripts/broad_supported/ValidateBamsWf_170107.wdl
```diff
@@ -0,0 +1,72 @@
...
+    call ValidateBAM {
+      input:
+        bam_file = input_bam,
+        output_basename = sub(sub(input_bam, sub_strip_path, ""), sub_strip_suffix, "") + ".validation"
+    }
+  }
+
+  # Outputs that will be retained when execution is complete
+  output {
+    ValidateBAM.*
```
cjllanwarne: Again, I think our samples should use the new style syntax as much as possible

scripts/broad_supported/ExtractSamHeadersWf_170107.wdl
vdauwera: Oh interesting. What does this "new style" hang on? Is it a purely stylistic recommendation or an actual engine support thing? If so, at what version of Cromwell does this kick in?

vdauwera:
@cjllanwarne I updated the output syntax as you recommended, but I'm getting this validation error (with wdltool 0.9):

````
ERROR: validation_report is declared as a File but the expression evaluates to a Array[String]:

    File validation_report = ValidateBAM.validation_report
````

The task output is defined as `File validation_report = "${output_basename}.txt"`

Is it because the new syntax is not yet supported by wdltool, or did I misunderstand your recommendation?

**Comment by vdauwera** [comment]
And while we're on this code, I have a question re: using the value of optional args in output filenames.

I have an optional arg declared as `String ? validation_mode`, then I use it in my command as `MODE=${default="SUMMARY" validation_mode} `. Is there any way to use the value of `validation_mode` in the output filename, so that it will be e.g. `myFoo.SUMMARY.txt` if I don't set a mode, but `myFOO.BAR.txt` if I set `ValidateBAM.validation_mode: "BAR"` in the inputs.json?

**Comment by cjllanwarne** [comment]
@vdauwera 

(1) Oops, I didn't see the scatter! Cromwell is correct in telling me I got this wrong. Because it's coming from a call inside a scatter the output is indeed an `Array[File]`. e.g. `Array[File] validation_reports = ValidateBAM.validation_report`

(2) That looks like it should work, so might be a bug in Cromwell. You could maybe try doing it with an upfront input declaration instead of the two interpolations later on? eg:
```
String? validation_mode
String validation_mode_text = "${default="SUMMARY"validation_mode}"
```

and then use `validation_mode_text` instead of the `default` stuff later on?

vdauwera:
Aah, you illuminate my path like the North Star, @cjllanwarne. Thank you!

vdauwera:
Hmm. I updated the relevant lines and now my WDLs pass validation locally, but the Cromwell in the dsde-methods SwaggerUI is choking with this error:

````
{
  "status": "fail",
  "message": "Workflow input processing failed.",
  "errors": [
    "Unable to load namespace from workflow: Unrecognized token on line 77, column 10:\n\n    Array[File] output_headers = ExtractSAMHeader.output_header\n         ^"
  ]
}
````

With reformatting it looks like this (approximate caret position):

````
Array[File] output_headers = ExtractSAMHeader.output_header
         ^
````

Any idea what's wrong? Could it be a Cromwell version problem? I think this Swagger is running C19.

cjllanwarne:
Oh no! Cromwell 19 is not going to support new style output format. I think that's C23+

cjllanwarne:
FWIW that part 2 might also be a C19 bug that's fixed later. 

Also as a side-note, I wouldn't be tooo surprised if the space between `String ?` is not liked in later versions so I'd stick to `String?` if possible

vdauwera:
OK, the new style outputs are working correctly on the updated swagger-cromwell-24. 

However the optional/default input declaration for the validation mode has not been working from me. I decided for now to just require the validation_mode to be specified explicitly as an input, since it's a very reasonable requirement. But I would be curious to see if the other way could be made to work.

vdauwera:
Oh and is it fair to say 

> Note that you may also see the ? modifier used next to the Type with a space, like this:  `Type ? variableName`. Both formats are currently allowed (up to Cromwell v24), but the extra space may be disallowed in future versions.

vdauwera:
@cjllanwarne ^^

PR #84, opened 2017-01-09
prefix() proposal.  Closes broadinstitute/cromwell#375

mcovarr:
This is an attempt to solve the problem presented [here](https://github.com/broadinstitute/cromwell/issues/375) of composing a string from a prefix string and a list of zero or more value strings.  There are definitely a lot of ways this problem might be solved and this is a concrete proposal just to get a discussion started.  😄 

Pinging @slnovak @vdauwera @knoblett @katevoss @jsotobroad @LeeTL1220 @yfarjoun

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -95,6 +95,7 @@
   * [Pair(X,Y) zip(X,Y)](#pairxy-zipxy)
   * [Pair(X,Y) cross(X,Y)](#pairxy-crossxy)
   * [Integer length(Array\[X\])](#integer-lengtharrayx)
+  * [Array\[String\] prefix(String, Array\[String\])](#arraystring-prefixstring-arraystring)
 * [Data Types & Serialization](#data-types--serialization)
   * [Serialization of Task Inputs](#serialization-of-task-inputs)
     * [Primitive Types](#primitive-types)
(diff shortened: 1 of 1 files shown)


Horneth:
I also like the idea of having an `mkString` for arrays in general. I find the scala one pretty simple and complete but it doesn't have to be the same. 
http://www.scala-lang.org/api/2.10.3/#scala.collection.GenTraversableOnce

And we could still have `prefix`, `suffix` sugar method on top.

SPEC.md
```diff
@@ -94,6 +94,7 @@
   * [Array\[Array\[X\]\] transpose(Array\[Array\[X\]\])](#arrayarrayx-transposearrayarrayx)
   * [Pair(X,Y) zip(X,Y)](#pairxy-zipxy)
   * [Pair(X,Y) cross(X,Y)](#pairxy-crossxy)
+  * [String prefix(String, Array\[String\])](#string-prefix-string-arraystring)
```
cjllanwarne: This link didn't work for me

SPEC.md
```diff
@@ -2491,6 +2492,18 @@ Array[String] zs = [ "d", "e" ]
...
+## String prefix(String, Array[String])
+
+:pig2: [Cromwell supported](https://github.com/broadinstitute/cromwell#wdl-support) :white_check_mark:
+
+Given a String and an Array[String], the `prefix` function returns a String comprised of each element of the 
+Array prefixed by the String, all joined together.  For example:
+   
+```
+Array[String] env = ["key1=value1", "key2=value2", "key3=value3"]
+String env_param = prefix(" -e ", env) # " -e key1=value1 -e key2=value2 -e key3=value3"
```
cjllanwarne: Nice! The extra space at the start is awkward but (shrug)

SPEC.md
mcovarr: Whoops, that's because I have a superflous dash.  Thank you for reviewing so thoroughly!  😄

mcovarr:
@cjllanwarne a Scala-esque `mkString` wouldn't make this redundant since `mkString` only inserts strings between elements of the array, it doesn't prepend the strings to every element of the array.  For this particular use case, I want a completely blank string if the array is empty, and ` -e key1=value` if there's one element in the array.

Having said that I'm happy to make `mkString`, `prefix`, and `suffix` functions as @Horneth suggested if we're agreed those would be a good idea.

cjllanwarne:
@mcovarr good point! I think `prefix` adds by far the most value. `mkString` is nice (especially if it means we can retire the `${sep=" " array}` syntax although I would prefer a better name. Suffix, meh, but there might be uses for it so why not...

Horneth:
👍

cjllanwarne:
👍

PR #94, opened 2017-02-21
JavaScript support for a Hermes WDL grammar

sidoruka:
# General idea
Currently Hermes WDL grammar fully supports compilers generation only for Python and Java languages.

We would like to propose improved JavaScript support for a Hermes WDL grammar and hope these improvements to be useful for generating a WDL compiler for JavaScript (JS).

# Changes summary for grammar.hgr
* Adding JS regex
  * JS uses [\s\S] instead of '.' because JS doesn't have DOTALL flag for regex
* Adding JS enum support for wf_output
* Adding JS enum suport for task_fqn
* Adding JS support for generation (Lexer overloaded functions)
  * Renaming **unescape** to **wdl_unescape** because of name conflict with JS native function

# WDL Parser.JS, tests and sample code
The generated WDL parser, together with tests and sample code are added to **javascript** folder
* **javascript** 
  * **tests** 
    * **cases** a set of WDL files and expected ast files to use for test runs (WDL files are copied from [pywdl](https://github.com/broadinstitute/pywdl) repository)
    * **parser-test.js** tests to compare generated ast with expected one (mocha framework is used for testing)
  * **sample.js** example of wdl_parser.js usage
  * **wdl_parser.js** parser that was generated from new grammar.hgr
  * **README.md** describes sample.js usage and running tests

PR #95, opened 2017-02-21
Update grammar file to match Cromwell supported WDL version

Horneth:


geoffjentry:
👍 

@vdauwera Note this change. Our internal `grammar.hgr` file (which itself is another problem that needs fixing) had diverged from the one here. That causes #94 to not be completely correct.

PR #97, opened 2017-02-28
GA4GH Tool Execution Challenge Phase 1

vdauwera:
Files and readme for the GA4GH Tool Execution Challenge Phase 1 entry for Cromwell running on Google Genomics API wdl_runner as documented [here](https://www.synapse.org/#!Synapse:syn8080249/wiki/415105).

For the record I also made a FireCloud entry [here](https://www.synapse.org/#!Synapse:syn8080249/wiki/415107) using the same files. 

The primary purpose of storing these in the WDL repo is to keep a record of our entry associated with WDL, independently of the challenge platform. It also makes for a good minimal example of using the wdl_runner for non-GATK purposes.


scripts/ga4gh/tool_execution_challenge_phase1/empty.options.json
```diff
@@ -0,0 +1,3 @@
+{
+
+}
```
geoffjentry: Consider this optional to actually change but a good rule of thumb is to end files with a `\n` as if you cat them your prompt winds up on its own line, which is handy

geoffjentry:
:100: / :100:

scripts/ga4gh/tool_execution_challenge_phase1/empty.options.json
vdauwera: oh good point, thanks

PR #98, opened 2017-03-03
Optionals methods and pair literals

cjllanwarne:
Doc changes for the upcoming Cromwell 25 release.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -21,6 +21,7 @@
     * [Function Calls](#function-calls)
     * [Array Literals](#array-literals)
     * [Map Literals](#map-literals)
+    * [Pair Literals](#pair-literals)
   * [Document](#document)
   * [Import Statements](#import-statements)
   * [Task Definition](#task-definition)
@@ -92,10 +93,13 @@
   * [String sub(String, String, String)](#string-substring-string-string)
   * [Array\[Int\] range(Int)](#arrayint-rangeint)
   * [Array\[Array\[X\]\] transpose(Array\[Array\[X\]\])](#arrayarrayx-transposearrayarrayx)
-  * [Pair(X,Y) zip(X,Y)](#pairxy-zipxy)
-  * [Pair(X,Y) cross(X,Y)](#pairxy-crossxy)
+  * [Array\[Pair(X,Y)\] zip(Array\[X\], Array\[Y\])](#arraypairxy-ziparrayx-arrayy)
+  * [Array\[Pair(X,Y)\] cross(Array\[X\], Array\[Y\])](#arraypairxy-crossarrayx-arrayy)
   * [Integer length(Array\[X\])](#integer-lengtharrayx)
   * [Array\[String\] prefix(String, Array\[X\])](#arraystring-prefixstring-arrayx)
+  * [X select_first(Array\[X?\])](#x-select_firstarrayx)
+  * [Array\[X\] select_all(Array\[X?\])](#arrayx-select_allarrayx)
+  * [Boolean defined(X?)](#boolean-definedx)
 * [Data Types & Serialization](#data-types--serialization)
   * [Serialization of Task Inputs](#serialization-of-task-inputs)
     * [Primitive Types](#primitive-types)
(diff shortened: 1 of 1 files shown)


SPEC.md
```diff
@@ -597,6 +601,16 @@ Map[Int, Int] = {1: 10, 2: 11}
...
 ```
 
+### Pair Literals
+
+:pig2: [Cromwell supported](https://github.com/broadinstitute/cromwell#wdl-support) :white_check_mark:
+
+Pair values can be specified using another Python-like syntax, as follows:
+
+```
+Pair[Int, String] twenty_threes = (23, "twenty-three")
```
ruchim: not sure if the name "twenty_threes" was intentional, or it was meant to be "twenty_three"

SPEC.md
```diff
@@ -2566,6 +2580,26 @@ Array[Integer] env2 = [1, 2, 3]
 Array[String] env2_param = prefix("-f ", env2) # ["-f 1", "-f 2", "-f 3"]
 ```
 
+## X select_first(Array[X?])
+
+:pig2: [Cromwell supported](https://github.com/broadinstitute/cromwell#wdl-support) :white_check_mark:
+
+Given an array of optional values, `select_all` will select the first defined value and return it. Note that this is a runtime check and you should be confident that a defined value will exist: if no defined value is found when select_first runs, the workflow will fail.
```
ruchim: replace select_all with select_first

SPEC.md
```diff
@@ -2566,6 +2580,26 @@ Array[Integer] env2 = [1, 2, 3]
...
+
+Given an array of optional values, `select_all` will select only those elements which are defined.
+
+## Boolean defined(X?)
+
+This function will return `false` if the argument is an unset optional value. It will return `true` in all other cases.
+
+:pig2: [Cromwell supported](https://github.com/broadinstitute/cromwell#wdl-support) :white_check_mark:
+
+Given an array of optional values, `select_all` will select only those elements which are defined.
```
ruchim: This is the same as line 2593.

ruchim:
:+1: with some small adjustments

katevoss:
@vdauwera could you review this and merge if it's ready to go? This was forgotten last Cromwell release, whoops!

SPEC.md
vdauwera: Considering the Pair contains two ways to express the quantity in question, I'm going to assume it was intentionally pluralized.

SPEC.md
```diff
@@ -2488,7 +2502,7 @@ Example 2:
   }
 
   output {
-    File outputFile = ${output_file_name}
+    File outputFile = output_file_name
```
vdauwera: Wait, is this a new syntax for outputs? And does this mean "you can write it either way" or "you have to write it the new way"?

Apologies if this was specified in release notes that I failed to read.

SPEC.md
vdauwera: "you should be confident that a defined value will exist" -> "requires that at least one defined value exists at the time the check is made" ?

vdauwera:
Back to @cjllanwarne

SPEC.md
cjllanwarne: It's been a while since I wrote it but I'm happy to agree with @vdauwera that this is indeed a pair of twenty-threes and will therefore leave it as-is

SPEC.md
cjllanwarne: AFAIK, this has always been the correct syntax for outputs:

 - `output_file_name` is a variable, and you're assigning its value to `outputFile`.
 - `${output_file_name}` is a string interpolation that you can use inside a command block. I don't know that it would work outside a command block and would argue that it shouldn't.
 - You might be able to do something like `= "${output_file_name}"` but this is a needless indirection. You'd be saying "insert this String into another String, and then use that". It works, but you might as well use the original String in the first place

cjllanwarne:
@vdauwera @ruchim @katevoss for your consideration: https://github.com/broadinstitute/wdl/pull/98/commits/0514cec6edbcb9c341d4902a13c5f36fb211e459

ruchim:
Saw the changes of the second commit, looks beautiful :100:

PR #100, opened 2017-03-26
if then else expressions

cjllanwarne:
Allows expressions of the form:
```
Boolean b = ...

Int memory = if b then "2GB" else "5GB"
String docker = if b then "ubuntu:latest" else "my:docker"
```

### DIFF
--- a/grammar.hgr
+++ b/grammar.hgr
@@ -25,6 +25,8 @@ grammar {
     r'output(?!{%_identifier_follow%})' -> output(:output)
     r'as(?!{%_identifier_follow%})' -> :as
     r'if(?!{%_identifier_follow%})' -> :if
+    r'then(?!{%_identifier_follow%})' -> :then
+    r'else(?!{%_identifier_follow%})' -> :else
     r'while(?!{%_identifier_follow%})' -> :while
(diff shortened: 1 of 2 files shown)


grammar.hgr
```diff
@@ -49,6 +51,8 @@ grammar {
       javascript: "'([^\\\\\\'\\n]|\\\\[\\\"\\'nrbtfav\\\\?]|\\\\[0-7]{1,3}|\\\\x[0-9a-fA-F]+|\\\\[uU]([0-9a-fA-F]{4})([0-9a-fA-F]{4})?)*'"
     } -> wdl_unescape(:string)
 
+    r'else' -> :else
+    r'then' -> :then
```
cjllanwarne: Actually suspect this wouldn't be necessary... but if it is, then it's probably also required in the output expression section too

geoffjentry:
don't forget to update the spec

katevoss:
And the changelog :)

katevoss:
@knoblett and @vdauwera there are updates to the Spec, looks good to me but it would be great to get your review too.

Horneth:
👍

vdauwera:
👍

PR #104, opened 2017-04-20
Generic pre-processing and GVCF generation WDLs

vdauwera:
This PR adds two WDLs:

1. A generic, stripped-down version of the Broad production / GOTC single sample pipeline (no QC steps included). Starts with unmapped bams per readgroup and ends with HaplotypeCaller GVCF generation. Inputs requirements are relaxed to be reference-agnostic (two jsons provided, for hg38 and b37) and can be applied to exomes instead of genomes by swapping out intervals files and specifying interval padding (which is important for all exome target lists but unnecessary for the Broad's WGS intervals list).

2. Same as above but without the GVCF generation, to serve as common pre-processing script for all variant discovery pipelines.

Note that the BQSR tasks have been tweaked to use all GATK3 tools instead of the more performant GATK4 tools, for benchmarking purposes.


vdauwera:
This was the workflow we used for demo purposes at BioIT

Merging so we can have it on record.

PR #114, opened 2017-06-06
Adding to Pair docs

ruchim:


geoffjentry:
👍

SPEC.md
```diff
@@ -626,12 +633,19 @@ Map[String, Int] = {"a": 1, "b": 2}
...
+Pair values can be specified inside of a WDL using another Python-like syntax, as follows:
 
 ```
 Pair[Int, String] twenty_threes = (23, "twenty-three")
 ```
 
+Pair values can also be specified within the [workflow inputs JSON](https://github.com/broadinstitute/wdl/blob/develop/SPEC.md#specifying-workflow-inputs-in-json) with a `Left` and `Right` value specified using JSON style syntax. For example, given a workflow `wf_hello` and workflow-level variable `twenty_threes`, it could be declared in the workflow inputs JSON as follows:
+```
+{
+  "wf_hello.twenty_threes": { "Left": 23, "Right": "twenty-three" }
```
cjllanwarne: Are those `Left` and `Right` capitalizations correct? I think in the output it's all lowecase?

cjllanwarne:
Minor nitpick, but otherwise 👍

SPEC.md
ruchim: Json to Pair coercion isn't sensitive towards capitalization (there's a test case in WDL4S for it) but I didn't realize the outputs were all lowercase, that creates a slight inconsistency. We could make it required lowercase for the future but currently Pair creation accepts "Left" or "left"

sooheelee:
Given the meaning of indexing in genomics (BWA indexing, BAM indexing, VCF indexing), this section reads confusingly for me. Also, it's unclear how a researchers would want to use this feature.

ruchim:
@sooheelee I see--alternatively, I could move the pair member access syntax directly inside of the Pair Literals section.

vdauwera:
The "indexing" jargon overlap is kind of inevitable -- and I don't think the burden of clarifying should be on every WDL technical doc that uses the term (that way lies some amount of madness). For one, it's a term that people familiar with scripting in bioinformatics should already understand. So the problem is really only for newbs. Obviously we want to make sure we help them navigate this stuff, so I would advocate for writing a doc about terminology for researchers looking at scripting for the first time. And when we write up a description of the Pair object for the user docs proper (WDL website/forum) we can make sure to do so in a way that is unambiguous and well furnished with domain-relevant examples.

PR #121, opened 2017-06-30
Floor, Round and Ceil docs

cjllanwarne:


ruchim:
👍 👍 👍

SPEC.md
```diff
@@ -2640,6 +2641,15 @@ This function will return `false` if the argument is an unset optional value. It
 - This function returns the basename of a file path passed to it: `basename("/path/to/file.txt")` returns `"file.txt"`.
 - Also supports an optional parameter, suffix to remove: `basename("/path/to/file.txt", ".txt")` returns `"file"`.
 
+## Int floor(Float), Int ceil(Float) and Int round(Float)
+
+:pig2: [Supported in Cromwell 28](https://github.com/broadinstitute/cromwell#wdl-support) :white_check_mark:
```
geoffjentry: This is a departure from `Cromwell Supported` everywhere else. I get why you're doing this, but I don't think it's necessary. Other features are only supported in version `> x` but don't get the same text.

geoffjentry:
A slight bit of this

download

followed by a 👍

SPEC.md
mcovarr: Users have explicitly asked for this.  Up to now our docs have just said something works, but when users go to FireCloud it might not because FC may be a version behind.

geoffjentry: So how do we handle a situation where behavior changes between two versions, e.g. `write_x`?

geoffjentry: Also I'd suggest that the wdl spec shouldn't be the ultimate source of "what can I do in FC", but that's neither here nor there

geoffjentry: Although I suppose that since supposedly we're both going to formalize the spec (which I don't understand why it'd have cromwell stuff in it at all at that point) *and* better wdl docs over the next couple of months I don't care that much. 

whether or not I believe that'll be reality is another matter :)

cjllanwarne: Hmm, this is actually copypasta. Whether `Cromwell supports` belongs in a spec is another issue but since it's here at all. I think that saying which version adds value.

PR #125, opened 2017-08-10
Add Best Practices workflow for calling somatic SNVs and Indels with M2 (b37)

vdauwera:
Tested on Cromwell 27

Mirrors FireCloud workspace https://portal.firecloud.org/#workspaces/help-gatk/Somatic_variant_discovery_b37


vdauwera:
Added to separate repository here: https://github.com/gatk-workflows/somatic-snvs-indels

Moving the GATK workflows out for better versioning/distribution via platforms like Dockstore

PR #131, opened 2017-09-14
New version of the prod single-sample pipeline super-optimized for GCP

vdauwera:
Not yet running in production but has been validated and merged. This pipeline, formerly named `PublicPairedSingleSampleWf_*`, takes per-sample readgroup unmapped bams (uBams) and takes them through mapping and pre-processing then finally HaplotypeCaller GVCF generation. This new version runs a lot cheaper than before on GCP due to some neat optimizations that involve autosizing runtime resource requirements. 

Still running tests with the public inputs json; there are some subtleties due to a couple of QC steps that need to be skipped or otherwise handled gracefully in the absence of some private internal resources. 

DO NOT MERGE UNTIL TESTS PASS


vdauwera:
Ugh I swapped the docker for CheckContamination but didn't check the executable path. Need to fix that in the WDL.

vdauwera:
Back to running small-scale tests; run started at 13:34

vdauwera:
Success! 16:53 -> 3:20 duration

vdauwera:
Alright, cleaned up my commits and added some caveats to the readme. Don't have a proper doc for the full pipeline yet but that can follow later; all of the important stuff is the same as in previous versions so people can check out the archive.

vdauwera:
This should be good to go. Will merge in the morning after one more look at the docs I added.

PR #135, opened 2017-09-19
Reorganized the workflows and added better versioning

vdauwera:
Functional reorg should make our workflows more easily navigable. Versioning through numbered directories should ensure that going forward, links used in repositories will not break when the next version comes out. This is a prerequisite to registering workflows in Dockstore.


danbills:
(suggestion: ) @kshakir has tuned me into ["semantic versioning"](http://semver.org/) which would make all these version numbers 1.0.0

vdauwera:
I'm not opposed to it in principle, but I'm not sold on the need for the third level of granularity. Could you outline briefly how you think it should be applied and how it would help?

danbills:
Quoting the spec:

> PATCH version when you make backwards-compatible bug fixes

I doubt it would be used often.  I think you could use it like cromwell and just omit trailing 0's.

Again, not super important, just thought it was worth a look.

vdauwera:
That was actually quite an interesting read, thanks for linking to it. Now I'm thinking it probably would make sense to also include a patch number, just to be safe in case we need it down the road. That would allow us to fix typos or amend documentation in the WDLs without disrupting the major/minor numbering. Thanks for bringing this up!

vdauwera:
Converted to semantic versioning; looking more closely at the early versions of the prod pipelines they were all just patch versions, so that's already come in handy. Thanks again @danbills

PR #138, opened 2017-09-23
Updated revert-to-ubam script

vdauwera:
Running tests via wdl_runner and in FireCloud

Do not merge until tests are confirmed passing.


vdauwera:
All pass provided they're given enough disk space

PR #140, opened 2017-10-07
Increase file size limits of the wdl_runner Cromwell

vdauwera:
Cromwell imposes default limits on the size of files that can be read in via readlines() which are quite constraining when dealing with some genomic cases including big lists of genomic intervals (I encountered the problem in the context of joint genotyping) or files of file names (FoFNs). They are meant to protect Cromwell from running out of memory from being asked to read in large files, which could occur especially if that is done in multiple workflows at the same time. Since this Cromwell only ever runs one workflow at a time it seems innocuous to increase the file size limits a bit to at least clear the threshold needed for the production joint genotyping WDL. 

The underlying issue will be addressed in Cromwell in a more principled way but in the meantime we need to be able to run this workflow.

### DIFF
--- a/runners/cromwell_on_google/wdl_runner/Dockerfile
+++ b/runners/cromwell_on_google/wdl_runner/Dockerfile
@@ -50,10 +50,10 @@ RUN mkdir /cromwell
 RUN cd /cromwell && \
     curl -L -O https://github.com/broadinstitute/cromwell/releases/download/29/cromwell-29.jar
 RUN ln /cromwell/cromwell-29.jar /cromwell/cromwell.jar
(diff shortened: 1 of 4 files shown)


patmagee:
Instead of making these changes here, would it make sense to extract this component into its own git repo first?

geoffjentry:
@patmagee I've left this here as @vdauwera has been out of office more than in office since the great upheaval, so i figured i'd give her time to figure out what to do with this :)

vdauwera:
Yeah I'm going to kill this PR bc it's not a good idea anyway, and will think about what to do about the runner.

PR #148, opened 2017-10-25
Add explicit failure when read_x() functions not completely handled

geoffjentry:
**Note that the ultimate result deviated from the original**

Makes it explicit that the inability to read the entire contents from a `read_x()` function (e.g. `read_lines`) results in the failure of the calling task or workflow. 

**Original**
This was something we meant to add to the spec as part of [Cromwell #1762](https://github.com/broadinstitute/cromwell/issues/1762) and forgot, thus it's not part of the spec. Carrying it forward as-is for now.

Explanations for the amounts. 

The ones w/ explicit character counts were based on maximum valid values. 

The rest were based on what seemed "reasonable" when trying to balance "an individual is going to want this to be as large as possible" vs "a server is going to want to not read in gigantic files constantly". For `read_lines` in particular we took some known large examples, multiplied it by a few times, and then rounded. However I'll note that we've already had requests for that particular value to be a couple of orders of magnitude larger.

### DIFF
--- a/versions/draft-3/SPEC.md
+++ b/versions/draft-3/SPEC.md
@@ -2093,7 +2093,7 @@ task do_stuff {
 }
 ```
 
-An implementation of the `read_lines()` function must read at least 128k of data from the file but may read more. Users attempting to read more than this can run into portability issues across implementations. 
+If the entire contents of the file can not be read for any reason, the calling task or workflow will be considered to have failed. 
 
 ## Array[Array[String]] read_tsv(String|File)
 
@@ -2119,8 +2119,7 @@ task do_stuff {
 
 Then when the task finishes, to fulfull the `outputs_table` variable, `./results/file_list.tsv` must be a valid TSV file or an error will be reported.
 
-An implementation of the `read_tsv()` function must read at least 1M of data from the file but may read more. Users attempting to read more than this run into portability issues across implementations. 
-
+If the entire contents of the file can not be read for any reason, the calling task or workflow will be considered to have failed. 
 
 ## Map[String, String] read_map(String|File)
 
@@ -2145,7 +2144,7 @@ task do_stuff {
 }
 ```
 
-An implementation of the `read_map()` function must read at least 1M of data from the file but may read more. Users attempting to read more than this run into portability issues across implementations. 
+If the entire contents of the file can not be read for any reason, the calling task or workflow will be considered to have failed. 
 
 ## Object read_object(String|File)
 
@@ -2182,7 +2181,7 @@ Which would be turned into an `Object` in WDL that would look like this:
 |key_2    |"value_2"|
 |key_3    |"value_3"|
 
-An implementation of the `read_object()` function must read at least 1M of data from the file but may read more. Users attempting to read more than this run into portability issues across implementations. 
+If the entire contents of the file can not be read for any reason, the calling task or workflow will be considered to have failed. 
 
 ## Array[Object] read_objects(String|File)
 
@@ -2229,7 +2228,7 @@ Which would be turned into an `Array[Object]` in WDL that would look like this:
 |     |key_2    |"value_2"|
 |     |key_3    |"value_3"|
 
-An implementation of the `read_objects()` function must read at least 1M of data from the file but may read more. Users attempting to read more than this run into portability issues across implementations. 
+If the entire contents of the file can not be read for any reason, the calling task or workflow will be considered to have failed. 
 
 ## mixed read_json(String|File)
 
@@ -2264,7 +2263,7 @@ task do_stuff {
 
 Then when the task finishes, to fulfull the `output_table` variable, `./results/file_list.json` must be a valid TSV file or an error will be reported.
 
-An implementation of the `read_json()` function must read at least 128k of data from the file but may read more. Users attempting to read more than this run into portability issues across implementations. 
+If the entire contents of the file can not be read for any reason, the calling task or workflow will be considered to have failed. 
 
 ## Int read_int(String|File)
 
@@ -2272,7 +2271,7 @@ An implementation of the `read_json()` function must read at least 128k of data 
 
 The `read_int()` function takes a file path which is expected to contain 1 line with 1 integer on it.  This function returns that integer.
 
-An implementation of the `read_int()` function must read at least 19 characters of data from the file but may read more. Users attempting to read more than this run into portability issues across implementations. 
+If the entire contents of the file can not be read for any reason, the calling task or workflow will be considered to have failed. 
 
 ## String read_string(String|File)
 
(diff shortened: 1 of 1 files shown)


cjllanwarne:
- See my reply in the openwdl group for [my hesitations](https://groups.google.com/a/openwdl.org/forum/#!topic/community/Msrtn2n9a7U) on putting this in the spec. In brief, I don't really like the idea of the language getting involved in setting implementation limits, even minimums, for engines. In my opinion WDL should pretend that these are infinite and if engines cannot handle a file input then that's a failure. 
- If we *do* want minimums in the WDL language spec then we might want say what happens when an engine attempts to read bigger files? Eg something like `"if an input file is bigger than the specified minimum and if the engine decides that it cannot or will not read a file of that size, then it will fail the workflow"`

geoffjentry:
TBH this wasn't a I-think-this-is-exactly right but rather something we intended to have it in the spec a few months ago so I just carried it forth as-is to get this ball rolling, as well as the "community driven" angle ball rolling!

In general I agree with you, although IIRC the [argument for minimums](https://github.com/broadinstitute/cromwell/issues/1762) in the first place was that at least with minimums a user can guarantee portability by hacking in different behavior if they know they're going to run afoul of things. This is in contrast to maximums or allowing engines to back off because now for any given implementation you don't know if it'll possibly work.

I do like the idea you raised in your email of defining an inability to read the entire thing as a failure.

vdauwera:
I understand the concern about language spec overreach but really like the idea of a minimum minimorum to guarantee portability.

Not sure I understand the "defining an inability to read the entire thing as a failure" point as it would seem obvious to me that if the engine can't read the whole thing it should fail that task or workflow but maybe I'm missing a subtlety somewhere.

geoffjentry:
@vdauwera I don't really agree that having a minimum provides portability. Having a **maximum** would for sure. On paper i see why the minimums are useful although I'll note that what we're already seeing in practice are users demanding the limits be raised (to absurd levels) and not the intended behavior of "find a way to not need `read_X()`". 

If you feel that in general an author would know whether or not their task could be written in a way that'd always wind up below the minimum that'd change my mind a bit. What I mean is, this was originally sold as "As an author if I knew that I might wind up above the limit I could write my tool differently if I wanted to be portable" - is that scenario really going to be common? Legit question, not me pushing back.

"would seem obvious to me that if the engine can't read the whole thing"

I might have misunderstood the CWL spec but I thought over there it only guarantees reading up to X amount and then moves on with life, so if that statement is true it's at least not obvious to someone.

vdauwera:
Oh hmm to be honest I didn't give it a huge amount of thought beyond "oh it would make sense to specify a reasonable range within which stuff is guaranteed to work everywhere". 

As a WDL author I would prefer not to have to think about this at all. I'm probably using FoFNs because I saw them used elsewhere and I don't realize there may be other ways to achieve the same result. I'm probably not even thinking about a FoFN as a thing that could exceed a file size limit in the first place, because the Fs in question are gigantic bam files and those are the sizes I'm used to worrying about. 

Not sure the above helps you at all but that's my expectation of reality. If we have good guidelines and alternatives for achieving the intended result then we can steer people through documentation and tutorials of course. 

Re: that last thing, are you saying that "only guarantees reading up to X amount and then moves on with life" miggt mean that the engine is allowed to just read part of the file then try to move on with execution, without telling me that it might not have read the whole thing? If so that's horrifying.

geoffjentry:
Actually I just realized I spread the same mistruth in the old ticket. They have a max limit

geoffjentry:
@vdauwera If this helps guide your opinion at all, we *have* run into people complaining about these limits when they're trying to use FOFNs. A local user who shall remain nameless wanted the limit to be upped to 5GB (!!!!) for FOFN usage.

vdauwera:
Hah, I bet I know who that is :)

To be clear I’ve run into those limits myself, and have fiddled with the settings on my own Cromwell to get past them. It took me by surprise because I had not been thinking about it as a problem that could affect me, even though I was aware of others tripping on the issue. It’s definitely something I expect will be a sore point with some users. Having clearly defined min and/or max will help, as would a more informative error message when the limit is reached. Also, the hypothetical “dry run” feature that would test all inputs that are provided up front for existence/permissions and suitability (eg size limits). I think I’ve asked for this already but can make a formal request if it hasn’t made it into the official List of Things People Want. Was heavily requested at our recent workshop too btw.

patmagee:
@vdauwera @geoffjentry minimum / maximum values make sense to a certain degree. But I would like to add my voice to the dissenting opinions here.  Defining a minimum does not guarantee portability as @geoffjentry mentioned, while defining the maximum would. However I feel like this issue is something that should be soleved at the level of execution engine. I could forsee exection engines which are designed and function in different ways allowing different limits simply because the supporting hardware and software are built to support it. 

Restricting this in the core language does not seem like the best approach. What has been said, which I do agree with, is that we should unify the error message that is returned when a size limitation is encountered.

geoffjentry:
@vdauwera "dry run" seems more a cromwell thing than a wdl spec thing, at least to me

@patmagee FWIW despite me having opened this PR I agree more with you and @cjllanwarne - since at least based on the opinions here that seems to be where things are trending I'll likely update the PR and publicize it a second time unless I hear a groundswell of counteropinions

geoffjentry:
@patmagee (and everyone else) - I had a conversation with @cjllanwarne and he raised what I thought was a good point. We were discussing a world where there were no limits (min or max) in the language spec and instead having prescribed failure w/ standardized error messages. 

He suggested codifying simply failure to read as a workflow failure and leaving standardization of error message to something like WES. I disagreed at first but he swung me around - if you view the general class of "failed due to resource limitations" that does seem like a WES thing to me.

geoffjentry:
To move this a bit I've updated the PR to simply note that any case where the entire contents of a file can't be read will mark the task/workflow as a failure

patmagee:
This change looks good to me.  Defining the explicit scenario where it should fail gives more portability in a way. We cannot guarantee that the execution engine will be able to read the specific file without reaching a limit, but we can guarantee that when it does read it, the entire contents of the file is read.

+1 to this

geoffjentry:
I'm going to add a couple of concrete examples today and then call for votes

briandoconnor:
+1

yfarjoun:
👍

patmagee:
:+1:

mlin:
LGTM

versions/draft-3/SPEC.md
```diff
@@ -2267,18 +2281,24 @@ The `read_string()` function takes a file path which is expected to contain 1 li
...
+If the entire contents of the file can not be read for any reason, the calling task or workflow will be considered to have failed. Examples of failure include but are not limted to not having access to the file, resource limitations (e.g. memory) when reading the file, and implementation imposed file size limits.
+
 ## Float read_float(String|File)
 
 :pig2: [Cromwell supported](https://github.com/broadinstitute/cromwell#wdl-support) :white_check_mark:
 
 The `read_float()` function takes a file path which is expected to contain 1 line with 1 floating point number on it.  This function returns that float.
 
+If the entire contents of the file can not be read for any reason, the calling task or workflow will be considered to have failed. 
+
```
Horneth: Did you mean to leave the examples out of this one ?

geoffjentry:
With a vote of 12-0 this PR has been approved. As there's already an implementation doing this (Cromwell), it will be merged.

PR #149, opened 2017-10-25
Remove references to Cromwell

geoffjentry:


patmagee:
+1 for me as well. 

What would be really great down the road would be a compliance test sweet which could ensure feature compliance.

Implementations should have be responsible for informing it's users which engine functions are supported

geoffjentry:
Yes, this wasn't meant to be in lieu of something more formal across the board, rather that was just not in the cards for the 5 minutes I had dedicated to working on this bit :P

PR #153, opened 2017-11-05
Array flatten function

orodeh:
Currently, there is no library function to flatten an array of array of files (Array[Array[File]]). A scatter, where each task call produces an array of files, is a natural way of ending up with such a structure. In order to flatten this array, you can write a task that takes the it as an argument, and manipulate it with python code. However, this task will also download all the files, taking significant time and disk space. To work around this, you can coerce the files into strings (their paths), and manipulate the paths.

You can see an example [here](https://github.com/HumanCellAtlas/skylab/blob/master/10x/count/count.wdl#L195). The `chunk_reads_join` task flattens the `fastq_chunks` file array, which is coerced into an `Array[Array[String]]`.

In order to avoid this circuitous implementation, this pull requests suggest a standard library function instead.


geoffjentry:
You'll want to target [draft-3](https://github.com/openwdl/wdl/blob/master/versions/draft-3/SPEC.md)

SPEC.md
```diff
@@ -2601,6 +2602,20 @@ Integer ylen = length(ys) # 3
 Integer zlen = length(zs) # 0
 ```
 
+## Array[X] flatten(Array[Array[X]])
+
+Given an array of arrays, the `flatten` function concatentes all the
```
geoffjentry: I'm assuming this won't de-dup, can you make that clear? Or note that it does if I'm wrong.

cjllanwarne: Typo on `concatenAtes`

SPEC.md
```diff
@@ -2601,6 +2602,20 @@ Integer ylen = length(ys) # 3
...
+member arrays in the order to appearance to give the result. For example:
+
+```
+Array[Array[Integer]] ai2D = [[1, 2, 3], [11], [21,22]]
+Array[Integer] ai = flatten(ai2D)   # [1, 2, 3, 11, 21, 22]
+
+
+Array[Array[File]] af2D = [["/tmp/X.txt"], ["/tmp/Y.txt", "/tmp/Z.txt"], []]
+Array[Integer] af = flatten(af2D)   # ["/tmp/X.txt", "/tmp/Y.txt", "/tmp/Z.txt"]
```
cjllanwarne: Does this also work with `Array`-likes? eg maps?
```wdl
Array[Map[Int, String]] array_of_maps = [ {1: "one"}, {2: "two", 3: "three"} ]
Array[Pair[Int, String]] array_of_pairs = flatten(array_of_maps) # [ (1, "one"), (2, "two"), (3, "three") ]
```

cjllanwarne: To be honest, I'm not sure how the second example is different from the first?

SPEC.md
```diff
@@ -2601,6 +2602,20 @@ Integer ylen = length(ys) # 3
 Integer zlen = length(zs) # 0
 ```
 
+## Array[X] flatten(Array[Array[X]])
+
+Given an array of arrays, the `flatten` function concatentes all the
+member arrays in the order to appearance to give the result. For example:
```
cjllanwarne: As written here, I think you'd need to call it twice to flatten an `Array[Array[Array[Int]]]`, but it might be nice for it to totally flatten any number of array nestings?

SPEC.md
cjllanwarne: Also should not be `Integer`

SPEC.md
geoffjentry: Agree w/ @cjllanwarne - my assumption would be that `flatten` would return a completely flattened array

SPEC.md
cjllanwarne: I agree - "No du-dup" is the correct answer IMO (since they're arrays, not sets)

orodeh:
Changed this pull request to be against master branch (draft-3), instead of the develop branch.

geoffjentry:
@orodeh I think something went awry as now you're pulling in half the world's changes. Might be fixable with a rebase.

orodeh:
I have a really bad experience with git rebase. I can fix the comments, open a new PR, and reference the old one (and vice versa).

geoffjentry:
Sounds good

SPEC.md
orodeh: This is certainly possible, although, that was not my original intention. I wanted to have a function like [flatten](http://www.scala-lang.org/api/2.12.3/scala/collection/immutable/List.html#flatten[B]:List[B]) in the scala standard library, or ocaml's [List.flatten](https://caml.inria.fr/pub/docs/manual-ocaml/libref/List.html). If we completely flatten all nesting levels, we would need another function for flattening only one nesting level, no?

geoffjentry: Actually my assumption was based on a bad memory of scala, I was thinking it fully flattened. You can see how much I have hands on keyboard these days :P

My main point stands either way though, I think it's worth being clear.

SPEC.md
orodeh: Honestly, I did not consider what would happen if a user applied `flatten` to an array-like structure. It is a useful extension though, in my opinion. If we use the scala [flatten](http://www.scala-lang.org/api/2.12.3/scala/collection/immutable/List.html#flatten[B]:List[B]) semantics, we would get:
```
Array[Map[Int, String]] array_of_maps = [ {1: "one"}, {2: "two", 3: "three"} ]
Map[Int, String] m = flatten(array_of_maps) # {1: "one", 2: "two", 3: "three"}
```

orodeh: I wanted to show an example with an empty `Array` the is eliminated, and also show that it works on `File`s.

orodeh:
This was rebased against master branch, and draft-3 of the specification. See [this pull request](https://github.com/openwdl/wdl/pull/154/files)

orodeh:
Since this pull request has been moved, please comment on the [new request](https://github.com/openwdl/wdl/pull/154).

orodeh:
Moved to https://github.com/openwdl/wdl/pull/154

PR #154, opened 2017-11-06
Array flatten function (v2)

orodeh:
This is a cleanup of this [pull request](https://github.com/openwdl/wdl/pull/153)

Currently, there is no library function to flatten an array of array of files (Array[Array[File]]). A scatter, where each task call produces an array of files, is a natural way of ending up with such a structure. In order to flatten this array, you can write a task that takes the it as an argument, and manipulate it with python code. However, this task will also download all the files, taking significant time and disk space. To work around this, you can coerce the files into strings (their paths), and manipulate the paths.

You can see an example here. The chunk_reads_join task flattens the fastq_chunks file array, which is coerced into an Array[Array[String]].

In order to avoid this circuitous implementation, this pull requests suggest a standard library function instead.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -2608,7 +2608,9 @@ Integer zlen = length(zs) # 0
 
 Given an array of arrays, the `flatten` function concatenates all the
 member arrays in the order to appearance to give the result. It does not
-deduplicate the elements. For example:
+deduplicate the elements. Arrays nested more deeply than 2 must be
+flattened twice (or more) to get down to an unnested `Array[X]`.
+For example:
 
 ```
 Array[Array[Integer]] ai2D = [[1, 2, 3], [1], [21, 22]]
@@ -2620,6 +2622,8 @@ Array[File] af = flatten(af2D)   # ["/tmp/X.txt", "/tmp/Y.txt", "/tmp/Z.txt"]
 Array[Array[Pair[Float,String]]] aap2D = [[(0.1, "mouse")], [(3, "cat"), (15, "dog")]]
 Array[Pair[Float,String]] ap = flatten(aap2D) # [(0.1, "mouse"), (3, "cat"), (15, "dog")]
 ```
+
+The last example (`aap2D`) is useful because `Map[X, Y]` can be coerced to `Array[Pair[X, Y]]`.
 
 ## Array[String] prefix(String, Array[X])
 
--- a/SPEC.md
+++ b/SPEC.md
@@ -2618,7 +2618,7 @@ Array[Array[File]] af2D = [["/tmp/X.txt"], ["/tmp/Y.txt", "/tmp/Z.txt"], []]
 Array[File] af = flatten(af2D)   # ["/tmp/X.txt", "/tmp/Y.txt", "/tmp/Z.txt"]
 
 Array[Array[Pair[Float,String]]] aap2D = [[(0.1, "mouse")], [(3, "cat"), (15, "dog")]]
-Array[Pair[Float,String]] ap = [(0.1, "mouse"), (3, "cat"), (15, "dog")]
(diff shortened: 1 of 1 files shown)


geoffjentry:
I mentioned it in the previous PR but for posterity over here I like this proposal. @orodeh - I'd encourage to try to see if anyone else wants to chime in and either when that dies down or if it never starts we can call for vote

geoffjentry:
@orodeh This has gone several days w/o commentary so I'm going to move it to voting now.

SPEC.md
```diff
@@ -2603,6 +2604,23 @@ Integer ylen = length(ys) # 3
...
+deduplicate the elements. For example:
+
+```
+Array[Array[Integer]] ai2D = [[1, 2, 3], [1], [21, 22]]
+Array[Integer] ai = flatten(ai2D)   # [1, 2, 3, 1, 21, 22]
+
+Array[Array[File]] af2D = [["/tmp/X.txt"], ["/tmp/Y.txt", "/tmp/Z.txt"], []]
+Array[File] af = flatten(af2D)   # ["/tmp/X.txt", "/tmp/Y.txt", "/tmp/Z.txt"]
+
+Array[Array[Pair[Float,String]]] aap2D = [[(0.1, "mouse")], [(3, "cat"), (15, "dog")]]
```
cjllanwarne: Maybe note that this is useful because `Map[X, Y]` can be coerced to `Array[Pair[X, Y]]`

SPEC.md
```diff
@@ -2603,6 +2604,23 @@ Integer ylen = length(ys) # 3
 Integer zlen = length(zs) # 0
 ```
 
+## Array[X] flatten(Array[Array[X]])
+
+Given an array of arrays, the `flatten` function concatenates all the
+member arrays in the order to appearance to give the result. It does not
+deduplicate the elements. For example:
```
cjllanwarne: Comment that arrays nested more deeply than 2 must be flattened twice (or more) to get down to an unnested `Array[X]`

geoffjentry:
@orodeh Via a quirk of timing @cjllanwarne managed to submit comments right as I opened for voting. I think if you want to/can address those prior to other people submitting votes it'll be fine to do so.  Otherwise we'll need to track down voters to make sure they still vote the same way with whatever changes are in effect.

cjllanwarne:
I was just looking at this, apparently in parallel! Maybe a "last comments, please" warning 24h before voting starts is a good idea for future PRs?

Regardless, I vote 👍 pending the small clarifications I mentioned in the comments.

orodeh:
@cjllanwarne I added your comments into the text, to clarify the semantics and usage of the `flatten` function. Thanks!

patmagee:
👍

antonkulaga:
+1

aprabhak2:
+1

chapmanb:
+1 -- this is really nice functionality. In my CWL workflows I often find myself needing this when moving back and forth between parallized batches (like tumor/normal calling) and single sample processing.

geoffjentry:
By a vote of 6-0 this PR passes. It will remain open until an implementation is realized. I know there's a [Cromwell PR](https://github.com/broadinstitute/cromwell/pull/2825) which will presumably be merged in soon, @orodeh let me know if dxWDL (or an impl I'm not aware of) already supports this.

orodeh:
This is already in Cromwell. It has been for a while now (wdl/src/main/scala/wdl/expression/WdlStandardLibraryFunctions.scala).

geoffjentry:
@orodeh At least its existence post-dates my comment :) 

Merging.

geoffjentry:
@orodeh This was merged nearly 2 months ago :)

orodeh:
@geoffjentry That was what I thought, but it still carries the label "Waiting for Implementation".

PR #159, opened 2017-11-15
Fixup Types section and include 'Pair's

cjllanwarne:
- Makes the Types section more readable
- Removes the out of date `grammar` box
- Corrects a comment on the behavior of optionals.
- Includes the Pair type (since it's used elsewhere throughout the SPEC, it's probably worth at least mentioning it here!)

EDIT: link to this change rendered in situ: https://github.com/cjllanwarne/wdl/blob/cjl_types_fixup/versions/draft-3/SPEC.md#types

### DIFF
--- a/versions/draft-3/SPEC.md
+++ b/versions/draft-3/SPEC.md
@@ -250,10 +250,10 @@ String s = "hello, world"  # A string value
 File f = "/path/to/file"   # A file
 ```
 
-In addition, the following compound types can be constructed, parameterized by other types (in the examples below `X` and `Y` represent any valid type, even nested compound types):
+In addition, the following compound types can be constructed, parameterized by other types. In the examples below `P` represents any of the primitive types above, and `X` and `Y` represent any valid type (even nested compound types):
 ```wdl
 Array[X] xs = [x1, x2, x3]                    # An array of Xs
-Map[X,Y] x_to_y = { x1: y1, x2: y2, x3: y3 }  # A map from Xs to Ys
+Map[P,Y] p_to_y = { x1: y1, x2: y2, x3: y3 }  # A map from Ps to Ys
 Pair[X,Y] x_and_y = (x, y)                    # A pair of one X and one Y
 Object o = { "field1": f1, "field2": f2 }     # Object keys are always `String`s
 ```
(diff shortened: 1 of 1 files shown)


versions/draft-3/SPEC.md
```diff
@@ -240,30 +240,37 @@ $float = (([0-9]+)?\.([0-9]+)|[0-9]+\.|[0-9]+)([eE][-+]?[0-9]+)?
...
-$primitive_type = ('Boolean' | 'Int' | 'Float' | 'File' | 'String')
-$array_type = 'Array' '[' ($primitive_type | $object_type | $array_type) ']'
-$object_type = 'Object'
-$map_type = 'Map' '[' $primitive_type ',' ($primitive_type | $array_type | $map_type | $object_type) ']'
-$type_postfix_quantifier = '?' | '+'
+
+In addition, the following compound types can be constructed, parameterized by other types (in the examples below `X` and `Y` represent any valid type, even nested compound types):
+```wdl
+Array[X] xs = [x1, x2, x3]                    # An array of Xs
+Map[X,Y] x_to_y = { x1: y1, x2: y2, x3: y3 }  # A map from Xs to Ys
```
patmagee: In this example here, `X` Does not represent any Valid Type but a `primitive_type` as defined by hermes. So the constraint on X is that it must be: `Boolean`, `Int`, `Float`, `File`, or `String`. But the statement on line 253 would lead me to think that `X` can be a compound type as well, whish is false.

orodeh:
I don't know if this is the right place for comments, but I think that there are actually two types of file paths: 
1) local Posix files (/A/B/C)
2)URLs like (gs://..., dx:// ...). 

I am not sure, could you specify a file as an HTTP address?

geoffjentry:
@orodeh I think that's far enough off of this PR to treat it separately. Historically we've treated `File` as both, although there's been a lot of discussion swirling - e.g. the post on the mailing list about non-localizable files.

orodeh:
👍

geoffjentry:
:+1:

ruchim:
:+1:

geoffjentry:
This passed 3-1 and as it's just a clarification is mergable as-is

PR #160, opened 2017-11-16
Add Struct Construct to allow for Complex types

patmagee:
In a previous discussion in the Cromwell github page under this issue https://github.com/broadinstitute/cromwell/issues/2283 the concept of revamping the `Object` type was brought up. 

An `Object` Is supposed to be a JSON like type, however its functionally very hard to interact with and use. Type coercions for nested members in an object are never reliable, Caching is next to impossible, and doing actions such as localizing or delocalizing files from within the object are not supported. There was a clear desire for users to have a more complex JSON like with some sort of typing which the `Object` type did not support.

Enter the `Struct`.

A struct, similar to its C counterpart, is a construct which allows users to define their own Types dynamically, directly wihtin a WDL file by combining preexisting types into a single new named type.

For example, if I had a set of patients, and for each patient I had  several FASTQ files, the patient id, patient age, and a list of siblings for that patient,  there is no way to easily pass that information around from task to task. However with Structs this would be easy.

```wdl
struct Patient {
  Array[File]
  String id
  Int age
  Array[String] siblings 
}
```

And then if I wanted to use the new type in a workflow or task I could do the following:

```wdl
workflow myWorkflow {
  Array[Patient] patients

  scatter(patient in patients) {
      ....
  }
}
```

By using structs, file localization and delocalization should be supported, as well as being able to properly type each field present within the struct. This removes any sort of ambiguity that exists with the `Object` type

### DIFF
--- a/versions/draft-3/SPEC.md
+++ b/versions/draft-3/SPEC.md
@@ -274,6 +274,11 @@ For more details on the `$type_postfix_quantifier`, see the section on [Optional
 
 For more information on type and how they are used to construct commands and define outputs of tasks, see the [Data Types & Serialization](#data-types--serialization) section.
 
+#### Custom  Types
+
+WDL provides the ability to define custom compound types called `Structs`. `Structs` are defined directly in the WDL and are usable like any other type. 
+For more information on their usage, see the section on [Structs](#struct-definition)
+
 ### Fully Qualified Names & Namespaced Identifiers
 
 ```
@@ -3081,6 +3086,7 @@ Where `/jobs/564759/sample.json` would contain:
   }
 ]
 ```
+
 
 ## De-serialization of Task Outputs
 
--- a/versions/draft-3/SPEC.md
+++ b/versions/draft-3/SPEC.md
@@ -1668,15 +1668,17 @@ alias them as follows:
 import http://example.com/example.wdl as ex alias Experiment as OtherExperiment
 ```
 
-In order to resolve multiple structs, simply add additional alias statements.
+In order to resolve multiple structs, simply add additional alias statements. 
 ```wdl
-import http://example.com/another_exampl.wdl as ex2 alia Parent as Parent2 alias Child as Child2 alias GrandChild as GrandChild2
+import http://example.com/another_exampl.wdl as ex2 
+    alias Parent as Parent2 
+    alias Child as Child2 
+    alias GrandChild as GrandChild2
 ```
 
 Its important to note, that when importing from file 2, all structs from file 2's global namespace will be imported. This Includes structs from
-another imported WDL within file 2, even if they are aliased. If a strut is aliased in file 2, it will be imported into file 1 under its
+another imported WDL within file 2, even if they are aliased. If a struct is aliased in file 2, it will be imported into file 1 under its
 aliased name.
-
 
 
 * Note: Alias can be used even when no conflicts are encountered to uniquely identify any struct
--- a/versions/draft-3/SPEC.md
+++ b/versions/draft-3/SPEC.md
@@ -1515,6 +1515,104 @@ workflow wf {
 
 In this example, the fully-qualified names that would be exposed as workflow outputs would be `wf.task1.results`, `wf.altname.value`.
 
+# Structs
+A struct is a C-like construct which enables the user to create new compound types that consist of previously existing types. Structs
+can then be used within a `Task` or `Workflow` definition as a declaration in place of any other normal types. The struct takes the place of the
+`Object` type in many circumstances and enables proper typing of its members.
+
+
+## Defining a Struct
+Structs are defined using the `struct` keyword and have a single section consiting of typed declarationns. They are evaluated first prior
+to workflows and tasks.
+```
+struct name { ... }
+```
+
+## Defining Declarations
+The only contents of struct are a set of declarations. Declarations can be any primitive or compound type, as well as other structs, and are defined
+the same way as they are in any other section. The one caveat to this is that declartions within a struct do not allow an expression statement after
+the initial member declaration.
+
+for example the following is a valid struct definition
+``` 
+struct name {
+    String myString
+    Int myInt
+}
+```
+Whereas the following is invalid
+
+```
+struct invalid {
+    String myString = "Cannot do this"
+    Int myInt
+}
+```
+
+You can also use compound types within a struct to easily encapsulate them within a single object. For example
+```
+struct name {
+    Array[Array[File]] myFiles
+    Map[String,Pair[String,File]] myComplexType
+    String cohortName
+}
+```
+## Optional and non Empty Values
+Struct declarations can be optional or non-empty (if they are an array type). 
+
+```
+struct name {
+    Arrray[File]+ myFiles
+    Boolean? myBoolean
+}
+```
+
+## Using a Struct
+When using a struct in the declaration section of either a `workflow` or a `task` or `output` section you define them in the same way you would define any other type.
+
+For example, if I have a struct like the following:
+```
+struct MyStruct {
+    String myName
+    Int myAge
+}
+```
+
+then usage of the struct in a workflow would look like the following:
+
+```wdl
+
+task myTask {
+    MyStruct a
+    command {
+        echo "hello my name is ${a.myName} and I am ${a.myAge} years old"
+    }
+}
+
+workflow myWorkflow {
+    MyStruct a
+    call myTask {
+        input:
+            a = a
+    }
+
+}
+```
+
+## Member Access
+In order to access members within a struct, use object notation; ie `myStruct.myName`. If the uderlying member is a complex type which supports member access,
+you can access its elements in the way defined by that specific type.
+
+for example if we have a struct like the following:
+```
+struct myStruct {
+    Array[File] myFiles
+    Map[String,String] myMap
+}
+```
+
+Accessing the nth element of myFiles would look like: `myStruct.myFiles[n]`, while indexing an element in myMap would look like `myStruct.myMap["entry"]`
+
 # Namespaces
 
 Import statements can be used to pull in tasks/workflows from other locations as well as to create namespaces.  In the simplest case, an import statement adds the tasks/workflows that are imported into the specified namespace.  For example:
--- a/versions/draft-3/SPEC.md
+++ b/versions/draft-3/SPEC.md
@@ -1523,12 +1523,16 @@ workflow wf {
 In this example, the fully-qualified names that would be exposed as workflow outputs would be `wf.task1.results`, `wf.altname.value`.
 
 ## Struct Definition
-A struct is a C-like construct which enables the user to create new compound types that consist of previously existing types. Structs
+A struct is a C-like construct which enables the user to create new compound types that consisting of previously existing types. Structs
 can then be used within a `Task` or `Workflow` definition as a declaration in place of any other normal types. The struct takes the place of the
 `Object` type in many circumstances and enables proper typing of its members.
 
-Structs are defined using the `struct` keyword and have a single section consiting of typed declarationns. They are evaluated first prior
-to workflows and tasks.
+Structs are declared separately from any other constructs, and cannot be declared within any `workflow` or `task` definition. They belong to the namespace of the WDL 
+file which they are written in and are therefore accessible globally within that WDL. Additionally, all structs must be evaluated prior to their use within a `task`, 
+`workflow` or another `struct`.
+
+Structs may be defined using the `struct` keyword and have a single section consisting of typed declarations. 
+
 ```wdl
 struct name { ... }
 ```
@@ -1540,7 +1544,7 @@ the initial member declaration.
 
 for example the following is a valid struct definition
 ```wdl
-struct name {
+struct Name {
     String myString
     Int myInt
 }
@@ -1548,15 +1552,15 @@ struct name {
 Whereas the following is invalid
 
 ```wdl
-struct invalid {
+struct Invalid {
     String myString = "Cannot do this"
     Int myInt
 }
 ```
 
-You can also use compound types within a struct to easily encapsulate them within a single object. For example
+Compound types can also be used within a struct to easily encapsulate them within a single object. For example
 ```wdl
-struct name {
+struct Name {
     Array[Array[File]] myFiles
     Map[String,Pair[String,File]] myComplexType
     String cohortName
@@ -1566,8 +1570,8 @@ struct name {
 Struct declarations can be optional or non-empty (if they are an array type). 
 
 ```wdl
-struct name {
-    Arrray[File]+ myFiles
+struct Name {
+    Array[File]+ myFiles
     Boolean? myBoolean
 }
 ```
@@ -1577,9 +1581,9 @@ When using a struct in the declaration section of either a `workflow` or a `task
 
 For example, if I have a struct like the following:
 ```wdl
-struct MyStruct {
-    String myName
-    Int myAge
+struct Person {
+    String name
+    Int age
 }
 ```
 
@@ -1587,20 +1591,19 @@ then usage of the struct in a workflow would look like the following:
 
 ```wdl
 
-task myTask {
-    MyStruct a
+task task_a {
+    Person a
     command {
-        echo "hello my name is ${a.myName} and I am ${a.myAge} years old"
+        echo "hello my name is ${a.name} and I am ${a.age} years old"
     }
 }
 
 workflow myWorkflow {
-    MyStruct a
-    call myTask {
+    Person a
+    call task_a {
         input:
             a = a
     }
-
 }
 ```
 
@@ -1609,23 +1612,48 @@ Structs can be assigned using an object literal. When Writing the object, all en
 
 ```wdl
 
-MyStruct a = {"myName": "John","myAge": 30}
+Person a = {"name": "John","age": 30}
 
 ```
 
 
 ### Struct Member Access
-In order to access members within a struct, use object notation; ie `myStruct.myName`. If the uderlying member is a complex type which supports member access,
+In order to access members within a struct, use object notation; ie `myStruct.myName`. If the underlying member is a complex type which supports member access,
 you can access its elements in the way defined by that specific type.
 
-for example if we have a struct like the following:
+For example, if we have defined a struct like the following:
 ```wdl
-struct myStruct {
-    Array[File] myFiles
-    Map[String,String] myMap
+struct Experiment {
+    Array[File] experimentFiles
+    Map[String,String] experimentData
+}
+
+```
+**Example 1:**
+Accessing the nth element of experimentFiles and any element in experimentData would look like: 
+```wdl
+workflow workflow_a {
+    Experiment myExperiment
+    File firstFile = myExperiment.experimentFiles[0]
+    String experimentName = myExperiment.experimentData["name"]
+        
+    
 }
 ```
-ssing the nth element of myFiles would look like: `myStruct.myFiles[n]`, while indexing an element in myMap would look like `myStruct.myMap["entry"]`
+
+**Example 2:**
+If the struct itself is a member of an Array or another type, yo
+
+```wdl
+workflow workflow_a {
+    Array[Experiment] myExperiments
+    
+    File firstFileFromFirstExperiment = myExperiments[0].experimentFiles[0]
+    File eperimentNameFromFirstExperiment = bams[0].experimentData["name"]
+    ....
+}
+    
+```
 
 ### Importing Structs
 A `struct` can be defined and imported from another WDL and follows similar rules to other imported objects
@@ -1643,8 +1671,6 @@ task a {
 
 ```
 
-
-Acce
 
 # Namespaces
 
--- a/versions/draft-3/SPEC.md
+++ b/versions/draft-3/SPEC.md
@@ -673,6 +673,7 @@ Engines should at the very least support the following protocols for import URIs
 * `file://`
 * no protocol (which should be interpreted as `file://`
 
+
 ## Task Definition
 
 A task is a declarative construct with a focus on constructing a command from a template.  The command specification is interpreted in an engine specific way, though a typical case is that a command is a UNIX command line which would be run in a Docker image.
@@ -1540,7 +1541,7 @@ struct name { ... }
 ### Struct Declarations
 The only contents of struct are a set of declarations. Declarations can be any primitive or compound type, as well as other structs, and are defined
 the same way as they are in any other section. The one caveat to this is that declarations within a struct do not allow an expression statement after
-the initial member declaration.
+the initial member declaration. Once defined all structs are added to a global namespace accessible from any other construct within the WDL.
 
 for example the following is a valid struct definition
 ```wdl
@@ -1656,20 +1657,29 @@ workflow workflow_a {
 ```
 
 ### Importing Structs
-A `struct` can be defined and imported from another WDL and follows similar rules to other imported objects
+Any `struct` defined within an imported WDL will be added to a global namespace and will not be a part of the imported wdl's namespace. If two structs
+are named the same it will be necessary to resolve the conflicting names. To do this, one or more structs may be imported under an
+alias defined within the import statement.
+
+For example, if your current WDL defines a struct named `Experiment` and the imported WDL also defines another struct named `Experiment` you can
+alias them as follows:
 
 ```wdl
-import structs.wdl as helpers
-
-task a {
-    helpers.MyStruct p
-    
-    command {
-        echo ${ p.name }
-    }
-}
-
+import http://example.com/example.wdl as ex alias Experiment as OtherExperiment
 ```
+
+In order to resolve multiple structs, simply add additional alias statements.
+```wdl
+import http://example.com/another_exampl.wdl as ex2 alia Parent as Parent2 alias Child as Child2 alias GrandChild as GrandChild2
+```
+
+Its important to note, that when importing from file 2, all structs from file 2's global namespace will be imported. This Includes structs from
+another imported WDL within file 2, even if they are aliased. If a strut is aliased in file 2, it will be imported into file 1 under its
+aliased name.
+
+
+
+* Note: Alias can be used even when no conflicts are encountered to uniquely identify any struct
 
 
 # Namespaces
--- a/versions/draft-3/SPEC.md
+++ b/versions/draft-3/SPEC.md
@@ -58,6 +58,13 @@
     * [Parameter Metadata](#parameter-metadata)
     * [Metadata](#metadata)
     * [Outputs](#outputs)
+  * [Struct Definition](#struct-definition)
+    * [Declarations](#struct-declarations)
+      * [Optional and non Empty Struct Values](#optional-and-non-empty-struct-values)
+    * [Using a Struct](#using-a-struct)
+      * [Struct Assignment From Object Literal](#struct-assignment-from-object-literal)
+      * [Struct Member Access](#struct-member-access)
+      * [Importing Structs](#importing-structs)
 * [Namespaces](#namespaces)
 * [Scope](#scope)
 * [Optional Parameters & Type Constraints](#optional-parameters--type-constraints)
@@ -1515,26 +1522,24 @@ workflow wf {
 
 In this example, the fully-qualified names that would be exposed as workflow outputs would be `wf.task1.results`, `wf.altname.value`.
 
-# Structs
+## Struct Definition
 A struct is a C-like construct which enables the user to create new compound types that consist of previously existing types. Structs
 can then be used within a `Task` or `Workflow` definition as a declaration in place of any other normal types. The struct takes the place of the
 `Object` type in many circumstances and enables proper typing of its members.
 
-
-## Defining a Struct
 Structs are defined using the `struct` keyword and have a single section consiting of typed declarationns. They are evaluated first prior
 to workflows and tasks.
-```
+```wdl
 struct name { ... }
 ```
 
-## Defining Declarations
+### Struct Declarations
 The only contents of struct are a set of declarations. Declarations can be any primitive or compound type, as well as other structs, and are defined
-the same way as they are in any other section. The one caveat to this is that declartions within a struct do not allow an expression statement after
+the same way as they are in any other section. The one caveat to this is that declarations within a struct do not allow an expression statement after
 the initial member declaration.
 
 for example the following is a valid struct definition
-``` 
+```wdl
 struct name {
     String myString
     Int myInt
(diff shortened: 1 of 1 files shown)


cjllanwarne:
This is awesome, and exactly what I was hoping for with structs! 

I'd like to raise for discussion (but happy if it gets pushed back for now) the idea of default values in structs, possibly based on other values. As a motivating case, here's a possible solution to the perennial "bam and index" problem:

```wdl
struct BamAndIndex {
  File bam
  File index = basename(bam, "bam") + ".bai"
}

task foo {
  # inputs...
  # command...
  output {
    BamAndIndex out = { bam = "out.bam" }
  }
}
```

patmagee:
@cjllanwarne this was something I was actually debating. I am not 100% certain whether expressions should be evaluated within a struct. I felt as though they should simply be a type definition with no business logic in them. 

Additionally adding "defaults" raises the question of how a struct would be defined within json. At the moment there is an open discussion of whether or not hardcoded values should be overridable by the json value. the current answer is they are not in other constructs. In your example above then, if I wanted to override the location of the index file, i would not be able to.

Another concern that I have is that a Struct is a new Concrete type. If default values worked properly it would mean you could leave out memembers of the struct in favor of their default. This seems odd to me, for example you would never define a json for a `Pair` type like the following:
```json
{
   "myWf.pair_input": {"left":"value"}
}
```

This would constitute an incomplete definition of the type. similarily in a struct definition, leaving out any of its members seems like an incomplete definition to me. 

But I am open for opinions on what other people think

geoffjentry:
@patmagee Is the intention here to replace `Object` and/or "An `Object` which doesn't suck"? If so, I'd recommend removing `Object` from the spec as well here.

patmagee:
@geoffjentry That would be the ultimate goal, but at the same time I think there should be a period of deprecation for the `object`. There are also still people who are probably using it and do ont care about anyhting addressed by the `struct` type

chapmanb:
+1 for this proposal. Patrick's initial post exactly defines what I'd like, and is equivalent in functionality to CWL's records.

As much as I also want an approach for BAM bai indices, I favor the simpler initial proposal and let the index support be a separate discussion.

Thanks much for putting this together.

geoffjentry:
@patmagee re deprecation, my $0.02 is that that's what `draft-2` is for. For instance at the Broad we will need to support `draft-2` for a verrrry long time for exactly this reason. I don't think it's a big deal to swap out what's really an attractive nuisance (IMO) from `draft-3` forward

patmagee:
@geoffjentry I tend to agree with you on that. I still think it might be worth Having others weigh in on the `Object` type to see if there is still a need not addressed by `struct`

cjllanwarne:
I agree re keeping them as static constraints on the type formerly known as Objects (i.e. no defaults!). 

I'm indifferent whether we remove or deprecate `Object` in WDL.

Some questions:
- Are structs global?
  - If I import a subworkflow WDL which outputs a struct, can I use it?
  - Do I need to import it separately? Or declare it again?
  - What if I use two subworkflows declaring the same struct? What if they're subtly different?
- Can struct fields be other structs?
- Can I have a recursive struct or can we explicitly disallow it? eg:
```wdl
struct Tree { 
  Tree? lhs
  Tree? rhs 
}
```
- Could you show an example of an `Object` literal being coerced into an appropriate struct? eg
```wdl
MyStruct s = { "field1": 10, "field2": 20 }
```
- Can a more-useful struct be coerced into a less-useful one? eg
```wdl
MyBigStruct b = { "field1": 10, "field2": 20 }
MySmallStruct s = b # Only needs "field1" from MyBigStruct
```
- Regardless of whether `Object` is still a type, is this a good time to tidy up the definition of the redundant and/or simply no-good `read_object(File)` and `read_json(File)` methods?

patmagee:
I will try to answer your questions in order 

- Structs should be global and defined within the same namespace as the `Task` and the `workflow`
    - A struct import from another wdl, should be referenced in the same way that tasks or wf's are referrecd. ie:
```wdl
#wdl 1
struct MyStruct {
    ....
}

#wdl 2
import wdl_1 as helpful_types

task a {
  helpful_types.MyStruct b
}
```
- I want to say that inner types within a struct can indeed be another struct. Even the Same struct, but there definately would need to be some coerced limit on the level of recursion.
  -   In order to support recursion, I wonder if a struct references itself, if that reference should be implicitly optional so that it does not define an infinite definition 
- object literal for:
```wdl
struct MyStruct {
  File a
  File b
  String name
  Int age
}

###
MyStruct s = {"a": "gs://..."," b": "gs://...", "name": "name", "age": 25}
```
- In my opinion, I do not think 1 struct should be coercible to another struct. At most you could use a previous struct to make a new Object literal for another struct

```
MyStruct s 
MyStruct p = {"a": s.a," b": s.b, "name": s.name, "age": s.age}
```

- As for you last point, I am not sure! Why dont we leav that an open discussion?

cjllanwarne:
What if something like this happens:
- "A defines `MyStruct`"
- "B imports `A.MyStruct`"
- "C imports A.MyStruct"
- "D imports B and C"

Now from D's point of view, it might have some results of type `B.A.MyStruct` and some of type `C.A.MyStruct` and everything starts getting really messy.

Thinking about how other languages deal with this, I think it's perhaps neater to have something like:
```wdl
import MyStruct from useful_types.wdl
```
And remove any namespace qualification on types as they appear further down the WDL. What do you think?

patmagee:
@cjllanwarne I like how that deobfuscates the situation, but it would change the way how imports work. Maybe @geoffjentry can wiegh in on this

patmagee:
@cjllanwarne this is probably not very solvable through a change in syntax, my feeling is that it will be up to the implementor to search the depedency / import tree to define equivalent Struct objects and view them as equivalent regardless of how they are referred.

ie
`B.A.MyStruct` would be equivalent to `C.A.MyStruct`

cjllanwarne:
I think my personal preference would be to remove qualifications and namespacing from struct definitions altogether.

If you import a file we add the set of structs in it to a current set of available structs, and if we import two clashing structs (ie two imported structs have the same name but different fields, then that's an invalid workflow)

orodeh:
👍 +1 this proposal, I have been waiting for this for a while!

Some specifics:
1) I would leave sub-typing out of this for now (coerce `struct A` to `struct B`, where `B` has fewer fields)
2) I would suggest not to allow default values. We can add them later if we want to, but we will never be able to roll it back. 

I am for removing the Object type from draft-3. It is hard to use, and is misleading users.

patmagee:
@cjllanwarne I am not a fan of invalidating a workflow because of a clash in the naming of two Structs. I really hope that people publish libraries of common workflows/tasks/structs etc, and in this case, an end user would not necessarily have any control over the naming schema of two of their imported libraries. 

I also feel like this is not necessarily strictly an issue for the `struct` type, but for all importable objects. At some level, the import resolver should be able to determine if two imports are equivalent or if there is a shared dependency. These dependencies would belong to different namespaces, but the underlying Type / Workflow / Tasks would be the exact same and interpreted as the same. 

I wonder if for First pass its good enough to have structs a member of the imported wdl's namespace like it is currently is, but maybe propose in a separate PR a new way to define Import statements?

versions/draft-3/SPEC.md
```diff
@@ -1515,6 +1522,130 @@ workflow wf {
 
 In this example, the fully-qualified names that would be exposed as workflow outputs would be `wf.task1.results`, `wf.altname.value`.
 
+## Struct Definition
+A struct is a C-like construct which enables the user to create new compound types that consist of previously existing types. Structs
```
yfarjoun: clean up use of `which` and `that` in the same sentance

versions/draft-3/SPEC.md
```diff
@@ -1515,6 +1522,130 @@ workflow wf {
...
+Whereas the following is invalid
+
+```wdl
+struct invalid {
+    String myString = "Cannot do this"
+    Int myInt
+}
+```
+
+You can also use compound types within a struct to easily encapsulate them within a single object. For example
```
yfarjoun: reword to avoid the word "you"

versions/draft-3/SPEC.md
```diff
@@ -1515,6 +1522,130 @@ workflow wf {
...
+    Map[String,Pair[String,File]] myComplexType
+    String cohortName
+}
+```
+#### Optional and non Empty Struct Values
+Struct declarations can be optional or non-empty (if they are an array type). 
+
+```wdl
+struct name {
+    Arrray[File]+ myFiles
```
yfarjoun: Arrrrrr

versions/draft-3/SPEC.md
```diff
@@ -1515,6 +1522,130 @@ workflow wf {
...
+    Boolean? myBoolean
+}
+```
+
+### Using a Struct
+When using a struct in the declaration section of either a `workflow` or a `task` or `output` section you define them in the same way you would define any other type.
+
+For example, if I have a struct like the following:
```
yfarjoun: reword to avoid "I"

versions/draft-3/SPEC.md
```diff
@@ -1515,6 +1522,130 @@ workflow wf {
...
+    }
+}
+
+workflow myWorkflow {
+    MyStruct a
+    call myTask {
+        input:
+            a = a
+    }
+
```
yfarjoun: extra nl

versions/draft-3/SPEC.md
```diff
@@ -1515,6 +1522,130 @@ workflow wf {
...
+```wdl
+
+MyStruct a = {"myName": "John","myAge": 30}
+
+```
+
+
+### Struct Member Access
+In order to access members within a struct, use object notation; ie `myStruct.myName`. If the uderlying member is a complex type which supports member access,
```
yfarjoun: typo in udder-lying. :-)

versions/draft-3/SPEC.md
```diff
@@ -1515,6 +1522,130 @@ workflow wf {
...
+you can access its elements in the way defined by that specific type.
```
yfarjoun: reword to avoid the word "you"

versions/draft-3/SPEC.md
```diff
@@ -1515,6 +1522,130 @@ workflow wf {
...
+
+for example if we have a struct like the following:
```
yfarjoun: reword to avoid "we"

versions/draft-3/SPEC.md
```diff
@@ -1515,6 +1522,130 @@ workflow wf {
...
+```wdl
+struct myStruct {
+    Array[File] myFiles
+    Map[String,String] myMap
+}
+```
+ssing the nth element of myFiles would look like: `myStruct.myFiles[n]`, while indexing an element in myMap would look like `myStruct.myMap["entry"]`
```
yfarjoun: bring the "Acce" from line 1647 back here...

yfarjoun: Perhaps add an example of access when one has an Array[myStruct] and the n-th element in myFiles in the m-th myStruct is being accessed (to point out the brackets before the `.`)

yfarjoun: actually, you should probably have a myStruct **variable** that you are accessing, not the struct itself, right?

versions/draft-3/SPEC.md
```diff
@@ -1515,6 +1522,130 @@ workflow wf {
...
+```
+
+
+### Struct Member Access
+In order to access members within a struct, use object notation; ie `myStruct.myName`. If the uderlying member is a complex type which supports member access,
+you can access its elements in the way defined by that specific type.
+
+for example if we have a struct like the following:
+```wdl
+struct myStruct {
```
yfarjoun: use more different names to the example struct's, and perhaps keep to FirstCap in type names?

versions/draft-3/SPEC.md
```diff
@@ -1515,6 +1522,130 @@ workflow wf {
...
+    call myTask {
+        input:
+            a = a
+    }
+
+}
+```
+
+#### Struct Assignment from Object Literal
+Structs can be assigned using an object literal. When Writing the object, all entries must conform or be coercible into the underlying type they are being assigned to
```
yfarjoun: is that the only way to assign to a struct? Could one assign to it "slowly"? 
For example:

```wdl 
MyStruct a
a.myName="John"
a.myAge=30
```

patmagee: I would say you cannot assign slowly, at least not at the moment

cjllanwarne:
@patmagee I agree with you about failure, that's probably a bit too harsh! 

Ok, so here's what I'm trying to avoid: the difference from previous `import`ed items is that these `Structs` would be the first time where a user of a subworkflow has to know the subworkflow import tree, in full. I think we should stick with a model where subworkflows can change their implementations and not affect importers, so long as the signature is the same. 

My objection is not so much about resolution (although I do think that **would** be very tricky to get right) as it is about readability. Eg. I don't think this is a good thing to aim for (especially if we have to use the long type name multiple times throughout):
```wdl
import foo.wdl

workflow a {
  call foo.fooTask
  output {
    foo.subfoo.gatk.gatkHelpers.gatkTypes.MyStruct s = fooTask.out
  }
}
```
Now imagine someone importing `a` as a subworkflow. Yet another link to add to the type name chain. And if `foo` ever changes implementation, the whole type name is going to have to be renamed (eg I don't want to have to rewrite my workflow if gatk adds a new subworkflow tier)!

I think something like this will be much more flexible, readable, and extensible:
```wdl
import foo.wdl
import MyStruct from gatkTypes.wdl as GatkStruct

workflow a {
  call foo.fooTask
  output {
    GatkStruct s = fooTask.out
  }
}
```
Yes, we still need to know that `fooOut` is a `MyStruct` from `gatkTypes.wdl` but that's always part of the workflow signature. What we don't need to know is the long convoluted chain of imports that got us there.

versions/draft-3/SPEC.md
cjllanwarne: I think @katevoss might disagree with you! FWIW I personally much prefer examples in second-person rather than convoluted passive-voice sentences.

patmagee:
@cjllanwarne I don't think this necessarily solves the issue, it just adds a way to easily overcome it. We should still probably allow for the "Chained" method for completeness. if users really want to do that and not properly import a struct, then it's probably their prerogative to do so

versions/draft-3/SPEC.md
cjllanwarne: Only valid on the 19th September

versions/draft-3/SPEC.md
cjllanwarne: Again, I prefer this to passive-voicing

versions/draft-3/SPEC.md
cjllanwarne: I don't like this. WDL values are immutable and I think they should stay that way.

versions/draft-3/SPEC.md
cjllanwarne: I think "Types names must be `TitleCase`" is a good idea

versions/draft-3/SPEC.md
katevoss: @cjllanwarne is right! We encourage documentarians to address their readers directly with the second-person in a friendly yet clear manner. 

As our coffee mugs say,
>It's more important to be nearly right and understandable than to be academically accurate and unintelligible.

versions/draft-3/SPEC.md
yfarjoun: OK. in that case, please also include an example of building a complex struct using other (previously defined) values:

```wdl
Array[File] someFiles=["file1.txt","file2.txt"]
Int someAge=30

struct MyStruct{
Array[File] files
Int age
}

MyStruct a={"files"=someFiles "age"=someAge}
```

Also given that the key/value pair has quotes on the key, it is possible to parametrize the assignments:

```wdl
Array[File] someFiles=["file1.txt","file2.txt"]
Int someAge=30
String key1="files"
String key2="age"
struct MyStruct{
Array[File] files
Int age
}

MyStruct a={key1=someFiles key2=someAge}
```
? if not, I'd advocate for removing the quotes from the keys.

cjllanwarne: I like this idea of removing quotes from the keys, not least because it makes sure keys are valid identifiers rather than arbitrary strings.

versions/draft-3/SPEC.md
patmagee: So votes on language? Should we keep the voice referring to the user or move to a passive one

versions/draft-3/SPEC.md
patmagee: @yfarjoun will do for examples.
@cjllanwarne I talked to @geoffjentry I think, UpperCase first letter should be convention > enforced. WDYT?

versions/draft-3/SPEC.md
```diff
@@ -1515,6 +1522,130 @@ workflow wf {
 
 In this example, the fully-qualified names that would be exposed as workflow outputs would be `wf.task1.results`, `wf.altname.value`.
 
+## Struct Definition
+A struct is a C-like construct which enables the user to create new compound types that consist of previously existing types. Structs
+can then be used within a `Task` or `Workflow` definition as a declaration in place of any other normal types. The struct takes the place of the
+`Object` type in many circumstances and enables proper typing of its members.
+
+Structs are defined using the `struct` keyword and have a single section consiting of typed declarationns. They are evaluated first prior
+to workflows and tasks.
```
cjllanwarne: I think it's implied here, but I think we should be explicit (especially since @yfarjoun already wrote one inline in one of the examples...):

> "Structs are declared separately, outside of any `workflow` or `task` definition". They are part of the namespace of the WDL file they are written in.

versions/draft-3/SPEC.md
```diff
@@ -1515,6 +1522,130 @@ workflow wf {
 
 In this example, the fully-qualified names that would be exposed as workflow outputs would be `wf.task1.results`, `wf.altname.value`.
 
+## Struct Definition
+A struct is a C-like construct which enables the user to create new compound types that consist of previously existing types. Structs
+can then be used within a `Task` or `Workflow` definition as a declaration in place of any other normal types. The struct takes the place of the
+`Object` type in many circumstances and enables proper typing of its members.
+
+Structs are defined using the `struct` keyword and have a single section consiting of typed declarationns. They are evaluated first prior
```
yfarjoun: consiting->consisting

versions/draft-3/SPEC.md
patmagee: @geoffjentry I would be interested to know what is currently allowable syntax with Object Literal Notation. Does it need to be JSON like? Is it possible to remove the quotes from keys under the current spec

versions/draft-3/SPEC.md
patmagee: Good catch. I think making this explicit is quite important

versions/draft-3/SPEC.md
cjllanwarne: I tend to be in the "if it's a convention that everyone is expected to follow, it might as well be enforced" camp (eg Java class names might as well be enforced `UpperCase` at this point IMO).

I don't see a big cost to enforcing UpperCase. Every other type name is `UpperCase` so people are already keyed to the idea that that's what types look like.

versions/draft-3/SPEC.md
geoffjentry: I prefer the current form

geoffjentry:
@patmagee A shower thought I had this morning - I'm reversing course on my opinion on removal of object in this PR as that would be the first backwards incompatible change in `draft-3`. Leaving object in allows us to punt the "how the hell do we discern version differences" question a bit further down the road, and once that's settled we can remove object then. WDYT?

cjllanwarne:
@patmagee if enough naive WDL authors take up their prerogative to write long chains of imports, that creates a huge disincentive to the authors of the base WDL to refactor. Eg 
> "I'd love to move this section into a subworkflow, but then anyone who wants to upgrade to my new version would have to rewrite their entire WDL rather than just import the new version, and anyone who wants to use their new version would have to do the same, and so on, and so on".

FWIW this is the exact reason we made subworkflows declare their outputs explicitly - to avoid these long chains of `workflowA.workflowB.workflowC.out` output names

versions/draft-3/SPEC.md
geoffjentry: FWIW I also prefer immutability.

@patmagee That's a good question. I realized yesterday that object isn't actually *defined* anywhere, period. It also doesn't help that AFAIK Broad folks aren't using object so it's not something we've seen a lot. And of course there's the "goodness only knows what Cromwell chooses to do w/ objects" aspect :)

It looks like the `isObjectLiteral` function in `wdl4s` is used absolutely nowhere  which makes me wonder if there's any support for it at all.

versions/draft-3/SPEC.md
geoffjentry: Going back to the you/passive thing, I do find bouncing between "you" and "we" to be odd.

versions/draft-3/SPEC.md
geoffjentry: My $0.02 is that convention is fine unless it's problematic for defining in parsing/etc. 

I get the allusions to other languages but at the end of the day WDL is being marketed as a language for people who don't want to be using programming languages. If it just starts taking every trope from popular prog langs, why would users bother using it when they could just go over to e.g. python?

patmagee:
@geoffjentry I am okay with leaving the Object in for now. As I mentioned before, I feel like it is out of the scope of this PR and probably should be separately looked at and debated. 

@cjllanwarne I see what you are saying, and I definitely agree with you, I just don't know how to properly implement it?  When I import another subworkflow or any WDL for that matter, would we need to explicitly declare All `structs` we want to use at the top? That could be quite burdensome for the user, especially if they are importing a WDL library of many different struct types.

The other way to think about this is to go back to your original suggestion and import all structs to the global namespace, but try to figure out a way to better resolve conflicts (Which the farther down this rabbit hole we get might be the best way to go). Maybe the best way is actually to fail the workflow as you suggested and force the user to alias the failing Struct. If this was the case then we should have to only Alias the conflicting types, and not all types contained in the WDL.

```wdl
import "https://exmaple.com/exampl.wdl" as ns alias Bam as Bam2 alias Fastq as Fastq2
```

cjllanwarne:
@patmagee that long list of imports *was* what I was proposing (just like [scala](https://github.com/broadinstitute/cromwell/blob/29_hotfix/engine/src/main/scala/cromwell/engine/workflow/lifecycle/execution/WorkflowExecutionActor.scala#L3)) but I see your point... 😄 

I like your idea of importing everything flat and aliasing only if required because it removes the burden of thinking about this completely until a collision occurs (and if base-WDL authors are conscientious and well-reviewed that should hopefully be rare).

patmagee:
@cjllanwarne @geoffjentry if a few more people weigh in on the importing suggestion (maybe aliasing) I will update the PR

orodeh:
I like the idea of flat import semantics, due to its conceptual simplicity. Aliasing is a good solution when collisions occur. 
👍

patmagee:
Added an alias statement within the import section for resolving struct conflict. This should fix the issues and messiness over importing the struct

versions/draft-3/SPEC.md
```diff
@@ -1656,21 +1657,30 @@ workflow workflow_a {
...
-        echo ${ p.name }
-    }
-}
+```wdl
+import http://example.com/example.wdl as ex alias Experiment as OtherExperiment
+```
 
+In order to resolve multiple structs, simply add additional alias statements.
+```wdl
+import http://example.com/another_exampl.wdl as ex2 alia Parent as Parent2 alias Child as Child2 alias GrandChild as GrandChild2
```
cjllanwarne: I'd split this over three lines for clarity. Also noticed a typo on `alia` as I did this, eg
```
import http://example.com/another_exampl.wdl as ex2 
  alia Parent as Parent2 
  alias Child as Child2 
  alias GrandChild as GrandChild2
```

versions/draft-3/SPEC.md
```diff
@@ -1515,6 +1523,165 @@ workflow wf {
...
+import http://example.com/example.wdl as ex alias Experiment as OtherExperiment
+```
+
+In order to resolve multiple structs, simply add additional alias statements.
+```wdl
+import http://example.com/another_exampl.wdl as ex2 alia Parent as Parent2 alias Child as Child2 alias GrandChild as GrandChild2
+```
+
+Its important to note, that when importing from file 2, all structs from file 2's global namespace will be imported. This Includes structs from
+another imported WDL within file 2, even if they are aliased. If a strut is aliased in file 2, it will be imported into file 1 under its
```
orodeh: typo: strut ---> struct

versions/draft-3/SPEC.md
```diff
@@ -1515,6 +1523,167 @@ workflow wf {
 
 In this example, the fully-qualified names that would be exposed as workflow outputs would be `wf.task1.results`, `wf.altname.value`.
 
+## Struct Definition
+A struct is a C-like construct which enables the user to create new compound types that consisting of previously existing types. Structs
+can then be used within a `Task` or `Workflow` definition as a declaration in place of any other normal types. The struct takes the place of the
+`Object` type in many circumstances and enables proper typing of its members.
+
+Structs are declared separately from any other constructs, and cannot be declared within any `workflow` or `task` definition. They belong to the namespace of the WDL 
+file which they are written in and are therefore accessible globally within that WDL. Additionally, all structs must be evaluated prior to their use within a `task`,
```
geoffjentry: Can you explain further what you mean by "all structs must be evaluated ..."?

versions/draft-3/SPEC.md
```diff
@@ -1515,6 +1523,167 @@ workflow wf {
...
+Structs may be defined using the `struct` keyword and have a single section consisting of typed declarations. 
+
+```wdl
+struct name { ... }
+```
+
+### Struct Declarations
+The only contents of struct are a set of declarations. Declarations can be any primitive or compound type, as well as other structs, and are defined
+the same way as they are in any other section. The one caveat to this is that declarations within a struct do not allow an expression statement after
+the initial member declaration. Once defined all structs are added to a global namespace accessible from any other construct within the WDL.
```
geoffjentry: Should make it explicit what happens in case of collision?

versions/draft-3/SPEC.md
geoffjentry: Agree w/ both of you. The "I" sounds super awkward in a spec document.

Mabye "if there is a struct like the following:"?

versions/draft-3/SPEC.md
geoffjentry: I see it descrbied below. Perhaps make a quick reference to that section?

cjllanwarne:
One other comment from me: I think it'd be useful to list `Struct`s somewhere under the `Types` section ([link](https://github.com/openwdl/wdl/blob/master/versions/draft-3/SPEC.md#types)). 

I wouldn't want to include all the details there, but maybe we could mention that they exist and link down to the main `Structs` text further down?

geoffjentry:
@patmagee As you're a member of the OpenWDL core group I'm assuming you're self-shepherding here. When you feel it is ready for voting as-is, please label it as `voting active` and announce that to the mailing list/gitter channel

chapmanb:
Double +1 good. I can't wait to have these for CWL record interoperability.

versions/draft-3/SPEC.md
```diff
@@ -1515,6 +1528,167 @@ workflow wf {
...
+
+In order to resolve multiple structs, simply add additional alias statements. 
+```wdl
+import http://example.com/another_exampl.wdl as ex2 
+    alias Parent as Parent2 
+    alias Child as Child2 
+    alias GrandChild as GrandChild2
+```
+
+Its important to note, that when importing from file 2, all structs from file 2's global namespace will be imported. This Includes structs from
```
cjllanwarne: Erroneous capital `I` on `Includes`

cjllanwarne:
I'm really excited to implement this, I think it's gonna be a huge QoL improvement for WDL authors. 👍

orodeh:
👍

ambrosejcarr:
👍

versions/draft-3/SPEC.md
vdauwera: I can’t speak to what one should normally write in a spec document but just keep in mind that passive voice and other lawyerly constructs are a leading source of confusion and difficulty, especially for non-native English speakers.

vdauwera: I can’t speak to what one should normally write in a spec document but just keep in mind that passive voice and other lawyerly constructs are a leading source of confusion and difficulty, especially for non-native English speakers.

vdauwera:
:+1:

mohawkTrail:
I submit you'd get a richer, coded and tested by other people, implementation for this feature if you adopt the protobuf definition for this type.

The complex types are naturally represented in a .proto IDL definition, and so can be known types to the WDL script.  Thus the complex structures can be be straightforwardly prepared by the workflow caller and accepted and used in the task, no matter its implementation language, and neatly interpreted and used as output.  Even the code of taking the binary format and creating a scala object is already written.

dinvlad:
I actually like the Struct proposal a lot, as I think it fits the existing WDL typing system very organically while enabling highly flexible hierarchical data types.

One thing that's still a bit unclear to me and more to a point raised by @mohawkTrail above, how would we feed struct inputs to a workflow (as opposed to passing them from one task to another)? Currently, we can pass any "standard" WDL type as a JSON object according to [SPEC.md#specifying-workflow-inputs-in-json](https://github.com/openwdl/wdl/blob/develop/SPEC.md#specifying-workflow-inputs-in-json), and it will be coerced into WDL types per [SPEC.md#type-coercion](https://github.com/openwdl/wdl/blob/develop/SPEC.md#type-coercion). However, will WDL coerce non-trivial JSON objects into Structs?

Enable arbitrarily complex JSON in the inputs would then open doors for interoperability with other common(-ish) data types, like YAML and Proto.

geoffjentry:
@mohawkTrail @dinvlad Note that now that voting is active, people are voting on the proposal **as-is**

geoffjentry:
@patmagee A half hearted :+1: - I'd have still liked to see some of the edits requested in the PR, but since the options are to vote yay or nay as is I dont view them as blockers

patmagee:
@geoffjentry I think this is likely to be a first pass at structs of you will. There are a lot of concern s which have been raised by @dinvlad and @mohawkTrail however i think some of those concerns (ie how o puts are defined) are questions for the engine implementation and not the language specification.

mohawkTrail:
I agree with @patmagee, and with @geoffjentry.  The WDL definition of struct is and should be consonant with WDL syntax and semantics (immutability for example).  My suggestion was on implementation, and should be invisible in WDL--the difference being in the serialization and deserialization, and ease-of-use in the caller's, tasks', and output consumers' languages.

If we were to pursue this at another time, we might permit alternate input and output formats for the WDL struct, and include a tool for taking a .proto definition, verifying it corresponds to a valid WDL struct, and producing the struct's WDL declaration for import.

rhpvorderman:
:+1:

PR #161, opened 2017-11-21
Specify how overriding inputs works

cjllanwarne:
As I started wiring up input-overrides in WOM, I've hit the fact that the SPEC as written says that declarations with expressions (eg `Int x = 5`) should not be evaluated as workflow inputs (implying that they wouldn't be valid entries in a user's inputs json).

I've tried to define this ability to override, and to be specific about where it would and would not be valid. 

NB: I don't think this is necessarily the final answer (I guess that people will probably want more options to overrule call input blocks, for example), but I believe this scheme as presented is internally consistent and hopefully starts up the conversation on what we want to allow to be overridden, and how easy to make that.

### DIFF
--- a/versions/draft-3/SPEC.md
+++ b/versions/draft-3/SPEC.md
@@ -70,6 +70,10 @@
   * [Task Inputs](#task-inputs)
   * [Workflow Inputs](#workflow-inputs)
   * [Specifying Workflow Inputs in JSON](#specifying-workflow-inputs-in-json)
+  * [Optional Inputs](#optional-inputs)
+  * [Declared Inputs: Defaults and Overrides](#declared-inputs-defaults-and-overrides)
+    * [Optional Inputs with Defaults](#optional-inputs-with-defaults)
+  * [Call Input Blocks](#call-input-blocks)
 * [Type Coercion](#type-coercion)
 * [Standard Library](#standard-library)
   * [File stdout()](#file-stdout)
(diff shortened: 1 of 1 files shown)


cjllanwarne:
NB: here's the change as rendered in situ in the doc: https://github.com/cjllanwarne/wdl/blob/cjl_workflow_input_overrides/versions/draft-3/SPEC.md#computing-inputs

patmagee:
I definitely agree that overridable inputs are a must-have feature, but I am not sure if we should allow ALL inputs to be overridable. There are many cases where I am making a workflow that I do not necessarily want to expose a certain set of inputs to the user, but instead, want to hardcode them in the workflow itself.

In my opinion, workflow inputs should essentially be `final` once they have been declared, unless they are an optional input.  So variables that are declared like the following:
```wdl
workflow a {
  String s = "silly String"
}
```
Would not be overridable but the following would be.

```wdl
workflow b {
  String? s = "silly string"
}
```

cjllanwarne:
@patmagee I'd rather not override `?` since that's already an indicator for optional types (and then we'd still need logic to tell whether it's an optional input vs an intermediate declaration from a task in an `if` block).

That said... I'm not against the idea of making the difference between "input" and "intermediate value" explicit rather than implied (eg I'm coming around to the idea of the `input {}` block that @orodeh mentioned).

How would people feel about something like this compromise:
- We add an explicit `input { }` block syntax that you can put inputs definitions into.
- If the `input { }` block exists, then everything inside can be overridden, and nothing outside can (to consider: in this case missing task inputs are... [error cases / still allowed]?)
  - The onus is now on the workflow author to be explicit, but the trade-off is that it's easy for users to see the inputs, and authors can be very careful about what they allow.
- If the `input { }` block is absent the inputs are inferred from the workflow using the rules in this PR (or something similar)
  - The workflow author loses some control about what is overridable to the workflow runner, eg in the case of a method developer who just wants to iterate fast without so much lockdown.

geoffjentry:
@patmagee @cjllanwarne was there some sort of consensus here or did it just kinda get dropped? Was going to move it to voting soon, but wanted to stoke the fire if the latter

orodeh:
Personally, I think this is a good compromise. 

In dxWDL I am facing the same problem. If you assume constants can be overridden, you end up with lots of workflow inputs, some of which the user does not intend to be inputs at all.

👍

cjllanwarne:
If this sounds sensible, I think this PR is valid to represent the default state of how inputs are inferred, so I suggest we vote on this PR as is.

I want to add the "and here's an `input { }` block syntax if WDL authors want to lock down the set of inputs" PR for separate consideration/voting

orodeh:
In my personal opinion, a user should be able to override a assigned, non default, declarations like:
```
Float threshLo = 0.5
Float threshHi = threshLo + 0.3
```
only if they are in an `input{ }` block. Otherwise, large workflows will have large numbers of inputs. This makes a difference for web/UI interfaces, that have to represent them on the screen. This is a situation we have at DNAnexus, trying to hook up our UI to WDL.

cjllanwarne:
@orodeh the feedback I usually get from our methods folks is, "lots and lots of possible inputs is exactly what we want during our rapid method iteration process", that's why I want to allow for that without having to type in lots and lots of boilerplate pass through variables.

I think the best compromise is to have a situation that's sort of like outputs: if you add an `output` block you get what you ask for. If you don't then you get sort-of a mess and you can't use your WDL as a sub-workflow. If we do the same for `input`, that seems to me like a good enough incentive for publishing WDL authors to play ball without forcing it on anyone who just wants to get started quickly.

Does that sound ok to you? Would you like me to make all that explicit in this PR before we start voting, or are you happy to vote on this PR as the "no `input` block" case and add the second half in another PR?

orodeh:
@cjllanwarne, I am it fine with requiring the `input` block on sub-workflows, and allowing regular workflows to have lots of inputs. I would much rather this was put into this PR, however.

cjllanwarne:
@orodeh @patmagee did you get a chance to look at the updates to this PR? It might be a little stricter than you originally proposed (but I think that could be a good thing, so long as the upgrade path from draft2 to draft3 is automate-able)

orodeh:
I guess I didn't notice this, but you made input definitions strict. So now, a task and a workflow must declare their inputs. I am fine with this change, I can see how it can be automated. 
👍

patmagee:
@cjllanwarne sorry I didnt see the updates either. But this looks better, I also like the strictness and enforcing this at both the workfow and task level. I think the input block is more in line with WDL, since there is an output block explicitly defined as well

LeeTL1220:
My 2 cents:  I am not excited about converting my WDL2 to WDL3.  If there is a script that automates that, I'd be very grateful (even if it does not 100% correctly guess what should be in an input block).  Is that even planned?

Otherwise, I like the idea that we can protect intermediate variables from being overridden.

LeeTL1220:
👍

cjllanwarne:
@LeeTL1220 (apologies to openWDL for a very Cromwell-centric comment):

* My intention is definitely to add an upgrade tool to `womtool` that can read in a draft2 file and print it out in draft 3. 
* I'll also note that Cromwell will continue support for draft 2 even after we add draft 3 so unlike the previous times that WDL has changed, you'll be able to take a new Cromwell without worrying about also upgrading all your WDLs at the same time.

LeeTL1220:
@cjllanwarne 🤗

geoffjentry:
:+1:

I'm not sure if this is 2-0 or 3-0 but either way it passes. Waiting for implementation. 

@cjllanwarne Let me know if you see this implemented somewhere

ruchim:
:+1:

PR #162, opened 2017-11-21
Proposal for 'after' keyword

cjllanwarne:
Tidies up the call section, and introduces a proposed new `after` keyword to provide non-output-based dependencies

### DIFF
--- a/versions/draft-3/SPEC.md
+++ b/versions/draft-3/SPEC.md
@@ -1557,37 +1557,44 @@ The rational for this is that a user may want to provide the following input fil
 
 ### Call Statement
 
-```
-$call = 'call' $ws* $namespaced_identifier $ws+ ('as' $identifier)? $ws* $call_body?
-$call_body = '{' $ws* $inputs? $ws* '}'
-$inputs = 'input' $ws* ':' $ws* $variable_mappings
-$variable_mappings = $variable_mapping_kv (',' $variable_mapping_kv)*
-$variable_mapping_kv = $identifier $ws* '=' $ws* $expression
-```
+A workflow calls other tasks/workflows via the `call` keyword.  A `call` is followed by the name of the task or subworkflow to run.  Most commonly that's simply the name of a task (see examples below), but it can also use `.` as a namespace resolver.
 
-A workflow may call other tasks/workflows via the `call` keyword.  The `$namespaced_identifier` is the reference to which task to run.  Most commonly, it's simply the name of a task (see examples below), but it can also use `.` as a namespace resolver.
-
-See the section on [Fully Qualified Names & Namespaced Identifiers](#fully-qualified-names--namespaced-identifiers) for details about how the `$namespaced_identifier` ought to be interpreted
+See the section on [Fully Qualified Names & Namespaced Identifiers](#fully-qualified-names--namespaced-identifiers) for details about how the identifier ought to be interpreted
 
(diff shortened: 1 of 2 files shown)


geoffjentry:
I thought I was going to hate `after` but I quite like it. 

Is there any harm in allowing `after` in cases where the outputs already provide a dependency?

vdauwera:
Ooh I like this idea. I can see this being useful if I want to have a task write to a dB then another read from the db, if there’s no actual file outputs to use as dependencies.

geoffjentry:
Yeah this is a fairly common complaint. Our stock answer is to wire in a fake output/input but that is horrible. Left to my own devices I’d probably have gone with the ‘dependsOn’ syntax from zamboni but ‘after’ reads so naturally with the call syntax

cjllanwarne:
@geoffjentry I don't see any harm in allowing both. The `after` will be redundant if there's already an output-based dependency but it won't *break* anything. Maybe it even makes things more readable in some cases 🤷‍♂️

orodeh:
Looks good to me.

Horneth:
Totally ToL and personal opinion: I find something like `andThen` easier to reason about since the order of execution matches the left to right order of reading / writing.

versions/draft-3/SPEC.md
```diff
@@ -1187,37 +1187,44 @@ workflow wf {
...
 
 All `call` statements must be uniquely identifiable.  By default, the call's unique identifier is the task name (e.g. `call foo` would be referenced by name `foo`).  However, if one were to `call foo` twice in a workflow, each subsequent `call` statement will need to alias itself to a unique name using the `as` clause: `call foo as bar`.
 
-A `call` statement may reference a workflow too (e.g. `call other_workflow`).  In this case, the `$inputs` section specifies a subset of the workflow's inputs and must specify fully qualified names.
+A `call` statement may reference a workflow too (e.g. `call other_workflow`).  In this case, the inputs section specifies the workflow's inputs by name.
+
+Calls can be run as soon as their inputs are available. If `call x`'s inputs are based on `call y`'s outputs, this means that `call x` can be run as soon as `call y` has completed. 
+
+To add a dependency from x to y that isn't based on outputs, you can use the `after` keyword, such as `call x after y after z`. But note that this is only required if `x` doesn't already depend on an output from `y`.
```
yfarjoun: I think that if there are multiple dependencies the notation should be less redundant. Perhaps

```wdl
call x after y and z
```
or
```wdl
call x after y,z
```

yfarjoun:
@Horneth your suggestion will make it harder to understand the flow of calls since to determine the dependence of call X one would need to examine *all* the call statements (to see if any of them have `andThen X`), while with the current proposal, you only need to examine the `call X` statement.

geoffjentry:
Agree with @yfarjoun. I think the `after` syntax fits better into the WDL mantra of simplicity & clarity

orodeh:
👍

vdauwera:
👍

geoffjentry:
:+1:

geoffjentry:
This has officially passed on a vote of 3-0 (really 4-0, but I forgot to vote).

As there are no implementations, this will hang out until such time as it actually exists

abaumann:
one plug for after with dependencies is this: https://github.com/openwdl/wdl/issues/183

although i dont think this definition _requires_ after to always run, only guaranteed ordering despite dependencies

geoffjentry:
@abaumann At this point this PR is as-is, which is not to say that one couldn't open a subsequent PR

geoffjentry:
@cjllanwarne It'd be good to update this to reflect it being against `development` and not `draft-3`

geoffjentry:
Merging as there is an implementation (Cromwell) and there are no conflicts

PR #163, opened 2017-11-28
Workflow and subworkflow outputs: clarification and tidy-up

cjllanwarne:
- Removes the deprecated wildcard format
- Mentions that subworkflows must declare their outputs, just like tasks
- General tidy-up and up-to-date-ification of the outputs section

### DIFF
--- a/versions/draft-3/SPEC.md
+++ b/versions/draft-3/SPEC.md
@@ -1417,14 +1417,14 @@ As an example:
 
 ### Outputs
 
-Each `workflow` definition can specify an optional `output` section.  This section lists outputs from individual `call`s that you also want to expose as outputs to the `workflow` itself.
-If the `output {...}` section is omitted, then the workflow includes all outputs from all calls in its final output.
-Workflow outputs follow the same syntax rules as task outputs.
-They can reference call outputs, workflow inputs and previous workflow outputs.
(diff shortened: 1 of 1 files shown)


orodeh:
Looks good to me. My only suggestion is to remove the provision for leaving the output section empty. If it is empty, then just output nothing. 

Workflows can generate large volumes of data. If the outputs are declared, it is easy, for example, to remove intermediate results.

cjllanwarne:
@orodeh I'm in two minds about that. As an engine developer I agree with you but the value of allowing omitting the `output` block is that we allow people to get started with WDL while minimizing the amount of boiler plate.

I agree that for any sort of large-scale WDL you definitely want to be specific (and indeed, you must have it to use your workflow as a subworkflow), but I don't think we can ignore the "getting started quickly" use case.

I think it would be valid for an engine to say "if you don't have an output block, you can't use this in a subworkflow, and this engine won't clean up intermediate results for you"

geoffjentry:
@orodeh @cjllanwarne was this consensus or did it get dropped? Will move it to voting tomorrow if there's no more commentary so just wanted to double check

orodeh:
👍

vdauwera:
👍

geoffjentry:
:+1:

geoffjentry:
This passes 3-0. 

As this is a cleanup to synch with existing behavior in Cromwell, it is immediately mergeable

PR #168, opened 2017-11-30
Remove 'while' loops from the SPEC

cjllanwarne:
I believe these `while` loops were a "let's fill in the details later" placeholder in the SPEC, which we never got around to filling in.

I'm happy for a similar feature to be proposed as required, but for now I think we should remove this not-fully-thought-through section.


vdauwera:
👍

orodeh:
Seems fine to me. We can always bring them back if we really want them. 
👍

geoffjentry:
:+1:

geoffjentry:
This passes 3-0.

As the point here was to remove something which has never been implemented, there's no need to wait for implementation. Merging.

PR #172, opened 2017-12-20
cromwell_on_google: Fix monitor_wdl_pipeline output on MacOS

mbookman:
Interpreting the workspace contents has several uses of sed, which
rely on extended regular expressions, namely the use of "+" for
one-or-more instances. On MacOS, we explicitly need the "-E" flag
to enable extended regular expressions.

### DIFF
--- a/runners/cromwell_on_google/monitoring_tools/monitor_wdl_pipeline.sh
+++ b/runners/cromwell_on_google/monitoring_tools/monitor_wdl_pipeline.sh
@@ -166,15 +166,17 @@ while [[ $(get_operation_done_status "${OPERATION_ID}") == "false" ]]; do
     # have an "exec.sh" with no "*-rc.txt" file.
     WS_CALLS_STARTED=$(
       echo "${WS_EXECS}" \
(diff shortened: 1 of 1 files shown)


geoffjentry:
Meta comment for @vdauwera - this reminds me that this stuff is still here :)

geoffjentry:
@mbookman Since this is from the older version of this repo I'm not sure what the process is here. I'll leave it up to you and @vdauwera to hash out

patmagee:
I would suggest moving this into a separate repo linked to cromwell and removing it from the wdl project. In fact all components not strictly WDL should probably be moved into their own repos (including parsers) and what not

geoffjentry:
@patmagee I've been giving @vdauwera some breathing room on this, she's been ever so slightly occupied with some sort of software launch, something like that. Or so i've been told :P

patmagee:
Yup go right ahead!

geoffjentry:
👍

patmagee:
@vdauwera is this ready to be merged?

PR #173, opened 2017-12-21
WDL Directory type

cjllanwarne:
Temporary Warning: This PR currently includes changes being made in #175 but not yet merged. Please only review from commit `"Directory type"` onwards.
----

I was a little bit scared to propose this since `Directory` is non-trivial to implement well at scale in a cloud environment with object stores. However, it's high up on every WDL author's wish list (and selfishly, WOM means that Cromwell will be supporting directories already, we would just need a syntax to interact with them in WDL).

I suggest we start small with just the data type and an ability to localize and delocalize directories in tasks. Any other interactions (eg engine functions for directory querying) can follow as or when required.

In addition to proposing adding a Directory type, I wanted to mention how they are localized and delocalized. I think this is a valid concern for the workflow definition since it might change execution outcomes if different engines interpret localization differently. Since I couldn't find any mention of how `File`s are localized for execution I added it all together in a new section.


versions/draft-3/SPEC.md
```diff
@@ -684,6 +690,34 @@ $task = 'task' $ws+ $identifier $ws* '{' $ws* $declaration* $task_sections $ws*
...
+}
+```
+
+#### Input Localization
+`File` and `Directory` inputs must be treated specially since they require localization to within the execution directory:
+- Files and directories are localized into the execution directory prior to the task execution commencing. 
+- When localizing a `File`, the engine may choose to place the file wherever it likes so long as it accords to these rules:
+  - The original file name must be preserved even if the path to it has changed.
+  - Two input files with the same name must be located separately, to avoid name collision.
+  - Two input files which originated in the same storage directory must also be localized into the same directory for task execution.
```
orodeh: This is a problem with DNAnexus cloud storage system, which allows two files to have the same name, and be placed in the same directory. This is allowed to support versioning.

orodeh:
I think this is a good approach. I can see how an implementation could reasonably use `tar` to implement directories.

I applaud the written down rules about file placement during localization. The one comment I have,  which I also posted inline, has to do with files that come from the same cloud directory. On systems that allow two files to share the same name and path (but not `id`), this is going to be a problem. DNAnexus is one such system, it allows this seemingly illogical requirement for versioning purposes.

versions/draft-3/SPEC.md
cjllanwarne: @orodeh Oh, huh, TIL! Out of interest, how do you handle that sort of localization collision today? It'll presumably make sense to codify this in such a way that you don't need to change anything

orodeh: Today, I keep the filename, but create a new local directory for them. For example,
assume file `A.txt` has two versions with ids `file-xxxx`, and `file-yyyy`.  If a task uses both files, then the first one will be localized as $HOME/A.txt, and the second to $HOME/file-yyyy/A.txt.

antonkulaga:
Having problems with Directories (from absence caching to .fai/bai issues) is a huge pain for me, so I totally support this PR!

versions/draft-3/SPEC.md
cjllanwarne: This sounds like a good enough scheme to me. And it seems pretty unlikely to me that this would happen by accident without the author realizing so I don't think it would reduce portability. Out of interest how are you deciding which File is "first" in this case? 

How about this as a draft update:

```
#### Special Case: Versioning Filesystems
If two or more files in a versioning filesystem might have the same name in the same directory, the following special procedure can be used:
  - The first file is always placed as normal according to the usual rules.
  - Subsequent files that would otherwise overwrite this file are instead placed in a subdirectory named for the version.

For example imagine two versions of file fs://path/to/A.txt are being localized (labelled version 1.0 and 1.1). The first might be localized as /execution_dir/path/to/A. The second must then be placed in /execution_dir/path/to/1.1/A
```

orodeh: Sounds like a good solution to me. 

The dxWDL runtime serves the localized files on a first come first serve basis. The first file to be localized occupies `fs://path/to/A.txt`. All subsequent files with this name are placed in `fs://path/to/TAG/A.txt`.

geoffjentry: "can be used"

Is the use of the word "can" intentional here? It's a loaded term in standards docs

geoffjentry:
@antonkulaga just keep in mind that this has nothing to do with caching, which is a cromwell feature.

antonkulaga:
I know, but I hope it will speed-up its development

geoffjentry:
@cjllanwarne would you be amenable to separating the directory & file stuff? this way file can go in quickly, as I suspect directory will not

cjllanwarne:
@geoffjentry done. Now follows on from #175

patmagee:
@cjllanwarne  I like this approach too and I think starting simple is the best option. 

Just a few comments:

1. can you please clarify what exactly constitutes a directory, although it seems easily implied that a `Directory` has a 1:1 correlation with a filesystem directory, I am not sure if that could always be the case. 

2. related to the above comment,  do we care what the implementors consider a `Directory`  or is it enough to state that, so long as it complies with the set behavior, then it IS, in fact, a directory

3. How will permissions work with loaded directories? Is this something defined by the engine, or should it be defined in the specification?  the easiest solution would be to try and keep the source permissions but I am not sure if that is possible under all circumstances

4. Are there any limitations to the Directory depth?

5. Are there any limitations to the file size

6. Should engine functions such as `size` be able to take a directory as well (this could conceivably be a VERY expensive operation) 

7. Also could you please add a few usage examples (Which would be amazing!!!)

jonn-smith:
@cjllanwarne This looks like a good way to address adding in a directory type.  It will certainly solve the simple cases where I had trouble.

geoffjentry:
@patmagee re your question #6 that's a really good point. I think that at least to start with that for this reason we should **not** support such functions. If someone really wanted to they could handroll it (which would imply that they know what they're doing). 

One of the guiding philosophies we've had as a team was to not worry about going all in with all hte bells and whistles but rather to move in the right direction and see what shakes out. I think this is a good place for that behavior.

patmagee:
@cjllanwarne how is this coming?

cjllanwarne:
@patmagee since you asked, I thought I'd run this past you. I had a realization last night that draft3 is getting kind of full (I believe it was originally intended as just the set of corrections and necessary breaking changes to reach a stable 1.0).

My gut feeling is that this PR might be a good candidate for tabling for now and pushing as a 1.1 feature. My hope is that we can get to 1.0 quickly and that once 1.0 is out, non-breaking-changes in point releases can happen at more regular intervals to avoid the current "I must get it in **now** or it'll never happen" feeling in PRs (of which I've been very guilty!).

(I know this is a strange place to bring up this thought, but IMO this is a classic example of the type of PR I think would be good to revisit for a not-too-distant 1.1)

patmagee:
@cjllanwarne now that Draft 3 has pretty much been wrapped up, I wanted to come back to this as the first major feature of 1.1. There are still some outstanding questions that I have in regards to directories, from my previous post. Have you had any chance to think of how we might go about this?

geoffjentry:
@cjllanwarne What do you want to do with this?

cjllanwarne:
@geoffjentry I want to become the change I want to see... which in this case means having this big feature working in real code before submitting it to an up/down vote and then a rush to implement (and hope we got it right).

I think the comments so far meant that this is now ready to prototype. So I want to:
* Migrate this target the `development` branch
* Prototype it to
  * Iron out any unforeseen issues 
  * Let people try it out and suggest improvements
* Return here for the vote.

WDYT?

patmagee:
@cjllanwarne @geoffjentry  The one question I have is about how to specify directory inputs? Should a directory always be `packaged` in a `tar` or some other sort of archive? Or can a user define a path (either local or in the cloud) and that path will simply be recursively copied.

There are certain implications in terms of performance for all implementing engines if we allow generic paths ie, the engine would need to recursively list all of the files, then copy them over.  This should certainly be considered if we are going to implement it.

The one thing a `tar` archive would easily allow is for certain engine functions such as  `size` to work on it quite easily.  The other thing a `tar` archive would allow is better preservation of the directory structure as the `Directory` gets moved between sources

geoffjentry:
@patmagee If I understood you correctly that seems like an implementation detail?

patmagee:
@geoffjentry yes it really is. And I I do not think we really need to figure this out here, but With a feature like `Directory` there are enough performance considerations for us to at least consider how this would practically be done in an implementation sense.  If we add the feature to the WDL spec, but it is essentially unusable because of either performance or difficulty in implementation, then its really a dead feature

antonkulaga:
> have is about how to specify directory inputs? Should a directory always be packaged in a tar or some other sort of archive? Or can a user define a path (either local or in the cloud) and that path will simply be recursively copied.

Guys, please, DO NOT COPY IT! It can happen that the tool needs a folder to write inside of it and do file-discovery, and those files can be many GBs, the last thing I want is to waste a lot of time and storage on copying many GBs of data in the workflow!

geoffjentry:
@antonkulaga we're talking about the specification of WDL, not implementation details here. WDL should not be opinionated on this matter.

@patmagee From a Cromwell perspective I'd advocate implementing it more similar to how CWL works (which **can** be oddly specific at times) since then we'd be able to take advantage of the WOM abstraction

cjllanwarne:
Closing this version of the PR - please see the updated version, properly rebased on master and making changes to the correct SPEC.md: #241

PR #175, opened 2018-01-03
 File input clarifications

cjllanwarne:
Clarifies how files are treated and localized when supplied as inputs

(Previously part of #173)
 
The biggest addition can be seen in situ here: https://github.com/cjllanwarne/wdl/blob/cjl_file_inputs/versions/draft-3/SPEC.md#input-declarations


cjllanwarne:
@orodeh could you quickly 👍 that this represents your implementation correctly? 

If so, @geoffjentry I think it should be ok for voting as I think it falls under the umbrella of "existing behavior clarification"

orodeh:
This is good. The only suggestion I have is to rephrase this sentence: 
"If two or more files in a versioning filesystem might have the same name ...". 

👍

cjllanwarne:
@orodeh updated to:
> Two or more files versions of a file in a versioning filesystem might have the same name and come from the same directory. In that case the following special procedure must be used to avoid collision

geoffjentry:
:+1:

aprabhak2:
👍

patmagee:
👍

orodeh:
👍

geoffjentry:
Passes 4-0, as it's a clarification it's being merged immediately

Issue #158, opened 2017-11-12
The need for an "Unlocalized File" type


yfarjoun:
wdl writers currently have two options when dealing with files:

1. Use a File type and have the file "localized" and sanitized (WRT to the original name of the file).
2. Use a String type, and access the raw URL directly (if you can), and deal with resulting issues (see below).

<separator between lists>

1. Is good for small files and "dumb" programs. It also makes call-caching possible as the MD5 of the file can be stored in a DB
2. Is a good idea when accessing many small parts of large files (say combining 80K vcfs). It requires smarter programs (that know how to interact with an index and to talk to the requisite filesystem, be it s3:// gs:// or http://. The problem with this option is that call-caching is effectively disabled here, as the variable will have a different raw url with every invocation of the task.

Since more and more of our pipelines are trying to use String instead of File as a cost-saving measure, I think it makes sense to see how we could enable call-caching without localizing a file. I have some ideas:

1. Have some decoration on a File type indicate that it should be not localized and that the raw string should be used instead. This should enable call-caching without localization, as the implementations should be able to md5 the underlying file)
2. In the run-time attributes (or elsewhere), indicate which variables shouldn't be localized (ugly)
3. Add a new type, say `RemoteFile` which acts like the solution in 1 but without a decoration.
4. Allow a decoration (mutable?) to indicate that changes in this variable should not invalidate a cache, then one could pull in only the MD5 of the file (which would be cached) as an additional variable from a separate task and and the String pointer to the file as `mutable`, thus guarantee that if the file changes, the task is run and the cache is pointing to the right place (possibly easiest to implement, but hardest to maintain as a wdl author)

Other ideas or thoughts on the matter?

abaumann:
+1 - also needed for output files where a tool might write/stream directly to another location (rather than a file local to the VM being delocalized and added to call caching)

geoffjentry:
I'll add that I've seen this request countless times, there's clearly a need for it. People suggesting it always have a different name for the concept (and often declare that previous names are awful and wrong, go figure), but the idea always remains the same -> essentially something which is treated as a `File` within WDL but a `String` for command purposes 

If someone wants to make a concrete proposal I'd be all for it existing.

patmagee:
Before we get to the proposal, I would Just suggest that we use option `3` over any type of Syntactic decoration. We do not use decorators in WDL as of yet, and introducing them would not only add complexity for the user at this time, but would probably add quite significant complexity to implementors.

[...]

PR #177, opened 2018-01-12
Proposed FileURI changes

LeeTL1220:
Begins the process for a new data type that can handle a local file or a URI (such as `gs://...`) seamlessly.

Addresses:

https://github.com/openwdl/wdl/issues/158

https://github.com/broadinstitute/cromwell/issues/2564

### DIFF
--- a/versions/draft-3/SPEC.md
+++ b/versions/draft-3/SPEC.md
@@ -248,9 +248,7 @@ Float f = 27.3             # A floating point number
 Boolean b = true           # A boolean true/false
 String s = "hello, world"  # A string value
 File f = "path/to/file"    # A file
-FileURI u1 = "gs://bucket/file" # URI for Google cloud storage file
-FileURI u2 = "path/to/file"    # URI for local file
-FileURI u3 = "file:///path/to/file"    # URI for local file 
+FileURI u1 = "file:///path/to/file"    # URI for local file 
 ```
 
 In addition, the following compound types can be constructed, parameterized by other types. In the examples below `P` represents any of the primitive types above, and `X` and `Y` represent any valid type (even nested compound types):
@@ -279,9 +277,23 @@ For more information on type and how they are used to construct commands and def
 
 #### Explanation of FileURI
 
-The FileURI represents a File that could be either a local file or a file in an object store.  
+The FileURI represents a path to a file which could be either local or in a remote object store.  
 
-Sometimes users do not wish to localize an input file, unless it is on a local filesystem (e.g. the code being run in the task can read an object store directly).  This functionality cannot be replicated with `File` and `String` without changing the WDL itself to accommodate the differing reading paradigms.  The `FileURI` can be used by WDL authors to enable direct reading from object stores when their task supports it.  
+Sometimes users wish to track an input as a file, but do not wish to localize or delocalize into the task execution directory. Instead, they wish to make sure:
+
+* If an input:
+  * That the file exists before the task begins
+  * The file is available to access within the task  
+* If an output
+  * The file is created by the time the task completes.
+
+The `FileURI` therefore enables file tracking by the engine and direct reading from object stores without using space in the execution directory to store the file.
+
+Other examples of FileURI:
+```wdl
+FileURI u1 = "gs://bucket/file" # URI for Google cloud storage file
+FileURI u2 = "path/to/file"    # URI for local file
+```
 
 ### Fully Qualified Names & Namespaced Identifiers
 
@@ -518,7 +530,6 @@ Below are the valid results for operators on types.  Any combination not in the 
 |`String`|`>=`|`String`|`Boolean`||
 |`String`|`<`|`String`|`Boolean`||
 |`String`|`<=`|`String`|`Boolean`||
-|`FileURI`|`+`|`FileURI`|`FileURI`|Append FileURI paths|
 |`FileURI`|`==`|`FileURI`|`Boolean`||
 |`FileURI`|`!=`|`FileURI`|`Boolean`||
 |`FileURI`|`+`|`String`|`FileURI`||
@@ -723,8 +734,8 @@ task t {
 
 ##### FileURI localization
 `FileURI` inputs also follow specific localization rules:
-- If the URI in the FileURI is a file mounted on the filesystem (e.g. `file:///home/foo/test.txt` or `/home/foo/test.txt`), then the File referenced at must be made accessible to the job at execution time by the engine.
-- If the URI points to an object storage URI (e.g. `gs://my-bucket/test.txt` for Google Cloud Storage) then FileURIs are essentially treated as Strings as far as localization.
+- If the URI in the FileURI is a file mounted on the filesystem (e.g. `file:///home/foo/test.txt` or `/home/foo/test.txt`), then the File referenced must be made accessible to the job at execution time by the engine.
+- If the URI points to an object storage URI (e.g. `gs://my-bucket/test.txt` for Google Cloud Storage) then an existence and accessibility check should be made but no further action is required for localization.
 
 ##### Special Case: Versioning Filesystems
 Two or more versions of a file in a versioning filesystem might have the same name and come from the same directory. In that case the following special procedure must be used to avoid collision:
--- a/versions/draft-3/SPEC.md
+++ b/versions/draft-3/SPEC.md
@@ -249,7 +249,8 @@ Boolean b = true           # A boolean true/false
 String s = "hello, world"  # A string value
 File f = "path/to/file"    # A file
 SFile u1 = "gs://bucket/file" # URI for Google cloud storage file
-SFile u2 = "/path/to/file"    # URI for local file
+SFile u2 = "path/to/file"    # URI for local file
+SFile u3 = "file:///path/to/file"    # URI for local file 
 ```
 
 In addition, the following compound types can be constructed, parameterized by other types. In the examples below `P` represents any of the primitive types above, and `X` and `Y` represent any valid type (even nested compound types):
@@ -705,6 +706,7 @@ task t {
 ```
 
 #### Input Localization
+##### File localization
 `File` inputs must be treated specially since they require localization to within the execution directory:
 - Files are localized into the execution directory prior to the task execution commencing. 
 - When localizing a `File`, the engine may choose to place the file wherever it likes so long as it accords to these rules:
@@ -712,6 +714,11 @@ task t {
   - Two input files with the same name must be located separately, to avoid name collision.
   - Two input files which originated in the same storage directory must also be localized into the same directory for task execution (see the special case handling for Versioning Filesystems below).
 - When a WDL author uses a `File` input in their [Command Section](#command-section), the fully qualified, localized path to the file is substituted into the command string.
+
+##### SFile localization
+`SFile` inputs also follow specific localization rules:
+- If the URI in the SFile is a file mounted on the filesystem (e.g. `file:///home/foo/test.txt` or `/home/foo/test.txt`), then SFiles are localized exactly the same as Files.  See [localization of files](file-localization) above.
+- If the URI points to an object storage URI (e.g. `gs://my-bucket/test.txt` for Google Cloud Storage), SFiles are not localized.
 
 ##### Special Case: Versioning Filesystems
 Two or more versions of a file in a versioning filesystem might have the same name and come from the same directory. In that case the following special procedure must be used to avoid collision:
--- a/versions/draft-3/SPEC.md
+++ b/versions/draft-3/SPEC.md
@@ -248,7 +248,7 @@ Float f = 27.3             # A floating point number
 Boolean b = true           # A boolean true/false
 String s = "hello, world"  # A string value
 File f = "path/to/file"    # A file
-FileURI u1 = "file:///path/to/file"    # URI for local file 
+FileURL u1 = "file:///path/to/file"    # URI for local file 
 ```
 
 In addition, the following compound types can be constructed, parameterized by other types. In the examples below `P` represents any of the primitive types above, and `X` and `Y` represent any valid type (even nested compound types):
@@ -275,9 +275,9 @@ For more details on the postfix quantifiers, see the section on [Optional Parame
 
 For more information on type and how they are used to construct commands and define outputs of tasks, see the [Data Types & Serialization](#data-types--serialization) section.
 
-#### Explanation of FileURI
+#### Explanation of FileURL
 
-The FileURI represents a path to a file which could be either local or in a remote object store.  
+The FileURL represents a path to a file which could be either local or in a remote object store.  
 
 Sometimes users wish to track an input as a file, but do not wish to localize or delocalize into the task execution directory. Instead, they wish to make sure:
 
@@ -287,12 +287,12 @@ Sometimes users wish to track an input as a file, but do not wish to localize or
 * If an output
   * The file is created by the time the task completes.
 
-The `FileURI` therefore enables file tracking by the engine and direct reading from object stores without using space in the execution directory to store the file.
+The `FileURL` therefore enables file tracking by the engine and direct reading from object stores without using space in the execution directory to store the file.
 
-Other examples of FileURI:
+Other examples of FileURL:
 ```wdl
-FileURI u1 = "gs://bucket/file" # URI for Google cloud storage file
-FileURI u2 = "path/to/file"    # URI for local file
+FileURL u1 = "gs://bucket/file" # URI for Google cloud storage file
+FileURL u2 = "path/to/file"    # URI for local file
 ```
 
 ### Fully Qualified Names & Namespaced Identifiers
@@ -530,11 +530,11 @@ Below are the valid results for operators on types.  Any combination not in the 
 |`String`|`>=`|`String`|`Boolean`||
 |`String`|`<`|`String`|`Boolean`||
 |`String`|`<=`|`String`|`Boolean`||
-|`FileURI`|`==`|`FileURI`|`Boolean`||
-|`FileURI`|`!=`|`FileURI`|`Boolean`||
-|`FileURI`|`+`|`String`|`FileURI`||
-|`FileURI`|`==`|`String`|`Boolean`||
-|`FileURI`|`!=`|`String`|`Boolean`||
+|`FileURL`|`==`|`FileURL`|`Boolean`||
+|`FileURL`|`!=`|`FileURL`|`Boolean`||
+|`FileURL`|`+`|`String`|`FileURL`||
+|`FileURL`|`==`|`String`|`Boolean`||
+|`FileURL`|`!=`|`String`|`Boolean`||
 ||`-`|`Float`|`Float`||
 ||`+`|`Float`|`Float`||
 ||`-`|`Int`|`Int`||
@@ -732,9 +732,9 @@ task t {
   - Two input files which originated in the same storage directory must also be localized into the same directory for task execution (see the special case handling for Versioning Filesystems below).
 - When a WDL author uses a `File` input in their [Command Section](#command-section), the fully qualified, localized path to the file is substituted into the command string.
 
-##### FileURI localization
-`FileURI` inputs also follow specific localization rules:
-- If the URI in the FileURI is a file mounted on the filesystem (e.g. `file:///home/foo/test.txt` or `/home/foo/test.txt`), then the File referenced must be made accessible to the job at execution time by the engine.
+##### FileURL localization
+`FileURL` inputs also follow specific localization rules:
+- If the URI in the FileURL is a file mounted on the filesystem (e.g. `file:///home/foo/test.txt` or `/home/foo/test.txt`), then the File referenced must be made accessible to the job at execution time by the engine.
 - If the URI points to an object storage URI (e.g. `gs://my-bucket/test.txt` for Google Cloud Storage) then an existence and accessibility check should be made but no further action is required for localization.
 
 ##### Special Case: Versioning Filesystems
@@ -1967,12 +1967,12 @@ WDL values can be created from either JSON values or from native language values
 |         |String-like||
 |         |`String`|Identity coercion|
 |         |`File`||
-|         |`FileURI`||
+|         |`FileURL`||
 |`File`   |JSON String|Interpreted as a file path|
 |         |String-like|Interpreted as file path|
 |         |`String`|Interpreted as file path|
 |         |`File`|Identity Coercion|
-|         |`FileURI`|Interpreted as file path|
+|         |`FileURL`|Interpreted as file path|
 |`Int`    |JSON Number|Use floor of the value for non-integers|
 |         |Integer-like||
 |         |`Int`|Identity coercion|
@@ -1982,11 +1982,11 @@ WDL values can be created from either JSON values or from native language values
 |`Boolean`|JSON Boolean||
 |         |Boolean-like||
 |         |`Boolean`|Identity coercion|
-|`FileURI`    |JSON String|Interpreted as a FileURI|
-|         |String-like|Interpreted as a FileURI|
-|         |`String`|Interpreted as a FileURI|
-|         |`File`|Interpreted as a FileURI|
-|         |`FileURI`|Identity Coercion|
+|`FileURL`    |JSON String|Interpreted as a FileURL|
+|         |String-like|Interpreted as a FileURL|
+|         |`String`|Interpreted as a FileURL|
+|         |`File`|Interpreted as a FileURL|
+|         |`FileURL`|Identity Coercion|
 |`Array[T]`|JSON Array|Elements must be coercable to `T`|
 |          |Array-like|Elements must be coercable to `T`|
 |`Map[K, V]`|JSON Object|keys and values must be coercable to `K` and `V`, respectively|
@@ -2191,7 +2191,7 @@ The `read_int()` function takes a file path which is expected to contain 1 line 
 
 If the entire contents of the file can not be read for any reason, the calling task or workflow will be considered to have failed. 
 
-## String read_string(String|File|FileURI)
+## String read_string(String|File|FileURL)
 
 The `read_string()` function takes a file path which is expected to contain 1 line with 1 string on it.  This function returns that string.
 
@@ -2400,9 +2400,9 @@ And `/local/fs/tmp/map.json` would contain:
 }
 ```
 
-## Float size(File|FileURI, [String])
+## Float size(File|FileURL, [String])
 
-Given a `File` (or `FileURI`) and a `String` (optional), returns the size of the file in Bytes or in the unit specified by the second argument.
+Given a `File` (or `FileURL`) and a `String` (optional), returns the size of the file in Bytes or in the unit specified by the second argument.
 
 ```wdl
 task example {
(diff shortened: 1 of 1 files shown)


versions/draft-3/SPEC.md
```diff
@@ -247,6 +247,8 @@ Float f = 27.3             # A floating point number
 Boolean b = true           # A boolean true/false
 String s = "hello, world"  # A string value
 File f = "/path/to/file"   # A file
+SFile u1 = "gs://bucket/file" # URI for Google cloud storage file
```
yfarjoun: I don't like `SFile` as the datatype.

I think that `DataObject` might be better.

cjllanwarne:
When (and assuming) https://github.com/openwdl/wdl/pull/175 merges, I think it'll be good to add a section on how this `SFile` would be treated differently from regular `File` with regards to localization.

I'm not a big fan of the name `SFile` since it's not immediately obvious to me what that means. Previous suggestions like `RemoteFile`, `FileUri`, `Uri`, `UnlocalizedFile`, etc are more verbose but less opaque IMO.

yfarjoun:
Regarding the name of this Type:
- I agree that SFile is annoying...it means "SmartFile"...but that doesn't help much either.
- "URI" might imply that we also include http:// or ftp:// or ssh:// etc. which we are not sure if we do or do not.
- "Remote" is not true, since the point is that it could be remote or not.
- "Unlocalized" is similarly not true, since it will be localized if it's a file:// and the backend is local.

That's why I suggested DataObject (I'd suggest Object, but that's taken)

geoffjentry:
Regardless of the name (and i think `SFile` is awful), this needs a **lot** more documentation before proceeding.

LeeTL1220:
Now that #175 is merged, I'll extend docs soon.

`SFile` is awful, but I just needed something we could find and replace easily.

LeeTL1220:
I'm not a huge fan of `DataObject`.  

I like having the word `File` in the type.

LeeTL1220:
I could get behind `FileUri`

cjllanwarne:
@LeeTL1220  Let's select the correct option out of `FileUri`, `FileUrl` and `FileUrn`. Based on https://danielmiessler.com/study/url-uri/ it looks like `URL` is most appropriate since we don't want URNs and URI is a subset of URL and URN?

LeeTL1220:
Now `FileURI`.  @cjllanwarne I have answered your comments that we had in conversation.  Feel free to review.

versions/draft-3/SPEC.md
```diff
@@ -274,6 +277,12 @@ For more details on the postfix quantifiers, see the section on [Optional Parame
 
 For more information on type and how they are used to construct commands and define outputs of tasks, see the [Data Types & Serialization](#data-types--serialization) section.
 
+#### Explanation of FileURI
+
+The FileURI represents a File that could be either a local file or a file in an object store.
```
cjllanwarne: I'd say `... represents a path to a file which could be either local or in a remote object store`

versions/draft-3/SPEC.md
```diff
@@ -274,6 +277,12 @@ For more details on the postfix quantifiers, see the section on [Optional Parame
 
 For more information on type and how they are used to construct commands and define outputs of tasks, see the [Data Types & Serialization](#data-types--serialization) section.
 
+#### Explanation of FileURI
+
+The FileURI represents a File that could be either a local file or a file in an object store.  
+
+Sometimes users do not wish to localize an input file, unless it is on a local filesystem (e.g. the code being run in the task can read an object store directly).  This functionality cannot be replicated with `File` and `String` without changing the WDL itself to accommodate the differing reading paradigms.  The `FileURI` can be used by WDL authors to enable direct reading from object stores when their task supports it.
```
cjllanwarne: I'd say 
```
Sometimes users wish to track an input as a file, but do not wish to localize or delocalize into the task execution directory. Instead, they wish to make sure:

* If an input:
  * That the file exists before the task begins
  * The file is available to access within the task  
* If an output
  * The file is created by the time the task completes.

The `FileURI` therefore enables file tracking by the engine and direct reading from object stores without using space in the execution directory to store the file.
```

versions/draft-3/SPEC.md
```diff
@@ -248,6 +248,9 @@ Float f = 27.3             # A floating point number
 Boolean b = true           # A boolean true/false
 String s = "hello, world"  # A string value
 File f = "path/to/file"    # A file
+FileURI u1 = "gs://bucket/file" # URI for Google cloud storage file
+FileURI u2 = "path/to/file"    # URI for local file
+FileURI u3 = "file:///path/to/file"    # URI for local file
```
cjllanwarne: I'd only have a single example here and include the other possibilities in the description section below

versions/draft-3/SPEC.md
```diff
@@ -509,6 +518,12 @@ Below are the valid results for operators on types.  Any combination not in the
 |`String`|`>=`|`String`|`Boolean`||
 |`String`|`<`|`String`|`Boolean`||
 |`String`|`<=`|`String`|`Boolean`||
+|`FileURI`|`+`|`FileURI`|`FileURI`|Append FileURI paths|
```
cjllanwarne: I don't think addition makes sense. I know `File` has it (which also has never made sense to me), but unless this is something you can see how you'll want to use, I'd just remove it.

versions/draft-3/SPEC.md
```diff
@@ -705,6 +721,11 @@ task t {
   - Two input files which originated in the same storage directory must also be localized into the same directory for task execution (see the special case handling for Versioning Filesystems below).
 - When a WDL author uses a `File` input in their [Command Section](#command-section), the fully qualified, localized path to the file is substituted into the command string.
 
+##### FileURI localization
+`FileURI` inputs also follow specific localization rules:
+- If the URI in the FileURI is a file mounted on the filesystem (e.g. `file:///home/foo/test.txt` or `/home/foo/test.txt`), then the File referenced at must be made accessible to the job at execution time by the engine.
```
cjllanwarne: typo, should be `"...the File referenced must..."`

versions/draft-3/SPEC.md
```diff
@@ -705,6 +721,11 @@ task t {
...
+- If the URI points to an object storage URI (e.g. `gs://my-bucket/test.txt` for Google Cloud Storage) then FileURIs are essentially treated as Strings as far as localization.
```
cjllanwarne: I wouldn't even call out `String` here (`String` localization is "do nothing", but we might as well say that directly). eg:

`"If the URI [...] Cloud Storage) then an existence and accessibility check should be made but no further action is required for localization.`

versions/draft-3/SPEC.md
```diff
@@ -2152,7 +2180,7 @@ The `read_int()` function takes a file path which is expected to contain 1 line
 
 If the entire contents of the file can not be read for any reason, the calling task or workflow will be considered to have failed. 
 
-## String read_string(String|File)
+## String read_string(String|File|FileURI)
```
cjllanwarne: There are more `read_x` functions which should also support `FileURI`

mckinsel:
Hey so what's supposed to happen in a WDL like this:

```
task my_task {
  File my_file
  File my_fileuri
  File my_string
  command {...}
  ...
}

workflow my_workflow {
  File my_file
  FileURI my_fileuri
  String my_string
  
  call my_task {
    input:
      my_file=my_file, my_fileuri=my_fileuri, my_string=my_string
  }
}
```
I think this would be a valid WDL, but I'm not sure what a WDL executor is required to do.

cjllanwarne:
@mckinsel 

* I think the following statements can be inferred from elsewhere in the WDL spec:

Task level:
The semantics of a `File` input to a task stay the same. So `my_task`'s three `File` inputs will be localized to the filesystem and their localized paths passed into the command block.

Workflow level:
In the workflow, you're coercing a `String` and a `FileURI` into `File`s as you pass them into the task.

* I think the following would be worth spelling out in this PR:

In both coercions from `X` to `File` the net effect will presumably be "take the string value and use it as the path to this `File` you'll be tracking"

mckinsel:
Hey thanks @cjllanwarne. I think "use it as the path" is doing a lot of work there, and that's the part that worries me. In WDLs today you can, at the workflow level, coerce a `String` to a `File` as in the example WDL above.  So the task only sees a `File` and only has to worry about its local
path. But, that shifts all the work on to the WDL executor, which at any time might be expected to know how to turn an arbitrary `String` into a local `File`.

That probably works out okay for most Broad users because cromwell+google can treat "gs://..." as a `String` or a `File` task input, and the whole thing just works. But in other environments, it's a headache, and it's come up a couple times when we've tried to port Human Cell Atlas WDLs. This PR really doubles down on the problem by creating a new and useful `String`-like type for referring to files that has ambiguous coercion and localization behavior.

For example, consider this WDL:
```
task my_task {
  File my_file
  command <<<
    wc -l "${my_file} | awk '{print \$1}' > line.count
  >>>
  ...
}

workflow my_workflow {
  File my_file

  call my_task {
    input: my_file=my_file
  }
...
}
```

Now suppose I want to run this WDL using Cromwell and a local backend. If I get an input JSON that has "gs://..." for `my_file`, that's not going to work because cromwell+local doesn't know how to localize a "gs://..." JSON string. But, I could separately download the file from gcp and replace the google storage string with a local file path, and it will work.

But this PR would encourage WDLs like this:

```
task my_task_0 {
  File my_file
  command <<<
    wc -l "${my_file} | awk '{print \$1}' > line.count
  >>>
  ...
}

task my_task_1 {
  FileURI my_fileuri
  command <<<
    gsutil cat ${my_file} | wc -l  > line.count
  >>>
  ...
}

workflow my_workflow {
  FileURI my_fileuri

  call my_task_0 {
    input: my_file=my_fileuri
  }

  call my_task_1 {
    input: my_fileuri=my_fileuri
  }
...
}
```
I'm pretty sure this would work with cromwell+google, but I don't think there's anything I could do to run this WDL with cromwell+local. I'd have to edit the WDL.

cjllanwarne:
@mckinsel I think I see your point - WDLs can certainly be written in an unportable way. That's neither new nor unique to WDL but I can see how this PR might exacerbate the issue by appearing to endorse it. 

My personal opinion is we shouldn't dismiss this or any other PR _only_ because people might shoot themselves in the foot with it. There are real world, portable WDLs that really do want URL inputs which would make use of this. I would hope that education on what makes a portable WDL and code reviews should be able to help with the rest.

### Pros of this PR:

* Can take full advantage of tools like gatk4 which take `URL` inputs to avoid the cost of localizing entire BAMs only to use the first 1/50th of it.
* For tools with URL inputs, it's nice for the engine to know that so that it can guarantee visibility and readability of the URL from the task context.
* For tools with URL outputs, it's nice to be able to guarantee that the file was created correctly.
* With an engine like Cromwell, call caching on tasks with file URL `String` inputs as though they are plain strings is very awkward. It's possible to get false negatives (the URL is different this time round) and false positives (the file at the same URL has changed since last time).

mckinsel:
@cjllanwarne yeah definitely see the pros of this PR, and having a `FileURI` type seems better than just overloading the meaning of `String`.

Let me describe the scenario that's actually motivating my concern here. The Human Cell Atlas is writing WDLs, they're going to run on something that looks like workbench, but they also have to be "portable". To show they're portable, I run them in other environments and see if they work. Suppose a WDL runs successfully in workbench but fails in some other environment. Do I say that WDL is not portable and can't be part of the HCA, or do I say that the WDL is fine and the other environment is the problem?

In the last WDL above, for example, the WDL author could point to the type coercions permitted in the WDL spec and say their WDL follows the spec perfectly; cromwell+local is just a non-conforming WDL executor. But reading between the lines of your latest comment, it sounds like you would disagree with that argument. There's the WDL spec, and then there's "what makes a portable WDL". Or there's WDL, a language that emphasizes flexibility, and some sort of RWDL, a language you could write conformance tests for. This PR has elements in both, but it's not clear where the line would be drawn.

versions/draft-3/SPEC.md
geoffjentry: What about if it's `http://`?

geoffjentry:
A few thoughts here ...

- Agree that "this allows one to write a non-portable WF" shouldn't in and of itself be disqualifying. We should actively encourage WDL authors to write their stuff in a way which **is** portable, and bring much shame to those who do not.
- It's been said for a while now, but there's a need for a "if one wants to write portable workflows, abide by the following rules" document

@mckinsel I think it's perfectly acceptable for HCA to reject workflows which are written in a way which are non-portable. For instance, many of the WDLs coming out of our department are highly specific to Google, IMO that's horrible.

orodeh:
I like the `FileURI` type, because it adds an important building block to the WDL type system: remote files. This is  a practical concern, because the GATK4 pipeline uses the Google NIO library, and encodes a remote file as a `String`. Porting that pipeline proved difficult. Remote files are nice, because they could potentially be accessed without mounting the remote filesystem, and without the remote side being a filesystem at all. They also do not require local disk space.

My concern is not with the `FileURI` type, but with the auto-conversion between `String` and `File`/`FileURI` types. This causes many scripts to require porting, which defeats one of WDL's main themes. While this conversion conversion useful for string manipulation, perhaps we can try to limit it. For example, `dxWDL` gives warnings on encountering such conversions inside workflows (but not tasks).

I think improving the type safety of WDL scripts would be very helpful. As a practical example, due to regulatory concerns, files on the dnanexus platform require strong permissions. A task can only read files that were explicitly given to it as input. If it tries to convert a random string to a file, it will be denied access. As it stands, many workflows perform such conversions, they work on Cromwell, but not with dxWDL.

geetduggal:
My 2 cents: I completely agree that smth like `FileURI` is better than a string :).  However, I also think it creates issues when specifying a workflow in a portable way as mentioned in the examples @mckinsel gave above (e.g. ideally if my workflow or task input takes a file, I can specify it in multiple ways and the execution engine will handle it appropriately).  In the spirit of this, I would almost prefer it to just stay as 'File', but execution engines can be enhanced to be smarter or supports certain 'subtypes' of files.  For example, one could imagine that if a supported prefix for a URI is provided (e.g. http://, gc://, s3://, dos://, ... ) it is interpreted by the execution engine as such, otherwise it is treated as a 'File as usual' (If I understand right, and I could be totally wrong here, I believe CWL takes spiritually takes this sort of approach with URIs: http://www.commonwl.org/v1.0/CommandLineTool.html#File).    Finally, different executors could provide support for different subsets of URI prefixes with perhaps a couple being required to be supported like the import statements (https://github.com/openwdl/wdl/blob/master/versions/draft-3/SPEC.md#import-statements).  In this case, the WDL workflow is always portable however different sets of file inputs to run a WDL may not be supported by all execution engines. In this case the user can simply default to providing a 'File as usual' as an input rather than than use a more convenient or optimized use of a URI.

orodeh:
@LeeTL1220 Reading through this carefully, I realized I missed something important. How is a `FileURI` materialized in a task? For example, in task `bam_head`, how does the engine create a POSIX file descriptor for the `head` command to use?

```
task bam_head {
  FileURI reads.bam
  command {
      head ${reads.bam}
  }
  output {
     String out = stdout()
  }
}
```

geetduggal:
Minor note in addition to @orodeh's comment: do we, _by definition_, assume the file descriptor can be handled in a POSIX way where, e.g. even reads of specific ranges of a file can be performed over the network?  This 'lazy' handling of files is potentially very useful powerful but it does seem like an implementing it can be non-trivial as it lends itself to a FUSE-ish type of concept.

LeeTL1220:
@orodeh As WDL author, you would not use FileURI for any task where you always want the file localized.

LeeTL1220:
@geetduggal We had envisioned a type that is backed by a file, but that the task should otherwise treat as a string.  In other words, the tools in the task itself will need to know how to handle a URL.

versions/draft-3/SPEC.md
LeeTL1220: Done

versions/draft-3/SPEC.md
LeeTL1220: Done

versions/draft-3/SPEC.md
LeeTL1220: Done

versions/draft-3/SPEC.md
LeeTL1220: Removed.

Done

versions/draft-3/SPEC.md
LeeTL1220: Done

versions/draft-3/SPEC.md
LeeTL1220: Not addressing `http://` for now.  Done, otherwise.

@cjllanwarne How would you want to handle http?  I would limit FileURI to local filesystems and cloud storage.  However, that can be confusing for end users, so I am open to suggestions.  Or maybe have `http://` be handled by `File`?

versions/draft-3/SPEC.md
LeeTL1220: @cjllanwarne Apologies, if I am missing something obvious, but I could not find any.  

No action.

geoffjentry:
Hi everyone - Between the comments & some behind the scenes conversation, I know that this PR is still a lot more contentious than we usually like to see prior to going to vote (which really is typically a rubber stamp). However, do to the looming deadline and a desire by some to see this in `1.0` this is now moved to `Voting Active`

**NB**: Voting is on the PR **as-is**, if your concerns are large enough that you're not willing to see it in that state, do not feel bad about voting "no", that's what voting is for :)

orodeh:
👎

mckinsel:
👎

yfarjoun:
👍

orodeh:
@geoffjentry @mckinsel 
Since voting is now active, this is a comment on a possible way to make this less controversial, assuming it was resubmitted. 

I thought about this some more. In the short term, if you want to mark a string as a FileURL, you can use the parameter_meta section. This would be similar to how dxWDL marks [streaming files](https://github.com/dnanexus/dxWDL/blob/master/doc/ExpertOptions.md#extensions). 

In the WDL standard, here is an idea: 
1) A FileURL type should support http operations, such as curl, or wget. 
2) There will be standard library functions to convert FileURL to File, and FileURL to a string. These could be overridden by an engine if needed. 
3) Auto-coercion between Files, Strings, and FileURLs will be limited.

I think this will make the FileURL type more concrete, and more difficult to misuse.

geetduggal:
Thanks for the discussion! I generally agree with @orodeh's suggestion.  IMHO in the short term an annotation is sufficient, and in the longer term, introducing a FileURL type that guarantees that the item in there is an HTTP URL seems nice.   The easy case is if a user provides an HTTP URL as an input, any operation you can do to read the HTTP file is supported (e.g. curl, wget, ...).  Different execution engines can support other types of inputs.  For example, Cromwell and DNAnexus could also support google cloud URLs.  In this case, the engine will do the work to translate it to an HTTP URL but nothing in the application code needs to be changed.

geoffjentry:
:-1:

patmagee:
👎

LeeTL1220:
We need something here so that we can write WDL that is portable, can leverage bucket streaming, and allows call caching.  Currently, doing all of this is impossible.  `FileURI` addresses all of these.

I'm open to new suggestions, but, if this gets "no" voted, then we have no solution to a known (and important) issue.

vdauwera:
:+1: from me to getting a working solution out there. I can live with having to document caveats and making it clear that this limits portability (if I understood correctly)

cjllanwarne:
After consideration I'm 👍 on this.

I appreciate the concerns raised in this PR and agree that some additional clarifications around this would be worthwhile, but this seems strictly better than encoding URLs as `String` as we are forced to do today.

* This PR doesn't force any engines to support new filesystems - just like a `File` input which is in an unknown filesystem, the engine would fail. This can quite happily mean anything other than a local `file://` URL could be considered a failure in some engines.
  * In fact this PR allows the engine to fail faster, and with a more friendly error message, than using the existing encoding of URLs as `String`. Today we can only fail at the last second when a task is given a filesystem URL it cannot access. Letting the engine pre-check that is valuable!
* I would support a future PR which removes the String/File interactions but I don't see that as a blocker to this PR.
* I would also support a way to indicate to the engine that a tool in a task requires a specific subset of URL types so that an engine can decide at submit-time whether it will ever be able to run a given WDL. But again, that doesn't feel like a blocker to this PR.
* The use of `parameter_meta` would be a fine short-term fix but in my opinion would be a huge amount of boilerplate. Eg imagine the scheme to decide which of these String fields are really FileURLs:
```wdl
# We want the left hand side of the value of the map to be treated like a FileURL:
Map[String, Pair[String, String]] in = ...
```
We'd need to end up with something like this which will be tricky to parse correctly and IMO moves us a long way away from WDL as an easy-to-read, easy-to-understand and easy-to-write language:
```wdl
parameter_meta {
  in: {
    values: {
      left: {
        fileurl: true
      }
    }
  }
}
```

My final though: I would be happy to have this voted in today but waiting for implementation until the next version of WDL, in order to give us time to prototype and iron out any unforeseen consequences, and also to let people submit these clarifications as part of the same WDL version that contains this new feature.

eitanbanks:
👍

SHuang-Broad:
👍

ldgauthier:
:+1:

mwalker174:
👍

lucidtronix:
👍

jonn-smith:
+1

jonn-smith:
👍

jacarey:
:+1:

lbergelson:
👍

asmirnov239:
👍

samuelklee:
:+1::+1:

LeeTL1220:
:+1:

gbggrant:
👍

LeeTL1220:
To the vote tallier:  There may be more votes on the first comment.

eitanbanks:
In case it's not obvious, there are a whole bunch of us who really want this functionality -- and who had no idea that it was up for voting.  So when we found out, we all jumped in to vote just now.

cmnbroad:
👍

kyuksel:
👍

sooheelee:
👍🏽

mlin:
I'm concerned about having [neither rough consensus, nor running code](https://tools.ietf.org/html/rfc7282) here. I think everybody agrees there's a lot of merit in this proposal, however, I don't quite understand the time pressure to merge it into the community WDL spec at this juncture. That shouldn't block implementing it in Cromwell to the extent it's needed for Broad use cases, indeed in dxWDL we implement certain [extensions](https://github.com/dnanexus/dxWDL/blob/master/doc/ExpertOptions.md#extensions) which have been useful to us. Actually, I personally prefer to know that *someone* has actually implemented an idea like this before enshrining it in a spec. Doing so always illuminates previously dark corners.

I hear the needs around file localization and call caching which this addresses, but ideally we'd also take time to air out alternative ways of addressing them. 

1. In many cases unlocalized remote files can still be exposed to tasks as POSIX paths, e.g. with a named pipe or a FUSE mount. The former applies when the input file is to be read sequentially, which is one of our dxWDL extensions that's been pretty useful. The latter allows random access, but there's a known association between FUSE and production disasters (P < 5e-8). Both have the advantage that the task gets a POSIX path and thus, is not invited to write `gsutil cat ${fileuri} | ...` or `dx cat ${fileuri} | ...`.

2. For the specific case of slicing BAM/CRAM/VCF off cloud storage, without having to download the whole sucker, the [GA4GH htsget protocol](http://samtools.github.io/hts-specs/) addresses this exactly, and is nicely [implemented by Google Genomics](https://cloud.google.com/genomics/deprecation) among several others. 

These would be a lot of fun to discuss at length, but can someone enlighten me on the urgency to put FileURI into the community spec right now?

geoffjentry:
@mlin I agree w/ a lot of what you say here but keep in mind that nothing is truly ratified by the spec until implemented by someone. Passing the PR phase merely means that the community believes it is a good idea, if issues turn up during implementation it's still a no-go

chapmanb:
I'd also suggest more discussion on this topic and trying to coordinate with the what the CWL community is doing. I know CWL has the idea of a location (a URI) and a path (a local file): http://www.commonwl.org/v1.0/Workflow.html#File @tetron, @mr-c  -- is there any best practice on how to label and reference non-local files within CWL so that we could coordinate?

From this discussion it's clear that WDL needs standard ways of defining local files but not that we have an approach that works for multiple implementers. Given this is so popular and needed, I'd worry about getting something locked in the spec which is not fully baked and misses use cases that are not obvious until after implementations explore it. So I agree with Mike's suggestion of first implementing in a few runners to explore and then work to make it portable so it can go into the specification.

geoffjentry:
@chapmanb Much like I just told @mlin - I agree with the spirit of the message but not the misunderstanding of the process :) Implementation **is** a required aspect of true adoption, but what we don't want to support is implementation-specific WDL. Thus the path is to gain general community approval via the PR process with a double check provided by implementation.

chapmanb:
Right on, but it seems like the discussion here is driven in two diverging directions. Folks who have to implement it aren't sure this fulfills what they'll need to build something portable, and folks who want the feature asking to make it happen because it solves a need. This dichotomy argues that in case we might need to implement something temporary to help people do their analysis and then explore a bit more before locking something down in the spec. I freely admit my bias as a implementer who likes to see things in practice first, but find it hard to see how best to spec this out without more clarity from everyone who has to support it.

geoffjentry:
@LeeTL1220 At some point many moons ago you convinced me that this couldn't simply be a `URI` type but I don't remember why that is. Can you fill in that blank?

@chapmanb I suspect that the disagreement and obvious attempt at ballot box stuffing would indicate a deeper divide than actually exists. Considering this languished for months w/o attention, I'd prefer to take this opportunity to drive forward an actual conversation first, perhaps with a hangout meeting if that'd be helpful.

geoffjentry:
@orodeh A while back you asked "Reading through this carefully, I realized I missed something important. How is a FileURI materialized in a task? For example, in task bam_head, how does the engine create a POSIX file descriptor for the head command to use?"

The answer is that "it does not". The point is that in the `FileURI` case the URI is placed in the command string directly. The point of this PR is to provide a type which is a `URI` but allows an engine to know it should be treating it as a `File` for any purposes where that might be helpful (the group proposing this specifically cares about call caching in Cromwell). 

I agree that there are other ways to achieve this but I get the sense from some of the comments that perhaps the intention of the type wasn't clear enough.

mckinsel:
@geoffjentry, my understanding is that there's a clear path to implementing this PR for the cromwell+google WDL executor. The discussion here suggests that there are concerns about implementing it anywhere else. So if this PR is adopted, it will likely be "ratified" by exactly one implementation.

Isn't that the "implementation-specific WDL" that everyone wants to avoid? This PR must solve a significant problem for a lot of users, given the teeming enthusiasm for it. But surely there's a way to address that problem without risking an effective fork of the community spec.

geoffjentry:
@mckinsel if it wasn’t clear I agree with everything you just said. When I was talking about implementions I was just making it clear what the current process is (itself not set in stone if/when we deem it to be insufficient)

LeeTL1220:
@geoffjentry I don't really remember, but I think .... We were talking about cromwell at that time, not an agnostic WDL spec... You (probably) need to support localization in the case of local files for local backends.  And you also need to be able to handle md5 the file contents for call caching.

geoffjentry:
Hi everyone -

For a variety of reasons the core group has decided that we're going to close this PR for now. 

Broad folk: @cjllanwarne will work with you all to concoct a plan to go forward.

PR #178, opened 2018-01-16
Clarification of object literal syntax

cjllanwarne:
This was spotted while fixing up a parsing bug in the winstanley plugin.

It turns out that the parser (and WDL) has an until-now overlooked syntax for specifying keys in objects as distinct from "create as a map then coerce it to an object" which is what I suspect all existing WDLs are doing.

Here's a [link to the change](https://github.com/cjllanwarne/wdl/blob/cjl_object_literals/versions/draft-3/SPEC.md#object-literals), in place in the doc.

### DIFF
--- a/versions/draft-3/SPEC.md
+++ b/versions/draft-3/SPEC.md
@@ -617,7 +617,7 @@ Map[String, Int] = {"a": 1, "b": 2}
 
 Object literals are specified similarly to maps, but require an `object` keyword:
 
-```
+```wdl
 Object f = object { 
   a: 10,
   b: 11
@@ -629,7 +629,7 @@ The object keyword allows the field keys to be specified as identifiers, rather 
 #### Object Coercion from Map
 
 Objects can be coerced from map literals, but beware the following behavioral difference:
-```
+```wdl
 String a = "beware"
 String b = "key"
 String c = "lookup"
--- a/versions/draft-3/SPEC.md
+++ b/versions/draft-3/SPEC.md
@@ -22,6 +22,7 @@
     * [Function Calls](#function-calls)
     * [Array Literals](#array-literals)
     * [Map Literals](#map-literals)
+    * [Object Literals](#object-literals)
     * [Pair Literals](#pair-literals)
   * [Document](#document)
   * [Import Statements](#import-statements)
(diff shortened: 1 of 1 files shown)


orodeh:
👍

versions/draft-3/SPEC.md
```diff
@@ -612,6 +613,43 @@ Map[Int, Int] = {1: 10, 2: 11}
...
+}
+
+# What are the keys to this object?
+Object object_syntax = object { 
+  a: 10,
+  b: 11,
+  c: 12
+}
+```
+If an `Object` is specified using the object-style `Object map_syntax = object { a: ...` syntax then the keys would are `a`, `b` and `c`. If the `Object` is specified using the map-style `Object map_coercion = { a: ...` then the keys are expressions, and thus `a` will be a variable reference to the previously defined `String beware =`.
```
cjllanwarne: Make the sentence order the same as the WDL example order, otherwise it could be needlessly confusing

cjllanwarne: Should be `... to the previously defined String a = "beware"`

versions/draft-3/SPEC.md
```diff
@@ -612,6 +613,43 @@ Map[Int, Int] = {1: 10, 2: 11}
...
+Object literals are specified similarly to maps, but require an `object` keyword:
+
+```wdl
+Object f = object { 
+  a: 10,
+  b: 11
+}
+```
+
+The object keyword allows the field keys to be specified as identifiers, rather than string literals (eg `a:` rather than `"a":`).
```
cjllanwarne: `String`

versions/draft-3/SPEC.md
```diff
@@ -612,6 +613,43 @@ Map[Int, Int] = {1: 10, 2: 11}
 Map[String, Int] = {"a": 1, "b": 2}
 ```
 
+### Object Literals
+
+Object literals are specified similarly to maps, but require an `object` keyword:
```
cjllanwarne: `... require an object keyword immediately before the {:`

patmagee:
@cjllanwarne thanks for the clarifications.
I wonder though with structs if `Object` literals are something that should be Deprecated? Mind you then that makes creating object literals require a defined `Struct` somewhere.

cjllanwarne:
@patmagee I think that's probably a good end game. But right now:

- As a clarification PR, I just want to correct the spec to to record and match the state of the world right now, rather than update for new functionality. I'm happy for a "new functionality" PR to come in at some point suggesting we deprecate object literals.
- We do want to be able to make literal `Struct`s in WDL and for that we'll need expressions which generate objects (so that we can coerce them into structs):
```
MyStruct x = object {
  str: "foo",
  int: 52
}
```

You're right - I could certainly imagine a world without `object {...}` literals and without the unconstrained `Object` type but even then we'd still need something like Struct constructors and  I'd want to see that come in via a separate PR (eg something like):
```
MyStruct x = MyStruct {
  str: "foo",
  int: 52
}
```
TL;DR: I also still feel a little unsure about whether to remove `Object` and `object {...}`. I think this clarification will be a useful baseline to make any such follow-up PRs against.

geoffjentry:
:+1:

geoffjentry:
Passes on a 2-0 vote

PR #180, opened 2018-01-16
Clarify globs

cjllanwarne:
Clarifies the set of files expected to be returned by the `glob()` function, including that it might return 0, 1, or many files in the output `Array[File]`.

### DIFF
--- a/versions/draft-3/SPEC.md
+++ b/versions/draft-3/SPEC.md
@@ -906,7 +906,7 @@ The array of `File`s returned is the set of files found by the bash expansion of
 
 In other words, you might think of `glob()` as finding all of the files (but not the directories) in the same order as would be matched by running `echo <glob>` in bash from the task's execution directory.
 
-Note that this usually will not include files in nested directories. For example say you have an output `Array[File] a_files = glob(a*)` and the end result of running your command has produced a directory structure like this:
+Note that this usually will not include files in nested directories. For example say you have an output `Array[File] a_files = glob("a*")` and the end result of running your command has produced a directory structure like this:
 ```
 execution_directory
 ├── a1.txt
--- a/versions/draft-3/SPEC.md
+++ b/versions/draft-3/SPEC.md
@@ -899,7 +899,9 @@ output {
 }
 ```
 
-The array of `File`s returned matches the bash evaluation of the glob string relative to the task's execution directory and in the same order. It's evaluated in the context of the bash shell installed in the docker image running the task. In other words, it contains the same files in the same order as would be matched by running `ls <glob>` in bash from the task's execution directory.
+The array of `File`s returned is the set of files found by the bash expansion of the glob string relative to the task's execution directory and in the same order. It's evaluated in the context of the bash shell installed in the docker image running the task. 
+
+In other words, it contains all of the files (but not the directories) in the same order as would be matched by running `ls <glob>` in bash from the task's execution directory.
 
 ### String Interpolation
 
--- a/versions/draft-3/SPEC.md
+++ b/versions/draft-3/SPEC.md
@@ -904,7 +904,7 @@ output {
 
 The array of `File`s returned is the set of files found by the bash expansion of the glob string relative to the task's execution directory and in the same order. It's evaluated in the context of the bash shell installed in the docker image running the task. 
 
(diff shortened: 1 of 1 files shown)


versions/draft-3/SPEC.md
```diff
@@ -891,14 +891,16 @@ output {
...
-Globs can be used to define outputs which contain many files.  The glob function generates an array of File outputs:
+Globs can be used to define outputs which might contain zero, one, or many files. The glob function therefore returns an array of File outputs:
 
 ```
 output {
   Array[File] output_bams = glob("*.bam")
 }
 ```
 
+The array of `File`s returned matches the bash evaluation of the glob from the execution directory the context of the bash version installed in the docker image running the task. In other words, it contains the files which would be matched by running `ls <glob>` from the task's execution directory.
```
Horneth: Is something off in the first sentence or is it just me ?

cjllanwarne: Hmm, you're right, it is a little bit strange on second reading. I'll try to break it into a couple of sentences instead

orodeh:
One of my users asked to sort the glob results. Can we do that?

cjllanwarne:
@orodeh I've added the clarification that the files should be returned in the same order that bash would return them if we call `ls <glob>`. This matches the Cromwell implementation.

rasmuse:
Nice that you're cleaning up, @cjllanwarne. This comment is just to check if I really understand you right and if you really mean what you wrote :)

## The output of `ls <glob>`

I'm not sure, but I suspect that the intended behavior is *not* to get the same files as returned by `ls <glob>`. Consider the following:

```bash
~$ mkdir tempdir && cd tempdir
~/tempdir$ touch file1
~/tempdir$ mkdir a_directory another_directory
~/tempdir$ touch a_directory/file2 another_directory/file3
~/tempdir$ ls *
file1

a_directory:
file2

another_directory:
file3
```

In other words, `ls <glob>` runs `ls` on all the files and dirs matched by `<glob>`. And I don't even think it's 100% clear what list of files and dirs should be considered the response here. `file1`, `file2`, `file3`? Or `file1` and the two directories? Or all five items?

### Maybe you mean `echo <glob>`?

Closer to what I would expect is `echo <glob>`:

```bash
~/tempdir$ echo *
a_directory another_directory file1
```

### Including directories?

The only potential problem is that only `file` is a file in the everyday sense of the word, but the other two matches are files in a wider sense which includes directories. Which brings me to a related question (which I can't find an answer to in the spec): does/should the WDL type `File` include directories? If then, I suppose `ls <glob>` can just be replaced by `echo <glob>` in @cjllanwarne's clarification?

## Link to bash `glob` docs?

By the way, I'm sure you know, but for reference, here's a link to the relevant bash docs on globs: https://www.gnu.org/software/bash/manual/bashref.html#Filename-Expansion -- would that be relevant to link from the spec or some other documentation?

cjllanwarne:
@rasmuse I think I addressed your comments and we discussed it in that Gitter conversation. Could you just add a thumbs-up here that it's looking good now?

geoffjentry:
Actually good you avoid using thumbs for this purpose? It makes vote tallying a bit murkier

rasmuse:
Yes, thanks, it's good this way. I'm a little confused about voting though - this issue is not marked "Voting Active" yet, but I'm ready to vote yes on it.

geoffjentry:
@rasmuse Right, please don't vote until we announce it ready for voting (which will likely be soon on this one)

versions/draft-3/SPEC.md
```diff
@@ -890,15 +892,37 @@ output {
...
+execution_directory
+├── a1.txt
+├── ab.txt
+├── a_dir
+│   ├── a_inner.txt
+├── az.txt
+```
+Then running `echo a*` in the execution directory would expand to `a1.txt`, `ab.txt`, `a_dir` and `az.txt` in that order. Since `glob()` does not include directories we discard `a_dir` and the result of the WDL glob would be `["a1.txt", "ab.txt", "az.txt"]`.
+
+##### Task portability and non-standard BaSH
```
geetduggal: Minor: is the mixed capitalization here for 'bash'  on purpose?

geetduggal:
:+1:

orodeh:
👍

mlin:
:+1:

aprabhak2:
👍

geoffjentry:
:+1:

Passes 5-0

PR #184, opened 2018-01-22
Let command <<<  contain literal bash scripts (as originally intended)

cjllanwarne:
EDIT: 
----
The intention of this PR has changed quite a bit.

Rather than amending the command blocks to allow escape sequences, I instead propose that we fix up the `command <<< >>>` blocks to allow them to contain literal bash without the need for WDL specific escape sequences. This moves `command <<< >>>` blocks closer to their original intention of being the "don't let WDL syntax get in the way of bash syntax" blocks.

Now this PR:

* Replaces `${}` in `command <<< >>>` blocks with `~{}` since this is not a valid (at least not a standard) bash construct - unlike the `${}` which it's replacing, which definitely is!

I'll leave my original comment below but note that it no longer reflects the intention of this PR.

Original:
----

~~From comments at the BOSC codefest.~~

~~It turns out maybe one of the most annoying language features for new authors might be one of the simplest to fix.~~

~~This is technically a breaking change (do we have a github label or some other way for marking that?) but I think it's one that's going to get a lot of sympathy from authors...~~

~~Introduces an escape character in the `command` section to avoid having to write this:~~
```wdl
task foo {
  String dollar="$"
  command <<<
    EXPORT hi=hello
    echo ${dollar}{hi}
  >>>
}
```
~~and instead be able to write this:~~
```wdl
task foo {
  command <<<
    EXPORT hi=hello
    echo \${hi}
  >>>
}~~

### CODE REVIEW

**Comment by cjllanwarne** [comment]
Since this will force users to change all existing `\` characters in a WDL script to `\\`, I wonder whether we'll be better off making this more focussed (eg only allow escaping `\$` and ignoring other `\`s?). Maybe that would be even more confusing? Any thoughts?

**Comment by orodeh** [comment]
My thought was to add an additional command block bracket syntax, in addition to `{,}` and `<<<, >>>`. For example:

```
command <<||
||>>
```

Inside a `<<||` block, the `${,}` notation is replaced by a symbol pair not used in bash. Something unlikely to cause confusion. For example: `~{, }`. This is used for WDL variables. `task foo` above, could be written more naturally:

```
task foo {
  command <<||
       export hi=hello
       echo ${hi}
  ||>>
}
```

The price is that WDL variables suffer from weird syntax inside commands. 
```
task wc {
    File data

    command <<||
         wc -l ~{data} | awk '{print $1}'
    ||>>
    output {
         Int num_lines = read_int(stdout())
    }
}
```

**Comment by cjllanwarne** [comment]
@orodeh a new block style _would_ work... 

In my opinion two `command` styles is probably already more than we really want. I find enough surprise and confusion amongst users about the existence of the `<<<`, `>>>` mode (and why it's needed) that I'm fearful about the user impact of yet another mode with yet another specific set of syntax.

I believe escape characters are widely enough known that it should be obvious to a someone who's never seen WDL before what `\${echo hi\}` means in a command block.

**Comment by jonn-smith** [comment]
@cjllanwarne I agree - sticking to simple / well-known conventions will facilitate users' adoption and happiness.

I was actually unaware that there is a current workaround for this (the ${dollar} example).

My preference in general is to have only one syntax for things if it can be helped - it keeps things simple and more intuitive.

I have no issue with the proposed escape scheme even though it's a breaking change.  You could always provide a script that would "upgrade" WDL files to the escaped version to ease migration pains.

**Comment by orodeh** [comment]
@cjllanwarne The issue with the `\$` sequence is that it is also can also appear in regular bash scripts. If a user wants to write a bash script that includes `\$`, it would (wrongly) be interpreted as WDL variable substitution. Perhaps you could use something instead of `\$`, for example `~$`. Yes, it is a lame, but it least it isn't confusing.

**Comment by jonn-smith** [comment]
@orodeh In that case wouldn't the user specify `\\$` (assuming \$ is the full escape sequence)?  

It would end up being very similar to escaping characters in a string.

**Comment by geoffjentry** [comment]
I would really like to avoid more block styles, that seems like it'd lead to more confusion, not less.

That said, to the followup point that @orodeh made, I heard a couple of folks at the codefest trying to find a viable character sequence and they kept running into conflicts.

**Comment by cjllanwarne** [comment]
@orodeh @jonn-smith @geoffjentry I wanted to try to nudge this PR towards some sort or conclusion. I believe we have three options:

1. Add a new block syntax which allows `${` to be interpreted literally and which has some other format for command sections, such as `~{}`
2. Add some form of escape character, acknowledging that it is a breaking change to existing command blocks.
3. Do nothing and leave users with the current workaround of declaring `String dollar = "$"` in order to get a dollar literal in their command.

To sum up the discussion so far:

* The general consensus above (and it matches my opinion) seems to be that a third block style is not the way to go.
* An escape sequence to me feels correct and general purpose. 
  * We'd need to provide an upgrade path which could automate the process of going through existing scripts and replacing any existing `\$`s with `\\$`s. In my opinion an upgrade script to port draft 2 to 3 will be an absolute necessity in any case. It's already part of the work I'm planning on doing for the path to draft 3 and 1.0.
* Doing nothing is always an option if this PR gets voted down 😄.

Since the conversation was sort of left hanging before I wanted to check whether there was any more to consider before I recommend opening this up to voting as-is?

**Comment by orodeh** [comment]
In my opinion, adding a third block style is the best option. It is backwards compatible, and also allows dropping a shell script "as is" into a command block.

**Comment by jonn-smith** [comment]
My vote is an escape sequence - it's consistent with common programming conventions so it will be somewhat intuitive.

**Comment by LeeTL1220** [comment]
My vote is escape sequence.  Same as @jonn-smith

**Comment by tlangs** [comment]
I think escape-sequence is the way to go.

**Comment by pshapiro4broad** [comment]
I think of the options on the table, adding a way to quote `$` is the best approach.

If there was a proposal for a new command block syntax that could eventually replace the existing ones -- e.g. it has all the features of the blocks we already have _and_ addresses this issue as well -- that would be a better solution. But adding a third syntax with yet another set of rules doesn't seem like a clean way to fix this.

**Comment by mckinsel** [comment]
Just a couple thoughts:

- If there were a command block style where WDL syntax and bash syntax didn't collide, that's the style I would use all the time. So if `<<|| ~{wdl_var} ||>>` (or whatever) were adopted, I would just switch every new WDL I write to that.
- WDLs don't have versions written into them, so if this PR is adopted I think the meaning of `\$` will be forever ambiguous. If someone sends you a WDL, how will you know if you're supposed to run your 2to3 tool on it?

**Comment by geoffjentry** [comment]
@mckinsel Regardless of styling, WDL versioning is likely coming soon: https://github.com/openwdl/wdl/pull/185

**Comment by cjllanwarne** [comment]
@pshapiro4broad @mckinsel @orodeh thanks for the input, my mind _is_ beginning to shift... Would I be correct in summarizing your positions as:

* You really want a new interpolation syntax which does not collide with common bash-isms, maybe `~{}~`, maybe something else. 
* The primary benefits of this are the ability to import directly from bash into WDL and the ability to use `${}` in the bash itself without resorting to side-channel hacks like `String dollar = "$"`.

In my opinion a third command block syntax is awkward because it makes a new author's life difficult trying to remember which is which and why they should use each one. I actually wonder when `command <<< >>>` blocks would ever be used for in this world - it seems like we'd be making them totally redundant (I think the `command { }` style will always have a place as the most concise and the most WDL-like syntax)

So: I wonder whether what we _really_ want to do is update the `command <<< >>>` blocks with an alternative interpolation symbol (maybe `~{}~`, maybe `<<<{}>>>`, maybe something else?)

Since these `command <<< >>>` blocks were originally intended to solve exactly this sort of problem, maybe it's time to fix **them** up so that they really can solve it for us?

**Comment by orodeh** [comment]
Fixing the `command <<< >>>` blocks would be much better!

**Comment by mckinsel** [comment]
@geoffjentry thanks for the pointer, that's great news.

@cjllanwarne, I like that idea a lot.

**Comment by geetduggal** [comment]
Thanks for the discussion. I too +1  @cjllanwarne 's last comment.  Arguments for using escape character appear to be (1) avoid introducing yet another way to refer to WDL vars and (2) escaping is well subscribed to or intuitive.  Arguments for alternative syntax appear to be (1) backwards compatible and (2) does not conflict with bash syntax.  I tend to prefer having alternative interpolation symbol so as to not conflict with existing bash scripts.  I have also run into issues in the past where escape characters introduce some difficult scenarios (e.g. escape characters exist in regexs and escape characters in bash -- what if you want to provide a regex as a command line argument in bash?  weird escaping issues can occur in this case).  Not that this is a huge concern here, but my intuition is to avoid escape characters when possible as a result.

**Comment by orodeh** [comment]
👍

**Comment by cjllanwarne** [comment]
@orodeh @jonn-smith @LeeTL1220 @tlangs @pshapiro4broad @mckinsel @geetduggal 

I've updated the title to indicate that this PR is now a totally changed suggestion to fixup the `command <<< >>>` style instead of introducing an escape character. I'd welcome any new feedback!

File: versions/draft-3/SPEC.md [hunk 1]
```diff
@@ -754,9 +743,19 @@ task test {
...
+task test {
+  String flags
+  command <<<
+    ps ~{flags}
+  >>>
+}
+```
 
-> **NOTE**: the `$expression` in this context can only evaluate to a primitive type (e.g. not `Array`, `Map`, or `Object`).  The only exception to this rule is when `sep` is specified as one of the `$var_option` fields
+> **NOTE**: the expression in this context must be interpolated as a string in the command. This is immediately possible for WDL primitive types (e.g. not `Array`, `Map`, or `Object`). To place an array into the command block a separater character must be specified using `sep` (eg `${sep=", " int_array}`).
```
pshapiro4broad: I Am Not a WDL Language Lawyer but I wonder if there's a benefit to distinguish 'string interpolation' with 'string replacement' in the text. It sounds like the same thing to me, and 'replace' seems like a simpler term.

pshapiro4broad:
This seems 👍  to me. Another thing to consider is supporting `~{...}` everywhere (e.g. in output sections and in `{` `}` command blocks). That way a WDL writer could use `~{}` all the time and not worry about which context they happen to be working in.

cjllanwarne:
@pshapiro4broad just to check: are you suggesting switching `${}` into `~{}` everywhere, or allowing both at the same time in `command {}` blocks?

pshapiro4broad:
> just to check: are you suggesting switching ${} into ~{} everywhere, or allowing both at the same time in command {} blocks?

I'm suggesting to allow both at the same time in `{` `}` blocks and anywhere else `${` `}` is currently supported. My thinking is that if the writer needs to know a new form for WDL string substitution, at least they don't have to remember when this new form is required.

versions/draft-3/SPEC.md
```diff
@@ -724,26 +724,22 @@ $task_sections = ($command | $runtime | $task_output | $parameter_meta | $meta)+
...
+Expression placeholders are denoted by `${...}` or `~{...}` depending on whether they appear in a `command { }` or `command <<< >>>` body styles:
 
-```
-$command_part = $command_part_string | $command_part_var
-$command_part_string = ^'${'+
-$command_part_var = '${' $var_option* $expression '}'
-```
+|Command Body Style|Placeholder Style|
+|---|---|
+|`command { ... }`|`${}` or `~{}`|
```
geoffjentry: Why not just have `~{}`?

cjllanwarne: My thought was that the `command {...}` syntax is already not good enough for bash syntax so we might as well allow users the `${...}` style placeholders. 

Eg even if `${...}` is not a WDL placeholder we still can't use it in a `command {...}` section because the final `}` is going to incorrectly close the command section.

cjllanwarne:
@geoffjentry AFAIK `${blah}` is a bash style construct but the PR would equally apply if that's valid in `sh` too

geoffjentry:
Technically too late to get this in past the deadline but YOLO.

moved this PR to `Voting Active`. Remember that you're voting on the PR **as-is**

orodeh:
👍

versions/draft-3/SPEC.md
cjllanwarne: Leaving aside whether "you must have matching numbers of `{`s and `}`s in your command block" is a requirement we'd want to add to WDL...

The more fundamental problem with that idea is that this "counting" would have to happen in the lexer, and "keeping track of opening and closing braces" is something that most lexers can't do (you can't do it with a FSA for the same reason you can't do it with a regex).

As it turns out Hermes has an "infinite" set of states with its stacks so I guess it might be able to handle that, but it would mess up the intellij, sublime, github, etc integration which are all FSA based.

geoffjentry:
:+1: with the following signing statement :)

I really dislike now having both `${}` and `~{}`. In 1.1 I'd like to see the following:

- Clarification to allow `~{}` used anywhere as an expression placeholder
- Deprecation of `${}` as an expression placeholder

And in 2.0 (or whatever) I'd like to see `${}` removed.

geetduggal:
:+1: and agree with @geoffjentry for future direction

geoffjentry:
It's a little bit difficult to ascertain the exact vote since multiple people :+1: my vote but either way this clearly passes. Waiting for implementation

patmagee:
I know I am late to the party but 👍

geoffjentry:
@cjllanwarne can you give this a rebase?

PR #185, opened 2018-01-24
WDL versioning 101, or should I say v101?

geoffjentry:


versions/draft-3/SPEC.md
```diff
@@ -636,6 +636,16 @@ $document = ($import | $task | $workflow)+
 
 `$document` is the root of the parse tree and it consists of one or more import statement, task, or workflow definition
 
+## Versioning
+
+For portability purposes it is critical that WDL documents be versioned so an engine knows how to process it. A WDL document is versioned with a `version` statement, for eample
```
yfarjoun: where must this line appear?

yfarjoun: also `eample`->`example`

yfarjoun: each spec should explicitly state its own version (not as an example). So the sentance should read, wdl-files adhering to the current version of this spec must have `version draft-3` as the first line in the file.

yfarjoun: I'd make it a comment and include the term "WDL" in it...so the version line I'd rally behind would be
```
#WDL version draft-3
```
That way if you happen to write draft-2 compliant code you are able to run it with an old engine.

versions/draft-3/SPEC.md
```diff
@@ -636,6 +636,16 @@ $document = ($import | $task | $workflow)+
...
+
+```wdl
+version draft-3
+```
+
+All WDL documents must have exactly one `version` statement. Any WDL document which does not have a `version` field may be treated as `draft-2`, however an implementation may choose to also support non-backwards incompatible changes which have made subsequent to `draft-2`.
```
yfarjoun: if the version statement isn't "commented" it would be impossible to version draft-2 WDL files.

versions/draft-3/SPEC.md
geoffjentry: I disagree w/ it being a comment, I think it should be a first class construct in the WDL syntax. That way it can be easily parsed without having to special case that specific comment.

re the spec vs example thing, my intention was that e.g. the next version would state that allowable values are `draft-3` and `whateverIsNext`. Perhaps I could start that trend with an N of 1 here?

geoffjentry: As to "where" I think technically the answer will be "anywhere" (again, going back to making the parsing reasonable), but that best practices would be to put it at the top.

versions/draft-3/SPEC.md
geoffjentry: Why would you need to version a `draft-2` document? If there's no version it **is** draft-2.

geoffjentry: Well, modulo some small set of WDLs in the wild which have been created since `draft-3` was created. I'd put forth that the people writing such WDLs are active enough that they'll manage to be able to add `draft-3` to their stuff.

versions/draft-3/SPEC.md
yfarjoun: I disagree with your disagreement :-) 

it needs to be special cased anyway, since you don't know what parser you need to use...the only way to protect against future changes to the spec and still be able to read any file and tell if it's a spec you support is to know EXACTLY where to find the version. thus: first line in the file. we are already talking about changing the block delimiter...how will your engine find the version if in draft 5 a wdl is unrecognizable to a draft-3 parser?

yfarjoun: "whatever is next" requires that we move off of the english language in "draft-3" and moe to something computable...how about 0.3 ? then you have a major.minor.point versions and you can be very clear about which version is greater than which

versions/draft-3/SPEC.md
yfarjoun: It would be nice to be able to be explicit....I don't need to.

versions/draft-3/SPEC.md
geoffjentry: The intention is that "whateverIsNext" is called "1.0.0" ;)

The sole reason for the existence of `draft-3` was to provide the ability to right some wrongs from `draft-2` prior to `1.0.0` while also allowing for the ability for backwards incompatible changes if necessary.

We (advisory board) did/do not want this period to go on forever, we've just been bad about driving the changes we felt necessary in a timely manner. This was one of them.

geoffjentry: **That** is out of the scope of this PR. If you want to change the versioning scheme, feel free to make a separate proposal.

yfarjoun: You still haven't responded to my concern about parsability...

consider that I might have the phrase `version blah` in many places in the wdl, without meaning to set the version (for example, we have a line that is 
```
version = version
```
in one of the wdl files right now in dsde-pipelines. Without actually parsing the file, you cannot tell if that is part of a command, part of a task, part of a workflow or part of a comment (maybe in the future we want to allow comment blocks?) and without knowing the version you cannot parse the file...so my sense is that by saying "the version line can be anywhere") you will actually limit the spec rather than enable breaking changes which is what the version is support to allow.

geoffjentry: It's because once in a while you make good points and I need to think about them for a bit :P

yfarjoun: Sure, I just didn't want it to drown in the other less important comments.

geoffjentry: Nope, I was going to take another pass through here - hopefully later today. Also part of me was hoping that other people would start chiming in to give other food for thought and/or let me read the direction the wind is blowing instead of us two just going back and forth. :)

cjllanwarne:
My initial thoughts reading the comment threads (thanks @yfarjoun and @geoffjentry for doing the hard thinking):

+1 to the version line being the first line or at least "before any other active elements"
+1 to version format being numeric, perhaps `[0-9]+\.[0-9]+(\.[0-9])*`
+1 to it being a real element not a comment 
+1 to "if a version tag doesn't appear before the first active element, it's draft-2"

cjllanwarne:
My gut feeling is also to change `"may be treated as draft-2"` to something more like "must be treated in a way that is backwards-compatible with draft-2"

geoffjentry:
@cjllanwarne re WDL versioning scheme being numeric, I'm still firmly of the belief that that's a separate PR. There's a fundamental difference between "This is how WDL is versioned" and "This is how one documents the version of a WDL".

Also on the backwards compat issue on versionless files I was trying to simultaneously allow two things:

1. The ability for an implementation to completely ignore the existence of `draft-2`. I think any verbiage beyond `must be treated as draft-2` (and thus if an impl doesn't support `draft-2` it goes to `/dev/null`) must be a `may`. 
2. The reason for the `may` is to allow Cromwell to be supportive if we choose to be (and anyone else who might want to do so)

geoffjentry:
@yfarjoun Going back to some of the stuff I left hanging last night I agree in some spots and not in others.

I agree that it should be the first line of the file. I'm pretty sure that'd be easy to express in Hermes, and it's definitely easy to express in BNF or CNF

I don't think it should be a comment, but a first class object in the namespace. . I don't see future versions really being an issue as a) it's so simple that it seems unlikely to change, b) any change would require a lot of handwringing about backwards compatibility, and c) one could always say "forms that look like XYZ are assumed to be ABC". Meanwhile making it a comment means an engine needs to introspect comments - in a world where most (currently all) implementations are grabbing the parsers built off the Hermes grammars, it'd be nice to have it part of the parse tree. 

Lastly, it looks cleaner, IMO.

For `draft-2` (or if one was **really** using `draft-1`, but it only ever existed as a theoretical construct) one could tack on a comment at the top if they really cared. I think this is a fairly academic issue as the reality is that it's an issue which will go away soon. Old WDLs that are actually used will find themselves upgraded sooner than later as implementations stop supporting old draft versions (which will inevitably happen)

cjllanwarne:
My gut feeling is to prefer "no version == draft 2 == no new features" as the carrot for encouraging upgrade and to keep portability even of draft-2 files.

geoffjentry:
@cjllanwarne Yeah, I think you're right. The reasons I cared aren't as useful as I thought. The space of WDLs in the world which exist in `draft-3` but predate this incoming change (however it shakes out) is quite small. And for Cromwell purposes we'll need to support `draft-2` anyways, so meh.

I think the only other use case I was thinking of was someone who just wanted to noodle around, we (or some other implementer) could allow that by going with the loose scheme. 

Why don't I change the verbiage to be purely strict. If Cromwell, or any other implementation, wants the room to be looser we can address that when that time arrives.

versions/draft-3/SPEC.md
```diff
@@ -636,6 +637,16 @@ $document = ($import | $task | $workflow)+
...
 
+## Versioning
+
+For portability purposes it is critical that WDL documents be versioned so an engine knows how to process it. A WDL document is versioned with a `version` statement, for eample
+
+```wdl
+version draft-3
+```
+
+All WDL documents must have exactly one `version` statement. Any WDL document which does not have a `version` field must be treated as `draft-2`.
```
cjllanwarne: You know I hate being annoyingly pedantically, but... 

The second sentence here now falsifies the first. It's probably clearer to say `"From draft-3 onwards, all WDL documents must ..."`

rasmuse:
Having read through the arguments here, I'd be in favor of
* if there is a `version` statement, it MUST occur on the first line and only there
* if there is no `version` statement, it MUST be treated as `draft-2`

When it comes to naming of versions, I like [semantic versioning](https://semver.org/) (2.0.0), but I'm not religious about it. In any case I think it would be good to agree on any semantic content of the version names *before* deciding on the syntax. I think that `0.3.0` would be a good and tidy name instead of `draft-3`, etc.

A couple of questions if the project aims to use semantic versioning:
* Does it matter at this point how engines may/should/must handle version numbers, e.g., is an engine allowed to execute workflows with a higher PATCH number than it explicitly supports?
* What would be an example of a backwards-compatible bug fix in the spec (i.e., a PATCH increment)? Will there ever be such changes?

geoffjentry:
Hi @rasmuse - I’d like to steer the conversation on this PR completely away from the question of how WDL itself is versioned. This is about how a WDL document notes the version it represents.

Every intention is that the next version is 1.0(.0?) and some form of semver in place but that’s a separate question. The ‘draft-‘ nomenclature is a historical artifact that predates even the team who originally built WDL. We chose to stick with that pattern for one more pre-1.0 version to allow for the possibility of backwards incompatible changes while things were still in flight

cjllanwarne:
@geoffjentry since it's come up a few times in this PR, I don't _think_ people are saying "we must rename draft-3 throughout the repository", only that "to represent draft-3 in a document we could use the literal `0.3.0` to make parsing easier in anticipation of future numerical versions"

cjllanwarne:
* although having said that, I'm pretty unconcerned what format it has in the document, I think it should be fairly easy to express `"X.Y.Z" OR "draft-3"`

cjllanwarne:
We can address more general inter-version compatibility issue outside of this PR, but do you think it's worth saying that `"A draft-3 versioned WDL cannot import a draft-2 WDL, nor vice versa"`?

geoffjentry:
For those who've been chiming in, I think I've covered the group consensus ...

- `version` as a first class WDL construct
- `version` is the first line of a file
- Any WDL file without a version is `draft-`
- All WDL files in a workflow must be the same version

Let me know if there are any other comments, if not I'll open to a PR

orodeh:
👍

cjllanwarne:
👍

rasmuse:
👍

geoffjentry:
Passes 4-0, requires implementation

PR #193, opened 2018-02-10
 Specify important runtime attributes and hints

cjllanwarne:
I'm sure there's plenty of room for refinement here.

Tries to do two things:
- Make an explicit list of `runtime` attributes so that we don't "leave it up to the engine" for attributes which materially affect the workflow outcome since doing so ruins portability.
- Introduces a `hints` section for everything else which would currently end up in `runtime` but which really can be left up to the engine.

### DIFF
--- a/versions/draft-3/SPEC.md
+++ b/versions/draft-3/SPEC.md
@@ -40,6 +40,7 @@
     * [String Interpolation](#string-interpolation)
     * [Runtime Section](#runtime-section)
       * [docker](#docker)
+    * [Hints Section](#hints-section)
       * [memory](#memory)
     * [Parameter Metadata Section](#parameter-metadata-section)
     * [Metadata Section](#metadata-section)
@@ -938,44 +939,44 @@ Any `${identifier}` inside of a string literal must be replaced with the value o
 
 ### Runtime Section
 
-```
-$runtime = 'runtime' $ws* '{' ($ws* $runtime_kv $ws*)* '}'
-$runtime_kv = $identifier $ws* '=' $ws* $expression
-```
-
-The runtime section defines key/value pairs for runtime information needed for this task.  Individual backends will define which keys they will inspect so a key/value pair may or may not actually be honored depending on how the task is run.
-
-Values can be any expression and it is up to the engine to reject keys and/or values that do not make sense in that context.  For example, consider the following WDL:
-
+The runtime section allows authors to supply values for a known set of runtime information fields which customize the execution of the task in ways which can meaningfully affect the output. The `runtime` section sits alongside `inputs`, the `command` section and the `output` section in a task definition, for example:
 ```wdl
-task test {
+task t {
   command {
-    python script.py
+    [...]
   }
   runtime {
-    docker: ["ubuntu:latest", "broadinstitute/scala-baseimage"]
+    docker: "ubuntu:latest"
+  }
+  output {
+    [...]
   }
 }
 ```
 
-The value for the `docker` runtime attribute in this case is an array of values.  The parser should accept this.  Some engines might interpret it as an "either this image or that image" or could reject it outright.
-
-Since values are expressions, they can also reference variables in the task:
+Attribute values can be any expression so long as they match the expected type for the runtime attribute.
 
 ```wdl
 task test {
-  String ubuntu_version
-
+  String ubuntu = "ubuntu"
+  String latest = "latest"
   command {
     python script.py
   }
   runtime {
-    docker: "ubuntu:" + ubuntu_version
+    docker: ubuntu + ":" + latest
   }
 }
 ```
 
-Most key/value pairs are arbitrary.  However, the following keys have recommended conventions:
+The following set of keys are supported in the `runtime` section for WDL draft 3:
+
+|Key|Type|Default|Description|
+|---|---|---|---|
+|`continue_on_return_code`|`Boolean` or `Integer` or `Array[Integer]`   |0   |If the task exits with a return code specified here, it is considered successful. `true` means that the task always succeeds as long as its command runs to completion. `false` is equivalent to `0` |
+|`fail_on_stderr`|`Boolean`|`false`|If true, the task fails if any content is written to `stderr` by the command. |
+|`docker`|`String?`|`None`|If specified, the `docker` image which this task must be run on. If empty, the task is run without any portable container. To make a WDL fully portable, docker images should be specified using fixed tags or docker hashes.|
+
 
 #### docker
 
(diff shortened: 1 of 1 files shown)


versions/draft-3/SPEC.md
```diff
@@ -994,6 +995,40 @@ task docker_test {
...
+* The only requirement on new `hints` from the spec is that no engine may implement a `hint` which materially affects the outcome of the task. These should instead be handled as spec changes for additions to the `runtime` section.
+* It is intended that over time the number of fields that an author places in the `hints` section decreases as workflow-submission APIs mature to define the same set of options in a more appropriate abstraction layer.
+
+The following `hints` are reserved by the WDL spec and have a known format. An engine is free to ignore them, but must not override them with an alternative format and implementation. Because hints do not materially affect workflow outcomes, they have no fixed default values. An engine is free to implement whatever defaults it likes:
+
+|Key|Type|Description|
+|---|---|---|
+|`memory`|`String` or `Int`|See [memory](#memory) section below.|
+|`cpu`|`Int`|A recommended minimum number of CPUs for the task. |
+
```
orodeh: The `disk` property is missing here. Is this an intentional omission?

orodeh:
I think this is a good split, that improves portability. The only gray area that I can think of, is the `cpu`, `disk`, and `memory` properties. These could, perhaps, be placed in the runtime attributes section, because the engine *must* provide an instance that has at least these properties.

versions/draft-3/SPEC.md
```diff
@@ -994,6 +995,40 @@ task docker_test {
...
+  }
+}
+```
+
+
+Hints allow a WDL author to provide recommendations to the engine when running a task. 
+
+* Unlike `runtime` these fields are not specified to indicate *what* a workflow does, but rather to provide hints to the engine about *how* to run a workflow.
+* Also unlike `runtime`, the `hints` section allows arbitrary keys and values beyond the set defined below to allow platform specific collaboration between WDL author and platform.
+* The only requirement on new `hints` from the spec is that no engine may implement a `hint` which materially affects the outcome of the task. These should instead be handled as spec changes for additions to the `runtime` section.
```
patmagee: I wonder how difficult this will be to enforce, and what specifically would be considered a material change? IE if I create a file should the MD5 sums match regardless of what the hints section reads? The hints seems like a logical place to put Provider specific information, IE using GPU on Google, but these might actually change the output to some marginal degree since they could be running a modified pipeline.

cjllanwarne: These are some good points and they don't have easy answers! So these are just my gut reactions:

* Enforcement: To some extent it's always going to require either trust in engine implementers *or* some sort of external validation system that can assert that a given workflow works on multiple engines. Maybe Dockstore can eventually fill that role (or at least record whether a workflow is engine-specific as a sort of public shaming?)

* Material change: This is also unfortunately squishy in a world of stochastic analyses. We struggled with the same thing in Cromwell for call caching so my list of "attributes which should be considered important" here is a transcription of the list of "attributes which affect call caching".

cjllanwarne:
@orodeh I suppose things like CPU could probably exist happily in both boxes (some methods really might require 2 CPUs, others might just go faster).

From a Cromwell point of view one nice thing about separating the "want"s from the "need"s is that we could decide to use CPU to drive call caching in a "need" scenario but not a "want" scenario, avoiding the current "one size fits all".

versions/draft-3/SPEC.md
cjllanwarne: Nope - just an oversight (and more evidence that a single place to record all these will be useful!)

patmagee:
@cjllanwarne @orodeh just to play the devils advocate, I am always hesitant to add features which add new sections to tasks/workflows. One of the powers of WDL is that there are not many things you need to define in order to have a functional workflow.  I can definitely see the use case for this, but why use a separate section as opposed to defining constant "shared" keys that all implementations used, and then extending the runtime section to allow for non-standard attributes.

Implementations would simply require the minimal subset of attributes are set: ie `cpu`, `disks`, `docker`, `memory`  (if they are not set by default), and then any additional attributes would be interpreted by specific implementations. This is essentially what we have now I guess

cjllanwarne:
I guess what you're getting at is, is there more cognitive load in (1) "knowing which section to add" vs (2) "which elements bundled together are known vs arbitrary keys". Since I stare at WDLs all day I'm not going to be very good at answering that so that's great input! 😄 

Since we missed the 1.0 deadline, I guess that turning "consensus" attributes into "enforced" attributes is maybe a breaking change (i.e. do we need to wait for 2.0)?

geoffjentry:
Not necessarily a commentary on what's currently here, but to add to the point @patmagee is making, I am constantly and increasingly worried about the added complexities being added to WDL. A common counter is "but I/we need to be able to express X!". That's great, but find a way to do so which is simple


IMO WDL has already become overly complex, so finding ways to reduce that instead of increasing that would be good.

patmagee:
for me, the difference between `hints` and `runtime` is not large enough to mandate an entirely new section. Not to mention it's not necessarily something that is going to be well understood by users. In languages like `CWL`, this is one aspect that although it makes it more flexible in certain scenarios, adds to the overall complexity and difficulty.

similar to @geoffjentry I like to take the approach of asking "Is there currently something in the language that supports this, which simply needs more clarification, OR if there is not, can we minimally extend the existing components to get to where we want to go". If we follow that approach we will likely not run into an overly complex language. WDL will not (nor do I think it should strive to be) all things for all people., but we can definitely fulfill the majority of use cases.

cjllanwarne:
I totally agree with the sentiment, I just want to make sure we're aiming to minimize cognitive load rather than simply minimize the raw feature count as a proxy for cognitive load.

Regarding this specific PR, if the consensus is that in this case "one section with two jobs" is easier to work with for newer authors than "two distinct sections" then I'm happy to go that way.

patmagee:
Yeah, I am okay if we have a large number of features, however, if it becomes complicated for the user to set up a task, or to even look at a task, then I think we are missing the mark with the language

geoffjentry:
@cjllanwarne What do you want to do with this?

cjllanwarne:
I think the consensus from the comments (tell me if I'm wrong) is that separating `hints` out isn't the right way to go here - at least not without some more thought.

In which case:

*  I think the list of attributes which are backend independent and which materially affect job outcomes is worth keeping (because things like `continueOnReturnCode` aren't really up to an engine to decide whether to implement)
* The hints stuff can go for now (and the existence of them in `runtime` can remain a sore spot to be fixed later, maybe once we have more data on backends, the sorts of attributes we need to capture, and whether APIs like WES can handle this sort of "how to run" information outside of the "what to run" workflow descriptions)
* I can close this PR and resubmit against `development` with that in mind

WDYT?

cjllanwarne:
Closing for now. Might return against `development` but in any case would need to be a fundamentally different PR...

PR #202, opened 2018-03-13
Don't forget flatten

cjllanwarne:
This PR is just grandfathering in https://github.com/openwdl/wdl/pull/154 for draft 3.

I guess the change must have been made:
* Before the line was drawn finalizing draft 2
* After the draft3 spec had been copied

I don't think it was a deliberate omission from draft 3...?


geoffjentry:
Yeah, definitely an oversight.

geoffjentry:
There's stuff in #154 that aren't here. Is **that** intentional?

cjllanwarne:
@geoffjentry please double check but AFAICT the only differences are whitespace (presumably @orodeh has an editor that tidies up newlines for him 😄 )

orodeh:
Thanks!

PR #206, opened 2018-04-13
Fix typo

aednichols:
Exactly what it says on the tin.


geoffjentry:
Merging as-is as it's just a typo fix and we haven't cut 1.0 quite yet.

PR #207, opened 2018-04-13
Fix typo II

aednichols:


geoffjentry:
I suppose we could have run this through a spell check :)

PR #219, opened 2018-06-04
Adding functions for conversion between Array[Pair[X,Y]] and Map[X,Y].

### LINKED ISSUE #216: Coercing Array[Pair[X,Y]] to Map[X,Y]
It might be nice to a have a function/method in the standard library to convert an array of pairs into a map. Using the left value as key and the right value as... well... value. This might be useful when directing output from a scatter to a sub workflow.

For example:
Lets say you have a workflow which takes a `Map[String, File]` as input. The key would be a sample name and the value a BAM file. 
You might have another workflow which uses a scatter (scattering over an `Array[String]` containing sample names) to produce these BAM files (as an `Array[File]`). This workflow should then give them to the before mentioned workflow as input.
Currently I can't find a way to do this in wdl. You can, however, produce a `Array[Pair[String, File]]` using the `zip` function. If you could convert this resulting `Array[Pair[String,File]]` into a `Map[String, File]`, this would become doable in wdl.

DavyCats:
This would close #216. It adds two functions to the standard library:
* `as_map`: converts an Array of Pairs to a Map
* `as_pairs`: converts a Map to an Array of Pairs

Let me known what you think.

### DIFF
--- a/versions/draft-3/SPEC.md
+++ b/versions/draft-3/SPEC.md
@@ -3053,6 +3053,18 @@ Array[String] zs = [ "d", "e" ]
 Array[Pair[Int, String]] crossed = cross(xs, zs) # i.e. crossed = [ (1, "d"), (1, "e"), (2, "d"), (2, "e"), (3, "d"), (3, "e") ]
 ```
 
+## Map[X,Y] as_map(Array[Pair[X,Y]])
+
+Given any two Object types in a Pair, the `as_map` function returns a Map in which the given left element is the key and the right element the value.
+
+```
+Array[Pair[String,Int]] x = [("a", 1), ("b", 2), ("c", 3)]
+Array[Pair[String,Pair[File,File]] y = [("a", ("a.bam", "a.bai")), ("b", ("b.bam", "b.bai"))]
+
+Map[String,Int] xmap = as_map(x) # {"a": 1, "b": 2, "c": 3}
+Map[String,Pair[File,File] ymap = as_map(y) # {"a": ("a.bam, "a.bai"), "b": ("b.bam", "b.bai")}
+```
+
 ## Integer length(Array[X])
 
 Given an Array, the `length` function returns the number of elements in the Array as an Integer.
--- a/versions/draft-3/SPEC.md
+++ b/versions/draft-3/SPEC.md
@@ -119,6 +119,8 @@
   * [Array\[Array\[X\]\] transpose(Array\[Array\[X\]\])](#arrayarrayx-transposearrayarrayx)
   * [Array\[Pair(X,Y)\] zip(Array\[X\], Array\[Y\])](#arraypairxy-ziparrayx-arrayy)
   * [Array\[Pair(X,Y)\] cross(Array\[X\], Array\[Y\])](#arraypairxy-crossarrayx-arrayy)
+  * [Array\[Pair(X,Y)\] as_pairs(Map\[X,Y\])](#arraypairxy-as_pairsmapxy)
+  * [Map\[X,Y\] as_map(Array\[Pair(X,Y)\])](#mapxy-as_maparraypairxy)
   * [Integer length(Array\[X\])](#integer-lengtharrayx)
   * [Array\[X\] flatten(Array\[Array\[X\]\])](#arrayx-flattenarrayarrayx)
   * [Array\[String\] prefix(String, Array\[X\])](#arraystring-prefixstring-arrayx)
@@ -3053,9 +3055,21 @@ Array[String] zs = [ "d", "e" ]
 Array[Pair[Int, String]] crossed = cross(xs, zs) # i.e. crossed = [ (1, "d"), (1, "e"), (2, "d"), (2, "e"), (3, "d"), (3, "e") ]
 ```
 
+## Array[Pair[X,Y]] as_pairs(Map[X,Y])
+
+Given any two Object types in a Map, the `as_pairs` function returns an Array containing each element in the form of a Pair. The key will be the left element of the Pair and the value the right element.
+
+```
+Map[String, Int] x = {"a": 1, "b": 2, "c": 3}
+Map[String,Pair[File,File] y = {"a": ("a.bam, "a.bai"), "b": ("b.bam", "b.bai")}
+
+Array[Pair[String,Int]] xpairs = as_pairs(x) # [("a", 1), ("b", 2), ("c", 3)]
+Array[Pair[String,Pair[File,File]] ypairs = as_pairs(y) # [("a", ("a.bam", "a.bai")), ("b", ("b.bam", "b.bai"))]
+```
+
 ## Map[X,Y] as_map(Array[Pair[X,Y]])
 
-Given any two Object types in a Pair, the `as_map` function returns a Map in which the given left element is the key and the right element the value.
+Given any two Object types in an Array of Pairs, the `as_map` function returns a Map in which the given left element is the key and the right element the value.
 
 ```
 Array[Pair[String,Int]] x = [("a", 1), ("b", 2), ("c", 3)]


versions/draft-3/SPEC.md
```diff
@@ -3053,6 +3055,30 @@ Array[String] zs = [ "d", "e" ]
 Array[Pair[Int, String]] crossed = cross(xs, zs) # i.e. crossed = [ (1, "d"), (1, "e"), (2, "d"), (2, "e"), (3, "d"), (3, "e") ]
 ```
 
+## Array[Pair[X,Y]] as_pairs(Map[X,Y])
+
+Given any two Object types in a Map, the `as_pairs` function returns an Array containing each element in the form of a Pair. The key will be the left element of the Pair and the value the right element.
```
cjllanwarne: I don't like `Object` here since that is an existing, concrete WDL type (at least for now). I also don't think it's needed here - how about `"Given a Map[X, Y], return an Array[Pair[X, Y]] containing ..."`

versions/draft-3/SPEC.md
```diff
@@ -3053,6 +3055,30 @@ Array[String] zs = [ "d", "e" ]
...
+Map[String, Int] x = {"a": 1, "b": 2, "c": 3}
+Map[String,Pair[File,File] y = {"a": ("a.bam, "a.bai"), "b": ("b.bam", "b.bai")}
+
+Array[Pair[String,Int]] xpairs = as_pairs(x) # [("a", 1), ("b", 2), ("c", 3)]
+Array[Pair[String,Pair[File,File]] ypairs = as_pairs(y) # [("a", ("a.bam", "a.bai")), ("b", ("b.bam", "b.bai"))]
+```
+
+## Map[X,Y] as_map(Array[Pair[X,Y]])
+
+Given any two Object types in an Array of Pairs, the `as_map` function returns a Map in which the given left element is the key and the right element the value.
```
cjllanwarne: Same comment about `Object`

cjllanwarne: Could you add a comment on what happens if there's a key collision, eg `[(1, 2), (1, 3)]`?

cjllanwarne:
Do you anticipate that `as_pairs(as_map(pairs))` should return items in the same order? Right now maps are "unordered" but this might be a good time to specify that the order of a map *must* match input order?

cjllanwarne:
PS I forgot to say but thanks so much for submitting this PR! Notwithstanding a couple of comments, it looks great!

DavyCats:
> Do you anticipate that as_pairs(as_map(pairs)) should return items in the same order? [...]

I hadn't really considered the order. Personally I wouldn't assume the order to be the same, as Maps are generally unordered, but that might indeed not be assumed by everyone (especially people without a programming background). Does anyone have any objections against making Maps ordered? As far as I'm concerned it doesn't really matter.

 > Could you add a comment on what happens if there's a key collision [...]

So there are three option I can think of in how to deal with this:
1. Throw an exception on key collision.
2. Ignore elements which would cause a key collision, i.e. keep the value associated with first occurrence of a key.
3. Override values if a later element has the same key.

Option 1 seems the safest to me, otherwise I would go for option 3, but I would like to know what other think about this.

patmagee:
@DavyCats Thanks a lot for putting this together, overall it looks good to me. 

Im not too opinionated on whether we should preserve order of the pairs, although I think the implications of enforcing that probably should be considered. My feeling is that the overall expectation will be that `as_pair(as_map(pairs))` will present that same array of pairs as was originally ingested. This would require the underlying engine to use a `linked hash map` of some sort

As for the key collision thats a bit of a connundrum. It almost seems inevitable that there will be a key collission, and it feels like a bit of a waste to fail a workflow because of it. One potential option would be to change the return type of the `as_map` function. This becomes a bit more unintuitive thought for the user.

```wdl
Array[Pair[X,Y]] pairs
Map[X,Array[Y]] mapped = as_map(pairs)
```

DavyCats:
One other way to solve the key collision problem (at least on our end :stuck_out_tongue: ) would be to leave it up to the user how they deal with key collisions. This could be done by adding an additional 'mode' argument to the function which would take one of the following values:
- `"strict"`: Throw an exception.
- `"ignore"`: Ignore later elements which would have the same key.
- `"override"`: Override values if later element have the same key.
- `"collect"`(?): Turn the values into an arrays containing all the values for a certain key.

This feels a bit awkward though.

Another option would be to add another function, `collect_as_map`, which would create a `Map[X, Array[Y]]` and then use the 'strict' mode for the `as_map` function. In that case, if there is the possibility for key collisions you should use `collect_as_map` otherwise you can use `as_map`.

cjllanwarne:
I think I mostly agree regarding the collision. To summarize what I think my (current) gut preference would be:
* As proposed, the function `as_map`:
  * One param: the array of pairs
  * Uses the `"ignore"` mode for collisions
  * Produces a `Map[X, Y]`

* Another function `collect_by_key`:
  * One param: the array of pairs
  * Produces a `Map[X, Array[Y]]`

I *think* I would personally leave it there for WDL 1.1 (and tbh I'm even on the fence about `collect_by_key`) but if people end up clamoring for the `"strict"` mode, we could maybe add a second (optional) boolean param to `as_map`: `"fail on collision"`
  * The default would be false
  * If `true`, that's your `"strict"` mode
  * If `false`, that's your `"ignore"` mode

geoffjentry:
My $0.02:

- Things should preserve order, even in cases where a programmer would intuit otherwise (e.g. maps). I've seen too many cases in workflows where that sort of thing causes confusion for users.
- TBH I think for now I'd prefer to see a collision as an error and see how people react. I'd rather us not try to either guess ahead of time what most people are going to want nor provide people with a million configuration options. IMO the **right** answer is "do whatever is most natural to most people, and only that" and I don't think we're armed with enough data yet.

versions/draft-3/SPEC.md
```diff
@@ -3053,6 +3056,44 @@ Array[String] zs = [ "d", "e" ]
...
+
+Array[Pair[String,Int]] xpairs = as_pairs(x) # [("a", 1), ("b", 2), ("c", 3)]
+Array[Pair[String,Pair[File,File]] ypairs = as_pairs(y) # [("a", ("a.bam", "a.bai")), ("b", ("b.bam", "b.bai"))]
+```
+
+## Map[X,Y] as_map(Array[Pair[X,Y]])
+
+Given an Array consisting of Pairs, the `as_map` function returns a Map in which the left elements of the Pairs are the keys and the right elements the values. The order of the key/value pairs in the resulting Map is the same as the order of the Pairs in the Array.
+
+In cases where multiple Pairs would produce the same key, the workflow will fail.
```
cjllanwarne: I believe the keys to maps must be primitives whereas the left side of pairs can be arbitrarily complex. Maybe we should also add that as a note alongside this comment? 
```md
In cases where the left side of the pair is not a primitive type, the workflow will fail
```

cjllanwarne: In fact... since hopefully we'll be able to detect this statically 99% of the time it should probably just be:
```
"The left hand side of the pair passed to `as_map` must be a primitive type"
```

patmagee:
@geoffjentry erroring out on a collision seems like the best approach for now, or at least the safest use case.  We probably would need to see real uses for this before we can make an objective call on the different `modes`  to run this in

geoffjentry:
@DavyCats I wanted to circle back on this and see where you think it is in terms of bringing to a vote. Are you looking for more feedback and/or still wanting to act on existing feedback? Or are you looking to keep it as-is?

For other folks who have commented, are you satisfied w/ it where it's landed?

DavyCats:
@geoffjentry If nobody has any further suggestions or comments, then I think this can be brought to a vote.

patmagee:
this lgtm and I would be ready to vote on it!

orodeh:
👍

geetduggal:
👍

rhpvorderman:
:+1:

cjllanwarne:
👍

aprabhak2:
👍

geoffjentry:
Regardless of fixing typos, this is going to need to be re-targeted to [the developmental spec](https://github.com/openwdl/wdl/blob/master/versions/development/SPEC.md)

DavyCats:
Fixed :sweat_smile:

cjllanwarne:
An implementation [now exists](https://github.com/broadinstitute/cromwell/pull/3895) in Cromwell's `develop` branch.

If you want to try it out you'll need to build from `develop` because it's not part of a release version yet. Also I wasn't 100% sure whether the next version will be 1.1 or 2.0 or something utterly different, I'm using `version biscayne` for now. That'll have to be updated once the version number is officially decided.

Issue #214, opened 2018-05-21
One idea to make imports more intuitive


cjllanwarne:
One thought I'd had on making imports a bit more intuitive is to allow an optional `namespace` field at the top of a WDL file like this:

```wdl
version 1.1

namespace "bam_processing/workflows"

import "bam_processing/tasks/my_tool.wdl"

workflow xyz {
  call my_tool.tool_task { input: ... }
}
```

How does this help anyone? Well.... when it comes to importing one of the tricky things is to work out where import paths are relative to. Sometimes they should be "relative to where I'm running from", sometimes they should be "relative to where this WDL file is". At the moment this is left up to the engine and often the zip bundle submitted alongside the main workflow.

This suggestion is that if a WDL specifies a `namespace`, that allows the engine to resolve imports relative to "the directory you get if you go back to the base of the namespace". In other words if the WDL above is in a directory `/path/to/tasks/bam_processing` then the "base" is `/path/to/` and the import can be resolved as `path/to/bam_processing/tasks/my_tool.wdl`

---

Unresolved question: import versioning

rhpvorderman:
Great! This feature would be very nice.
I think it is easier that imports are relative to the file and not `$PWD` but it would be better if this was explicit in the WDL file itself.

This would solve numerous issues with the versioning of imports, as you can put different versions in different directories.

aednichols:
@GregTD42 this may be of interest related to your recent Cromwell issue

microbioticajon:
I would like to add my voice to this feature request too. I just ran into the the issue outlined in https://github.com/broadinstitute/cromwell/issues/6441.

Within a "package" relative imports ./ and ../ work nicely but when a wdl is imported within an import it is not intuitive for the user to work out what the relative path of the bundled dependency will be.

PR #220, opened 2018-06-05
Allow for imports relative to current location

geoffjentry:
Feedback I've received from out in the wild was that this was a massive, massive pain point. 

Declare as part of the spec that imports which specify a relative path are relative to the location of the WDL document issuing the import statement.

### DIFF
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -760,7 +760,9 @@ Engines should at the very least support the following protocols for import URIs
 
 * `http://` and `https://`
 * `file://`
-* no protocol (which should be interpreted as `file://`
+* No protocol (see below)
+
+In the event that there is no protocol, if this is an absolute path (i.e. starts with `/`) it is interpreted as a `file://`. Otherwise it is **relative** to the location of the current document. For instance, if the current doc is at `file://foo/bar/baz/qux.wdl` and specificies `import some/task.wdl` this will be resolved to `file://foo/bar/baz/some/task.wdl`. Likewise if the current doc is at `http://www.github.com/openwdl/coolwdls/myWorkflow.wdl` and imports `subworkflow.wdl`, it will be resolved to `http://www.github.com/openwdl/coolwdls/subworkflow.wdl`. It is up to the implementation to provide a mechanism which will allow these imports to be resolved correctly.
 
 
 ## Task Definition
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -732,7 +732,7 @@ A WDL file may contain import statements to include WDL code from other sources
 $import = 'import' $ws+ $string ($ws+ 'as' $ws+ $identifier)?
 ```
 
-The import statement specifies that `$string` which is to be interpted as a URI which points to a WDL file.  The engine is responsible for resolving the URI and downloading the contents.  The contents of the document in each URI must be WDL source code.
+The import statement specifies that `$string` which is to be interpreted as a URI which points to a WDL file.  The engine is responsible for resolving the URI and downloading the contents.  The contents of the document in each URI must be WDL source code.
 
 Every imported WDL file requires a namespace which can be specified using an identifier (via the `as $identifier` syntax). If you do not explicitly specify a namespace identifier then the default namespace is the filename of the imported WDL, minus the .wdl extension.
 For all imported WDL files, the tasks and workflows imported from that file will only be accessible through that assigned [namespace](#namespaces).
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -762,7 +762,7 @@ Engines should at the very least support the following protocols for import URIs
 * `file://`
 * No protocol (see below)
 
-In the event that there is no protocol, if this is an absolute path (i.e. starts with `/`) it is interpreted as a `file://`. Otherwise it is **relative** to the location of the current document. For instance, if the current doc is at `file://foo/bar/baz/qux.wdl` and specificies `import some/task.wdl` this will be resolved to `file://foo/bar/baz/some/task.wdl`. Likewise if the current doc is at `http://www.github.com/openwdl/coolwdls/myWorkflow.wdl` and imports `subworkflow.wdl`, it will be resolved to `http://www.github.com/openwdl/coolwdls/subworkflow.wdl`. It is up to the implementation to provide a mechanism which will allow these imports to be resolved correctly.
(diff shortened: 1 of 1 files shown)


yfarjoun:
If I understand correctly this could solve a huge problem that we currently have with imports!

versions/development/SPEC.md
```diff
@@ -760,7 +760,9 @@ Engines should at the very least support the following protocols for import URIs
 
 * `http://` and `https://`
 * `file://`
-* no protocol (which should be interpreted as `file://`
+* No protocol (see below)
```
yfarjoun: should probably bump the minor version

versions/development/SPEC.md
```diff
@@ -760,7 +760,9 @@ Engines should at the very least support the following protocols for import URIs
...
+
+In the event that there is no protocol, if this is an absolute path (i.e. starts with `/`) it is interpreted as a `file://`. Otherwise it is **relative** to the location of the current document. For instance, if the current doc is at `file://foo/bar/baz/qux.wdl` and specificies `import some/task.wdl` this will be resolved to `file://foo/bar/baz/some/task.wdl`. Likewise if the current doc is at `http://www.github.com/openwdl/coolwdls/myWorkflow.wdl` and imports `subworkflow.wdl`, it will be resolved to `http://www.github.com/openwdl/coolwdls/subworkflow.wdl`. It is up to the implementation to provide a mechanism which will allow these imports to be resolved correctly.
```
yfarjoun: what about `import ../../quux/quuz.wdl`? allowed?

cjllanwarne: IMO as long as that makes sense relative to the "current" URL, I don't see why not (although I would personally still like some scheme like https://github.com/openwdl/wdl/issues/214 to indicate how to get back to the "import root" without resorting to long trails of `../`)

versions/development/SPEC.md
cjllanwarne: Of the WDL spec? this is currently against `versions/development`

versions/development/SPEC.md
cjllanwarne: I've got to believe there's a more sensible `/` for HTTP?

eg if I were to say `import /chrisl/coolwdl/task.wdl` from `http://www.github.com/openwdl/coolwdls/myWorkflow.wdl` I might expect that to look for `http://www.github.com/chrisl/coolwdl/task.wdl` rather than on the local FS

versions/development/SPEC.md
geoffjentry: I had the same question @yfarjoun - regardless, yes this would be for WDL 1.1.

versions/development/SPEC.md
geoffjentry: @yfarjoun So @cjllanwarne and I have talked about this a bit, and a similar discussion was had in Toronto more generally among Cloud Workstream members (so applying to things beyond WDL as well). The sentiment from the latter (shared by myself) is that people **shouldn't** use `../` but they're **going** to so don't explicitly disallow it. IOW make it one of those best practices/linter type of things.

Why shouldn't they? It assumes POSIX-like file structures, and that might fall apart at some point in some arena. However, for practical purposes it will work, so ....

geoffjentry: I thought about that a bit. My objection is that in file system semantics `/xyz` implies an absolute path. Yes, I did just complain about people assuming filesystem semantics above, but this feels like a case where more people will be bit by it than helped. 

OTOH I do like the simplicity of what you describe.

patmagee: @geoffjentry I agree that we cannot explicitly restrict the use of `../` paths. as annoying as that is. People will undoubetly use it, but I like the approach you are taking. we will support but only grudgingly. We can clearly describe in best practices what way users SHOULD do this

patmagee: @cjllanwarne although what you describe is simple it seems like you could quickly run into a frustrating situation where people have a hodge podge of imports and nested calls that cannot be reasily imported. For example if a restapi exposes a wdl at:
```http://www.wdl-host.com/public/my-workflow.wdl```
and that wdl imports something from `/....`, it becomes a bit ambigious as to what the `/` relative path means. when the workflow was written and then exposed by a restapi it may not have known anything about where it was going to be  hosted. Or even if it was going to be hosted at all.

Personally I do not really like any paths that start with `/` but really everything should be relative to the current WDL (other then http/ftp`) calls. The moment we start allowing absolute paths everything becomes alot harder and is more subject to breaking because of permissions / moved or deleted file

patmagee:
Continuing my comment from above, are absolute paths things that people will really use in the wild? using absolute paths in the context of an execution engine running on a local backend makes sense, to some degree, especially if the user running things controls everything.

The moment the workflow moves away from being in a local setup, however, absolute paths bring a large number of problems, the first of course being absolute to what? absolute to the file system, absolute to the HTTP host, absolute within a zip file. This is all very contextual and seems to introduce a lot of places for failure, whereas always assuming relative paths to the current context would fix this.

For example, allowing absolute paths would allow users to define imports according to their absolute file structure which is not very resuable at all. Relative paths can still lead to this but its more flexible

LeeTL1220:
Hello @geoffjentry, @yfarjoun  and @cjllanwarne  If I may chime in...  

One quick point:
- In the GATK4 repo WDL, we have a real use for `../` for our common tasks.  Only by convention do we not use it now.

patmagee:
@LeeTL1220 is there a situation where you would use `/` as opposed to `../`?

LeeTL1220:
@patmagee Currently, no.

geoffjentry:
@patmagee My main concern with explicitly disallowing absolute paths is that it'd be a breaking change as they're explicitly allowed currently. Specifying an import w/o a schema has been specified on becoming `file://` and when the host is not included on a file URL the extra slash (ie absolute path) is not optional.

Much like the use of `..` I'd prefer to put this in the "please don't do this, but ...." bundle and then look to remove it when we're talking about `2.0` instead of `1.1` (yes, I'm invoking semver based laziness). I suspect in practice it's just not that big of a deal.

cjllanwarne:
For example: 
* `the [previous pull request](/openwdl/wdl/pull/220)`
* the [previous pull request](/openwdl/wdl/pull/220)

geoffjentry:
@cjllanwarne  "The previous page is sending you to an invalid url (http:///openwdl/wdl/pull/220)."

cjllanwarne:
@geoffjentry are you opening that from the github page or from an embedded link in an email (which has no URL context?)

geoffjentry:
Made a few edits based on a conversation with @cjllanwarne - if I don't get any pushback on what's here now I'll put it up for a vote tomorrow

orodeh:
👍

geetduggal:
👍

rhpvorderman:
:+1:

DavyCats:
:+1:

ffinfo:
👍

patmagee:
👍

cjllanwarne:
👍

aprabhak2:
👍

cjllanwarne:
Now available in Cromwell's `develop` branch (thanks to https://github.com/broadinstitute/cromwell/pull/3916).

Since the WDL 1.0 spec was not opinionated on this matter I've also allowed this import style in WDL 1.0 workflows.

PR #223, opened 2018-06-30
Log and Debug statements

cjllanwarne:
See the changes in-place in [my branch's SPEC.MD](https://github.com/cjllanwarne/wdl/blob/cjl_log_statements/versions/development/SPEC.md#info-and-debug-logging).

Adds a new type of element into the workflow section, allowing users to indicate log and debug statements to the engine. I've left it entirely up to the engine (and/or whatever mode that engine happens to be running in) what to actually do with the logs it produces (*EDIT*: if anything).

Let me know what you think!

### DIFF
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -1809,11 +1809,11 @@ workflow log_something {
 
 In 'pseduo-console-output', the above workflow might produce:
 ```
-#> run static_logs.wdl
+#> run log_something.wdl
 
 starting call to x
 completed call to x
-LOG: x produced the value 55
+DEBUG: x produced the value 55
 starting call to y
 completed call to y
 LOG: y produced the value 66
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -1791,13 +1791,13 @@ workflow foo {
 
 ### Info and Debug Logging
 
-Log and debug statements allow a workflow author to signal to the engine to log a value at either *info* or *debug* level.
+Log and debug statements (`info` and `debug`) allow a workflow author to signal to the engine to log a value at either *info* or *debug* level.
 
 The exact mechanism of displaying logged statements to the user is left to the engine and may well be different for different workflows:
 
-* A command line invocation of a workflow may well 'log' to the console
-* A server running a workflow might 'log' into some form of completion report that it provides to the user on job completion. 
-* An engine if free to provide any additional context that it chooses to into the log statements (a timestamp or a fully qualified path to the current workflow, for example)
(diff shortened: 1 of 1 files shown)


geoffjentry:
Can you talk about the "why" here? What motivated this, how do you see it used, why do you envision it being used, etc?

cjllanwarne:
@geoffjentry 
Two thoughts on *why*:

First, to give authors and developers one small quality-of-life tool while writing and testing workflows. I think WDL's great at getting *syntax* out of people's way as they write workflows but they still have to make sure that the workflows they're writing are actually behaving properly.

Second, by allowing authors to opt-in to writing certain details, we may perhaps allow command lines and log files to be more aggressive in cutting out what they *don't* need to display the rest of the time. In other words by letting an author decide what is and isn't worth logging, a CLI no longer has to assume "log everything just in case" and can probably clear up a ton of (usually) irrelevant detail (unless the author wants it, of course!)

---

As a side note, there are various ways I've spotted of people "faking" log messages, or tricking specific engines into producing them in engine-specific ways. this suggestion is just one potential way to cut out the middle man and make it obvious to a reader in these cases what the author is trying to do (i.e. get a message out from the middle of the workflow).

For example right now if you want to see what some intermediate value is in the middle of a workflow, you're more or less forced to declare an intermediate string value and then pipe that through as a workflow output.

versions/development/SPEC.md
```diff
@@ -1790,6 +1789,83 @@ workflow foo {
...
+Log and debug statements (`info` and `debug`) allow a workflow author to signal to the engine to log a value at either *info* or *debug* level.
+
+The exact mechanism of displaying logged statements to the user is left to the engine and may well be different for different workflows:
+
+* A command line invocation of a workflow may well 'log' to the console's stdout, for example.
+* On the other hand a server running a workflow might 'log' into a completion report that it provides to the user on job completion. 
+* An engine if free to provide any additional context that it chooses as it processes the log statements (a timestamp or a fully qualified path to the current workflow, for example)
+* An engine may even decide not to expose any logs to the end-user at all, if that makes sense in its current execution mode.
+
+The value following the `log` or `info` can be any expression. It must evaluate to a `String` or `String?` value, or a value which can be coerced into `String` or `String?`
```
yfarjoun: should there be a discussion about **when** a log/debug statement is evaluated? calls are executed whenever their inputs are defined "enough", is it the same for log? Perhaps provide an example of an out-of order workflow and be explicit about when each thing is executed.

cjllanwarne:
Let's move this discussion to #224. In hindsight I think this sort of "is this a good general idea" discussion will work better in issue form.

PR #228, opened 2018-07-13
Proposal: remove object from the language

EvanTheB:
- Map, Object, and Struct overlap in functionality
- Object is a Map[String,*], breaking the otherwise typed nature of wdl.
- The spec is not clear or consistent about object literals
- 'object', 'variable', 'value' are intermingled in the spec, adding to confusion
- gatk-workflows do not use object

I was initially looking at something else and tried to fix a few descriptions of object (first commit). But then I thought, I do not see any evidence of use, and I do not see the utility, let us simplify the language.

### DIFF
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -1,20 +1,20 @@
 # Workflow Description Language
 
-## Table Of Contents
-
-<!---toc start-->
+Table of Contents
+=================
 
 * [Workflow Description Language](#workflow-description-language)
-  * [Table Of Contents](#table-of-contents)
   * [Introduction](#introduction)
   * [State of the Specification](#state-of-the-specification)
 * [Language Specification](#language-specification)
   * [Global Grammar Rules](#global-grammar-rules)
     * [Whitespace, Strings, Identifiers, Constants](#whitespace-strings-identifiers-constants)
     * [Types](#types)
-    * [Fully Qualified Names & Namespaced Identifiers](#fully-qualified-names--namespaced-identifiers)
+      * [Custom  Types](#custom--types)
+    * [Fully Qualified Names &amp; Namespaced Identifiers](#fully-qualified-names--namespaced-identifiers)
     * [Declarations](#declarations)
     * [Expressions](#expressions)
+      * [If then else](#if-then-else)
     * [Operator Precedence Table](#operator-precedence-table)
     * [Member Access](#member-access)
     * [Map and Array Indexing](#map-and-array-indexing)
@@ -22,7 +22,6 @@
     * [Function Calls](#function-calls)
     * [Array Literals](#array-literals)
     * [Map Literals](#map-literals)
-    * [Object Literals](#object-literals)
     * [Pair Literals](#pair-literals)
   * [Document](#document)
   * [Versioning](#versioning)
@@ -32,9 +31,11 @@
     * [Task Inputs](#task-inputs)
       * [Task Input Declaration](#task-input-declaration)
       * [Task Input Localization](#task-input-localization)
+        * [Special Case: Versioning Filesystems](#special-case-versioning-filesystems)
+    * [Non-Input Declarations](#non-input-declarations)
     * [Command Section](#command-section)
-      * [Expression Placeholders](#command-parts)
-      * [Expression Placeholder Options](#command-part-options)
+      * [Expression Placeholders](#expression-placeholders)
+      * [Expression Placeholder Options](#expression-placeholder-options)
         * [sep](#sep)
         * [true and false](#true-and-false)
         * [default](#default)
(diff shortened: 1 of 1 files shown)


geoffjentry:
@EvanTheB IMO the right thing to do here would be to effectively deprecate `Object` instead of removing it. It was discussed before and the thought was that it was too big a change even though we all agreed that it should eventually be removed. What about something where text is added to the `Object` description telling people what constructs to prefer and informing them that this remains purely for backwards compatibility?

Also, FWIW I don't think that `gatk-workflows do not use object` should matter here

orodeh:
I agree, object should be removed, the sooner the better.

cjllanwarne:
Although I totally agree that long-term WDL is much better off without `Object`, I could maybe retrofit a reason that "`Object` remained in 1.0 in case there was something incomplete about `struct`, preventing it being useful as a replacement".

I think I'm inclined to "deprecate" `Object` (whatever that means) and aim for removal in WDL 2.0. Hopefully by that time we'll have some real examples of `struct`s being used and a better sense that WDL can live happily without any untyped `Object`s.

EvanTheB:
https://github.com/openwdl/wdl/pull/160

I did try to search for an object removal discussion, but it was in the 'struct' thread.

Philosophically I am against deprecation, spec version A has y, spec version B does not. If engines want to warn version A users, or shoehorn support in version B with a translation, or crash in version B with a nice error message, that is good and can be done by engines. What value in a spec to say y is supported but in some other version of the spec it will not be?

I mention gatk-workflows because it is a large repo of wdl, my challenge was for any supporter of object to find a wdl that uses it.

@cjllanwarne Struct is great. WDL has an 'untyped object', it is String. I can put any data in a string (or a File). If I really need access to that data, that for some reason I do not know the structure of, I can read it with a task. Though how I can 'need' access to data if I don't know the structure is a bit beyond me.

patmagee:
I am really in favor of removal over depreciation in this instance. The object type is causing us a lot of headaches for one, but more then that is it's not really a functional type. Files are not localized, numbers are not cast appropriately. Ilnit to mention object access becomes a problem when trying to actually use it within any task.

I'm sure there are some people who use object to do some really nasty things in wdl.  But I think it would be okay to pigeonhole them into using wdl v1 ie they require the object. Between struct and map we have a complete replacement of object, so there should be no reason for them to continue to use it

versions/development/SPEC.md
```diff
@@ -1990,8 +1926,8 @@ workflow myWorkflow {
 }
 ```
 
-#### Struct Assignment from Object Literal
-Structs can be assigned using an object literal. When Writing the object, all entries must conform or be coercible into the underlying type they are being assigned to
+#### Struct Assignment from Map Literal
+Structs can be assigned using an map literal. When Writing the literal, all entries must conform or be coercible into the underlying type they are being assigned to.
```
EvanTheB: This change implies that map literals do not need to have uniformly typed values.

cjllanwarne:
Some thoughts: 

1. I liked the idea of having a testing period for `struct` to iron out any problems before removing the `Object` type that it's supposed to replace.
1. Having said that, if nobody is using `Object` in anger then "replacing their functionality" is a no-op. In particular I think the "`Object` is not really a functional type" argument is the strongest one here - if they don't work as people would expect then they're actively unhelpful when people find out about them.
1. Removing `Object` would make the next release `2.0` rather than `1.1`. Personally I think version numbers scare people way more than they should and IMO versions should be based on what changes we want to go into a release, not the other way round (ie having what goes into a release based on what version number we'd like it to have), so I have no problem with going straight from `1.0` to `2.0`. But others might disagree.

chapmanb:
I've been following this and agree with all the comments so far; Object removal sounds good. We did a proof of concept with it replicating CWL records, but all of the limitations Patrick mentioned it didn't work. This led to the struct discussions, so +1 on whatever helps us move there as fast as possible.

versions/development/SPEC.md
```diff
@@ -646,51 +619,15 @@ Array[Int] b = [0,1,2]
 
 ### Map Literals
 
-Maps values can be specified using a similar Python-like sytntax:
+Maps values can be specified using a similar Python-like syntax, note the last
```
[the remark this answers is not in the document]
EvanTheB: Ah yes I made an error here, thank you. I remember being confused. 

What is the purpose of the object literal syntax? using map literals I can specify an expression, so any literal or reference to a declared name. Using object literals I can omit the `""`, but whether I write "a" or a I can only use literals. 

Map literals seem like a superset, therefore we should remove object literals. I must be missing something.

versions/development/SPEC.md
```diff
@@ -879,7 +815,7 @@ task test {
 ```
 
 > **NOTE**: the expression result must ultimately be converted to a string in order to take the place of the placeholder in the command script.
-This is immediately possible for WDL primitive types (e.g. not `Array`, `Map`, or `Object`).
+This is immediately possible for WDL primitive types (e.g. not `Array`, `Map`, or `Struct`).
```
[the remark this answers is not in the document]
EvanTheB: `arrays` and `maps` too then? they are also not a single type

geetduggal:
👍

geoffjentry:
I'm removing my objection to the removal of objects. It does seem like conversation is still flowing here so we'll wait a bit before moving to voting

EvanTheB:
I think still to be addressed is @cjllanwarne comments about the object literal. I am not clear on the value, it allows the "" to be omitted around the key.

I also noticed that structs can only be assigned to from map literals, where that should probably read any coercible map (literal, name, function, call etc).

cjllanwarne:
Re `object { }` literals, they:
* Allow people to write without having to add `"`s constantly (for a 'user friendly' language, this shouldn't be underestimated)
* Enforce the key types *as you type* (you can't put anything other than a valid String there so you can't make face-palm style mistakes. Even mistakes that could be found by validation would only be found an entire context shift later)
  * As a side note: even `Map`s with string keys aren't always coercible into objects (what if the key is `"hello, there"` - that's not a valid struct field name but we can't know that until evaluation mid-workflow)
* Don't allow weird "key-lookup" behavior (the fields should be known, so encouraging struct creation using a type that lets keys be fetched from other values is weird IMO)
* Let us guarantee that a struct is made with the right fields at type-checking time. We can't do that with a `Map` (because the keys are expressions which are run-time things)

I want to re-emphasize that last point. It's a bad experience to only find out mid-workflow that some struct you added has the wrong set of fields because we couldn't validate the literal upfront. Yes, we *could* have special cases like "try to evaluate things upfront, if we happen to spot that it's a Map that's about to be coerced into a struct" or "if they fields are all string literals, try to guarantee that the fields are right" but that's extra logic to implement, prone to mistakes and we still can't catch everything. 

I'm *strongly* in favor of a literal that guarantees that `struct` fields are correct *as they are typed in*. They can be renamed `MyStruct foo = struct { a: ..., b: ... }`, or a totally different syntax like `MyStruct foo = MyStruct { a: ..., b: ... }`, `MyStruct foo = new MyStruct { a: ..., b: ... }`, etc, etc... but relying on `Map` literals (which are intended for a totally different usage) followed by an expensive coercion step is not a good answer, IMO.

EvanTheB:
@cjllanwarne Your last two points convinced me, when typing a 'struct literal', expressions for the keys do not make much  sense, and it is nice to simpify the parse time type checking. I am sad to have two similar syntaxes in the 'simple' language, but maintaining the status quo wins it for me. Anyone else thoughts?

geoffjentry:
Poking this as the activity seems to have died down.

@EvanTheB I'm not sure I completely followed the discourse, am I correct in understanding that you're ok w/ closing this?

EvanTheB:
@geoffjentry Where I am at now I would like to re-introduce object literals, but still get rid of the object type and associated baggage. I will make an update soon.

EvanTheB:
My first rebase, hope I didn't get it wrong.

EvanTheB:
@geoffjentry Yes ready to go. (aside I would propose that the contents page should be autogenerated outside of PRs, I will take it to the gitter)

geoffjentry:
@EvanTheB I'm going to open this up for voting (sorry for **this** delay as well). Would you mind rebasing onto the current spec?

versions/development/SPEC.md
```diff
@@ -832,16 +794,15 @@ For example imagine two versions of file `fs://path/to/A.txt` are being localize
...
+For example, this task takes an input, but performs some calculation which can then be used by the command:
 
 ```wdl
 task t {
   input {
-    Object inputs
+    Int size
   }
-  File objects_json = write_json(inputs)
+  size_clamped = if size > 10 then 10 else size
```
cjllanwarne: Typo: needs a type

versions/development/SPEC.md
```diff
@@ -1986,54 +1961,54 @@ struct Person {
...
 workflow myWorkflow {
-    Person a
-    call task_a {
+    Person harry = object {name: "Harry", age: 11}
+    call myTask {
         input:
-            a = a
+            a = harry
     }
+    Person harry_p = myTask
```
cjllanwarne: Typo: `Person harry_p = harry`?

versions/development/SPEC.md
```diff
@@ -3054,15 +2869,14 @@ Array[Pair[Int, String]] zipped = zip(xs, ys)     # i.e.  zipped = [ (1, "a"), (
 
 ## Array[Pair[X,Y]] cross(Array[X], Array[Y])
 
-Given any two Object types, the `cross` function returns the cross product of those Object types in the form of a Pair object.
+Return the cross product of the two arrays. Array[Y][1] appears before
```
cjllanwarne: Typo: should be `Array[1][Y]`? 
(But TBH this sentence could probably do with some revising)

orodeh:
👍

patmagee:
👍

versions/development/SPEC.md
EvanTheB: With this one I was trying to illustrate how tasks return is compatible with Struct.

geoffjentry:
@evantheb please revert your last comment. Once voting is active changes are not allowed

geoffjentry:
This PR passes unanimously and is waiting for implementation.

@cjllanwarne You had some comments which arose after voting was live and would need to be patched in separately. Not sure how you want to handle that.

orodeh:
👍

delagoya:
Ping on this issue. I see that there is a merge conflict. @EvanTheB Does that need to be resolved before a full merge can happen?

geoffjentry:
@delagoya An implementation needs to actually implement these changes prior to merging

delagoya:
Sorry should have bee more clear: There is a merge conflict on SPEC.md on this PR with master branch (assuming master is the canonical release). Can that be addressed now?

geoffjentry:
@delagoya `master` is indeed the correct branch - we decided to track different versions in the `versions` subdirectory instead of via branches. 

It could be rebased now to remove the conflict, but it'll likely need to be rebased again by the time it gets merged (going back to my point that merge conflict is not the reason it hasn't been merged). Are you more interested in removing the conflict or seeing it merged?

geoffjentry:
@delagoya If you'd like to contribute the implementation of this PR to one of the WDL implementations out there it'd hurry things along :)

delagoya:
define "WDL implementations" ... does [chanzuckerberg/miniwdl](https://github.com/chanzuckerberg/miniwdl) count?

geoffjentry:
@delagoya Good question. At the moment I'd say "no" since it doesn't actually **run** the WDL. But it's not my call to make. Let me consult with the WDL lizard people

geoffjentry:
@delagoya I think we're going to wait for a bit prior to using miniwdl as an implementation. There are certainly aspects where simple static analysis is sufficient but that introduces grey area into the spec. And of course, with this PR there's the matter of "I didn't put it in yet" is effectively "I implemented it" since it's a pure removal :) (Also, similarly it really likely doesn't need implementation in a practical sense, but grey areas).

delagoya:
That's fine, but I think it would be worth our while to define "implementation" and have an official test case suite for implementations to pass. Maybe port over a portion of the centaur tests from Cromwell? Enough to test out all of the features of WDL without being too burdensome to run.

geoffjentry:
@delagoya :100:% agree. A conformance suite is a topic which has come up numerous times, the blocker has always been someone (or someones) with the time to put towards the effort. 

All reasonable suggestions I've heard start by peeling out the WDLs which aren't testing Cromwell functionality from Centaur and working from there. My personal favorite has been the suggestion to use the WES orchestrator to administer the workflows - there needs to be **some** interface (a la how CWL uses a well known command line interface). 

I'd suggest moving this convo to the mailing list if we want to go further down this path

delagoya:
Good topic to tackle for the GA4GH Cloud Hackathon in Jan if not before then. Will move this convo elsewhere though

geoffjentry:
@delagoya I don't think that's intended to be a hackathon, rather a F2F on a very specific topic (auth). I could be wrong though. Either way it'll provide some face time :)

patmagee:
Closing this. Please see: https://github.com/openwdl/wdl/pull/283

PR #229, opened 2018-07-13
Clarify the expression placeholder and string interpolation behaviour

EvanTheB:
Using ${} in a command or a string literal is similar. Make it the same. Allow any expression. Disallow custom syntax ('sep' and friends). Add sep as a library function. 

Simplify the syntax and clean up some descriptions. There is still a massive way to go to make the spec self-consistent, but I wanted to get a comment before I put in the miles. A language with simple as its goal probably shouldn't have a 4000 line spec. I think golang is simpler.

### DIFF
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -403,7 +403,7 @@ Declarations are declared at the top of any [scope](#scope).
 
 In a [task definition](#task-definition), declarations are interpreted as inputs to the task that are not part of the command line itself.
 
-If a declaration does not have an initialization, then the value is expected to be provided by the user before the workflow or task is run.
+If a non-optional declaration does not have an initialization, then the value must be provided by the user before the workflow or task is run.
 
 Some examples of declarations:
 
@@ -466,15 +466,16 @@ $expression = $expression '/' $expression
 $expression = $expression '+' $expression
 $expression = $expression '-' $expression
 $expression = $expression '<' $expression
-$expression = $expression '=<' $expression
+$expression = $expression '<=' $expression
 $expression = $expression '>' $expression
 $expression = $expression '>=' $expression
 $expression = $expression '==' $expression
 $expression = $expression '!=' $expression
 $expression = $expression '&&' $expression
 $expression = $expression '||' $expression
-$expression = '{' ($expression ':' $expression)* '}'
-$expression = '[' $expression* ']'
+$expression = '{' ($expression ':' $expression (',' $expression ':' $expression )*)? '}'
+$expression = object '{' (($string | $identifier) ':' $expression (',' ($string | $identifier) ':' $expression )*)? '}'
+$expression = '[' ($expression (',' $expression)*)? ']'
 $expression = $string | $integer | $float | $boolean | $identifier
 ```
 
@@ -490,7 +491,7 @@ Below are the valid results for operators on types.  Any combination not in the 
 |`Boolean`|`<=`|`Boolean`|`Boolean`||
 |`Boolean`|`||`|`Boolean`|`Boolean`||
 |`Boolean`|`&&`|`Boolean`|`Boolean`||
-|`File`|`+`|`File`|`File`|Append file paths|
+|`File`|`+`|`File`|`File`|Concatenate file paths|
 |`File`|`==`|`File`|`Boolean`||
 |`File`|`!=`|`File`|`Boolean`||
 |`File`|`+`|`String`|`File`||
@@ -545,6 +546,7 @@ Below are the valid results for operators on types.  Any combination not in the 
 |`String`|`+`|`Float`|`String`||
 |`String`|`+`|`Int`|`String`||
 |`String`|`+`|`String`|`String`||
+|`String`|`+`|`File`|`File`||
 |`String`|`==`|`String`|`Boolean`||
 |`String`|`!=`|`String`|`Boolean`||
 |`String`|`>`|`String`|`Boolean`||
@@ -604,16 +606,13 @@ runtime {
 
 ### Member Access
 
-The syntax `x.y` refers to member access.  `x` must be an object or task in a workflow.  A Task can be thought of as an object where the attributes are the outputs of the task.
+The syntax `x.y` refers to member access. `x` must be an object, previously called task in a workflow, or instance of a struct. A task can be thought of as an struct where the attributes are the outputs of the task.
 
 ```wdl
 workflow wf {
   input {
     Object obj
-    Object foo
   }
-  # This would cause a syntax error,
-  # because foo is defined twice in the same namespace.
   call foo {
     input: var=obj.attr # Object attribute
   }
@@ -636,7 +635,7 @@ Given a Pair `x`, the left and right elements of that type can be accessed using
 
 Function calls, in the form of `func(p1, p2, p3, ...)`, are either [standard library functions](#standard-library) or engine-defined functions.
 
-In this current iteration of the spec, users cannot define their own functions.
+Users cannot define their own functions.
 
 ### Array Literals
 
@@ -848,10 +847,7 @@ task t {
 ### Command Section
 
 The `command` section is the *task section* that starts with the keyword 'command', and is enclosed in either curly braces `{ ... }` or triple angle braces `<<< ... >>>`.
-It defines a shell command which will be run in the execution environment after all of the inputs are staged and before the outputs are evaluated.
-The body of the command also allows placeholders for the parts of the command line that need to be filled in.
-
-Expression placeholders are denoted by `${...}` or `~{...}` depending on whether they appear in a `command { }` or `command <<< >>>` body styles.
+It defines a bash shell script which will be run in the execution environment after all of the inputs are staged and before the outputs are evaluated.
 
 #### Expression Placeholders
 
@@ -862,7 +858,7 @@ Expression placeholders differ depending on the command section style:
 |`command { ... }`|`~{}` (preferred) or `${}`|
 |`command <<< >>>`|`~{}` only|
 
-These placeholders contain a single expression which will be evaluated using inputs or declarations available in the task.
+These placeholders are the mechanism for using task inputs and variables to construct a command. They contain a single expression which will be evaluated using inputs or declarations available in the task.
 The placeholders are then replaced in the command script with the result of the evaluation.
 
 For example a command might reference an input to the task, like this:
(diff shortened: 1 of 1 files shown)


orodeh:
Can you give examples for the old syntax vs. the new syntax?

versions/development/SPEC.md
```diff
@@ -487,7 +488,7 @@ Below are the valid results for operators on types.  Any combination not in the
 |`Boolean`|`<=`|`Boolean`|`Boolean`||
 |`Boolean`|`||`|`Boolean`|`Boolean`||
 |`Boolean`|`&&`|`Boolean`|`Boolean`||
-|`File`|`+`|`File`|`File`|Append file paths|
+|`File`|`+`|`File`|`File`|Concatenate file paths|
```
cjllanwarne: This has always made no sense to me. IMO `File + File` should be removed

versions/development/SPEC.md
```diff
@@ -601,7 +603,7 @@ runtime {
 
 ### Member Access
 
-The syntax `x.y` refers to member access.  `x` must be an object or task in a workflow.  A Task can be thought of as an object where the attributes are the outputs of the task.
+The syntax `x.y` refers to member access. `x` must be an object, task in a workflow, or struct. A Task can be thought of as an object where the attributes are the outputs of the task.
```
cjllanwarne: I think something like `or struct instance.` would be more accurate here.

orodeh:
Found the details in the pull request body. It simplifies cases where we have redundant syntax. Looks good to me.

versions/development/SPEC.md
EvanTheB: I think you are right. It would simplify this table, and clarify much of the spec, if the way that string and File were related was explicit. For example: ```Files will be coerced into Strings, and Strings into files, based on the destination type of the expression```. One issue is that some of the library functions have different meaning for string and file -> read_json from a string, or from the file named by this 'String(File)'

EvanTheB: Oh and I meant to say, step 2: ```String + String = String``` would then be the only defined + operator.

versions/development/SPEC.md
```diff
@@ -834,158 +833,63 @@ task t {
 
 ### Command Section
 
-The `command` section is the *task section* that starts with the keyword 'command', and is enclosed in either curly braces `{ ... }` or triple angle braces `<<< ... >>>`.
-It defines a shell command which will be run in the execution environment after all of the inputs are staged and before the outputs are evaluated.
-The body of the command also allows placeholders for the parts of the command line that need to be filled in.
-
-Expression placeholders are denoted by `${...}` or `~{...}` depending on whether they appear in a `command { }` or `command <<< >>>` body styles.
+The `command` section is the *task section* that starts with the keyword 'command', and is triple angle braces `<<< ... >>>`.
```
EvanTheB: Apologies, friday night drinks meant some unrelated changes snuck in.

geetduggal:
👍 Thanks for the thoughts on the simplification.

versions/development/SPEC.md
patmagee: +1 to removing File + File. It really does not make much sense to me. If `FIle` was treated as any `URI` or `PATH` then it might make snese. But in this case any `File` is an actual location in the cloud or on the local instance. Concatenating them is silly..

versions/development/SPEC.md
patmagee: Is this suggesting that `command { }` be deprecated in favor of `command <<< ....>>>`? Additionally, what type of expresison placeholders are you proposing?

versions/development/SPEC.md
```diff
@@ -3126,6 +3030,20 @@ This function will return `false` if the argument is an unset optional value. It
...
+
+```
+sep(' ', ["a", "b", "c"])
+> "a b c"
+
+A = ["file_1", "file_2"]
+sep(' ', prefix('-i ', A))
+> "-i file_1 -i file_2"
+```
+
```
patmagee: I like the idea of converting this to a more delcarative engine function, as opposed to the old way. I wonder if we go through the trouble of making this an engine funciton, why not remane it to `join` as well. Join is much more indicative of the action taking place over `sep`

```
join('  ', ["a","b","c"])
>"a  b  c"
```

versions/development/SPEC.md
EvanTheB: This is a mistake, i do think that but it should not be in this PR.

versions/development/SPEC.md
EvanTheB: join is familiar to python people, but it is also a famous stumbling block for newcomers - wdl's target audience. I am fine with it either way, but chose sep as status quo should be preferred.

geoffjentry: I don't really have a dog in this particular fight but my eyes perked up at the mention of Python. FWIW in the early days we tried to channel "what does Python do" when facing some situation where we couldn't decide what to do.

geoffjentry:
@EvanTheB Doesn't seem to be a lot of chatter here - does this seem appropriate to bring to a vote?

EvanTheB:
@geoffjentry I am willing, in summary:

* ${} in a command and a string literal do the same thing - arbitrary expression evaluation
* Remove special `sep`, `if`, `true/false` syntax from same
* Add those as library functions

Should I squash and rebase or would that be done by the merger?

geoffjentry:
@EvanTheB I don't think we need to rebase pre-voting, but it'll need to be done eventually. I'll put this out.

orodeh:
👍

DavyCats:
:+1: looks good to me

geoffjentry:
:+1:

cjllanwarne:
Sorry to have to 👎 this - I really like the majority of the PR but there are a few (apparently ride-along) changes in there that I don't like as-is. 

FWIW it wouldn't take much to make me an "aye" on this:

What I *really* like and *really* want to 👍:
- Making placeholders simply wrappers around expressions
- Removing the placeholder attributes and making the expression syntax powerful enough to replace them

What I don't like and why I am 👎on this PR as-is:
- No-longer having the "what can be immediately interpolated" vs "what needs to be transformed before interpolation" explicit
- A couple of the ride-along changes that don't seem to support the original intention (ie to simplify placeholders). A few examples:
  - Changing the way optionals are interpolated. Currently un-supplied optionals are interpolated as empty strings. This PR changes that (without enough fanfare, IMO).
  - Changing the hermes-style expression syntax to disallow empty arrays and maps
    - eg using the new `$expression = '[' ($expression (',' $expression)*)? ']'`
    - we can no longer declare `Array x = []`
    - In any case, this whole section is a bit misleading as it's not actually how the hermes grammar defined these expressions! IMO these don't belong in the spec, especially not if they're out of date.
  - Adding another reference to `bash`, which I think we'd ideally be trying to move away from
  - Apparently endorsing the `File + File` = `File` by clarifying its entry.

geoffjentry:
@cjllanwarne @EvanTheB Just note that now that voting is active further changes aren't allowed. 

If the PR passes - @cjllanwarne you can make a followup proposal to make the changes you want
If the PR fails - @EvanTheB You can choose to make edits that you think will help and try again

geoffjentry:
Actually I noticed just now that I'd been lax and that the voting period had already expired. Thus this PR passes.

@EvanTheB - Can you rebase this so that it doesn't have a merge conflict?

@cjllanwarne As I suggested, if you'd like to see changes to this, please open a followup PR

EvanTheB:
@cjllanwarne Yes there was too much else included here as you say, I failed to remove all of it before voting. I will save comments for a new PR

illusional:
Hi all,

I'm interested in reviving this PR and happy to help drive this process. I've got selfish reasons where I need to join an array of strings together.

I'm interested in at least getting a separator function before 2.0 (even if it's called something else), and if it should be able to coexist with the string interpolation option until it's removed in 2.0.

I've added a `sep` function and have issued PRs for:

- Cromwell (https://github.com/broadinstitute/cromwell/pull/5494)
- MiniWDL (https://github.com/chanzuckerberg/miniwdl/pull/381)

What's the process from here.

patmagee:
hey @illusional, I think the best approach is to get your change merged into a "development" version of the specification. With Cromwell they normally have a running dev version, but I would as @cjllanwarne what the best approach here would be. the `sep` engine funcition won't be backported to 1.0, so unfortuantely the fastest way to get it into your WDL's is a "development" version of the spec

illusional:
Thanks @patmagee, I guess that's why we have the development spec. 

As such, I've further actioned the MiniWDL branch to remove the interpolator options from the grammar, and have implemented the `sep` function. The grammar has already been corrected so I presume Cromwell should be fine.

### Tests

In terms of testing, I could add the following two conformance tests, both of which are expecting `out` to be `"value1,value2,value3"` (these are in fact the test cases I use for miniWDL).

1. Simple test the new `sep` function

    ```wdl
    version development
    task SepTest {
        input {
            Array[String] inp = ["value1", "value2", "value3"]
        }
        command {}
        output {
            String out = sep(",", inp)
        }
    }
    ```

2. Test that the sep function works in an interpolator block

    ```wdl
    version development
    task SepTest {
        input {
            Array[String] inp = ["value1", "value2", "value3"]
        }
        command <<<
            echo ~{sep(",", inp)}
        >>>

        output {
            String out = read_string(stdout())
        }
    }
    ```

I'll endeavour to get these PRs approved in the engines so this PR can be completed.

mlin:
Hello all, https://github.com/chanzuckerberg/miniwdl/pull/381 by @illusional has implemented in miniwdl's development grammar,

1. removal of all the key/value placeholder options
2. adding the `sep()` library function

This PR has numerous other text improvements, which is great but I'm a little unclear if there are additional semantic changes "waiting for implementation." Does anyone have a read on that?

versions/development/SPEC.md
```diff
@@ -1082,7 +991,7 @@ Therefore to ensure that a WDL is portable when using `glob()`, a docker image s
 
 ### String Interpolation
 
-Within tasks, any string literal can use string interpolation to access the value of any of the task's inputs.  The most obvious example of this is being able to define an output file which is named as function of its input.  For example:
+Any string literal can use string interpolation to access the value of any expression. The most obvious example of this is being able to define an output file which is named as function of its input. This uses the same syntax and semantics as expression placeholders in the command section of a task. For example:
```
patmagee, line 994: @mlin @cjllanwarne can you please confirm that cromwell or miniwdl implements this feature

mlin: Yes-ish...miniwdl honors interpolations in any string literal except those in meta/parameter_meta sections

illusional:
Hey @patmagee, is there anything outstanding for this PR? I've implemented `sep` in the following PRs:

- MiniWDL (https://github.com/chanzuckerberg/miniwdl/pull/381): _merged_
- Cromwell (https://github.com/broadinstitute/cromwell/pull/5494): _awaiting review_

versions/development/SPEC.md
cjllanwarne: Yes, should be the same in Cromwell

cjllanwarne:
This looks almost ready to merge, pending conflict resolution. It looks like another PR has perhaps erased a few type conversions which this PR was correcting?

illusional:
Hi @cjllanwarne, @mlin and @EvanTheB, I have moved the changes into a new PR: #366 to remove the conflicts. There are no additional changes, so I'm really hoping it won't need to go through the review process again.

mlin:
Closing #229 as superseded by #366

PR #231, opened 2018-07-13
Add task to the list of scopes

pshapiro4broad:
If section "Scope" lists all WDL scopes, it must include `task`.

Also, I think the code sample was in error, in that it was providing `var` to `my_task` but `var` wasn't a declared input.

### DIFF
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -2112,6 +2112,7 @@ However, you can import two workflows with different namespace identifiers that 
 Scopes are defined as:
 
 * `workflow {...}` blocks
+* `task {...}` blocks
 * `call` blocks
 * `while(expr) {...}` blocks
 * `if(expr) {...}` blocks
@@ -2122,7 +2123,7 @@ Inside of any scope, variables may be [declared](#declarations).  The variables 
 ```wdl
 task my_task {
   input {
-    Int x
+    Int var
     File f
   }
   command {
@@ -2139,7 +2140,7 @@ workflow wf {
     Int x = 3
     call my_task {
       Int x = 4
-      input: var=x, f=file
+      input: var = x, f = file
     }
   }
 }


versions/development/SPEC.md
```diff
@@ -2112,6 +2112,7 @@ However, you can import two workflows with different namespace identifiers that
 Scopes are defined as:
 
 * `workflow {...}` blocks
+* `task {...}` blocks
 * `call` blocks
```
EvanTheB: Why is a call block a scope?
```
call x { Int a = 1 input: b = a }
```
It is not valid according to the Call Statement description, and I can't see the point
```

pshapiro4broad: I was wondering that too. Since `call` declares a bound name in the current scope, maybe that's the intent of including it here?

geoffjentry:
Agree that scope seems like a mess. A variety of things pop out:

- While loops are no longer even in the spec, but still remain [in the grammar](https://github.com/openwdl/wdl/blob/master/versions/1.0/parsers/grammar.hgr#L430)
- Also part of the grammar file, `declaration` (the thing which Scope is defined as providing) is a member of [wf_bod_element](https://github.com/openwdl/wdl/blob/master/versions/1.0/parsers/grammar.hgr#L424) which is available to [workflow](https://github.com/openwdl/wdl/blob/master/versions/1.0/parsers/grammar.hgr#L423), the aforementioned [while loops](https://github.com/openwdl/wdl/blob/master/versions/1.0/parsers/grammar.hgr#L430), [if blocks](https://github.com/openwdl/wdl/blob/master/versions/1.0/parsers/grammar.hgr#L431), and [scatter blocks](https://github.com/openwdl/wdl/blob/master/versions/1.0/parsers/grammar.hgr#L433).
- The grammar doesn't seem to enforce declaration location, as per what @EvanTheB pointed out, the scopes all just define a `list(wf_body_element)`.
- Regardless of the ultimate fate of calls, it's clear that there's some cleaning up of contradictions which needs to happen
- Tasks appear to be scope-adjacent, but the spec does say `In a task definition, declarations are interpreted as inputs to the task` which I think implies not a scope (to the original point of @pshapiro4broad 's PR). I went back and forth on this so don't read this as a strong statement

geoffjentry:
I opened up #232 to kill the while loop situation

EvanTheB:
@geoffjentry I hadn't even known the grammar existed there as part of the spec! It makes the grammar snippets in the markdown document smell fishy, SPOT and all that. Is there a notion of whether the grammar or the markdown file is authoritative?

Tasks don't really need a scope, in that they do not have 'nesting' (scatters and ifs, ignoring 'call' for now). But it does simplify the description if there is a single statement describing scopes, and tasks just don't care much because they have no nesting. Else a definition of 'declaration' needs to be made orthogonal to scope, so that both scope and task can use it which seems hard.

pshapiro4broad:
> Tasks don't really need a scope, in that they do not have 'nesting'
> Tasks appear to be scope-adjacent

I don't see how tasks are not scopes (that is, a task must introduce a new scope). A task allows the user to bind names (aka `variables`) to values. The task can then use the names to obtain the bound value. The fact that tasks can't have nested scopes doesn't seem relevant. 

Maybe the problem here is that the doc shouldn't say "x is a scope" but instead should say "x introduces a new scope". In most language definitions I've seen, a scope is defined as the range over which a bound name is defined.

EvanTheB:
@pshapiro4broad Yes I think you get my point, because WDL doesn't define what a scope is, we are left guessing and using competing made up definitions. Your 'range over which a bound name is defined' looks great, will it be added to this PR or a new one?

orodeh:
I agree:
1) calls are not scopes
2) it is not legal to place a declaration inside a call "scope"

geoffjentry:
@EvanTheB re the grammar file, it's murky as there's no official statement as to which is the ground truth. IMO in a perfect world there's a grammar file which is the official source of truth and then a prose spec to flesh it out. However I don't think the hermes grammar is appropriate for that as very few people understand hermes well enough to make any meaningful changes. 

I've been hoping some motivated individual would submit a BNF or something like that, but that hasn't yet happened. I personally view the hermes grammar as a sorta/kinda reference implementation of the spec - I'm only aware of one WDL implementation not using the hermes grammar (the IntelliJ plugin) so for all intents and purposes the hermes grammar **is** the reality that people experience. 

re the grammar snippets in the spec doc, I've **never** liked having those there as that too tightly binds the prose w/ the specific grammar used. IMO whenever folks are putting together PRs in areas where the grammar is directly referenced in the spec file it should be removed in favor of real examples.

EvanTheB:
@geoffjentry Yes I think I will have to abandon lark, I thought EBNF had a ISO standard, but if none of the tools we want support it then what is the point. I am guessing we would target Java, Python, C target languages, in that order.

geoffjentry:
@pshapiro4broad "I don't see how tasks are not scopes " - I'm saying that at the moment scopes are defined as things which can have a declaration, and declarations are not legal in the strict sense in the task body.

geoffjentry:
@pshapiro4broad What do you want to do with this? Admitting that I was a large part of the problem the discussion here is pretty wide ranging and there doesn't seem to be much consensus. I can either call for a vote as is, you can make mods, or I can close it. Up to you.

cjllanwarne:
@pshapiro4broad I think my changes in #257 might supersede this PR - what do you think?

pshapiro4broad:
Yes, I agree that #257 supersedes this PR. Thanks @cjllanwarne !

PR #232, opened 2018-07-17
Remove remaining vestiges of while

geoffjentry:


cjllanwarne:
I whole-heartedly endorse this.

As a meta comment, it was one of my hopes to go through the entire spec prior to 1.0 and remove all of the blatant lies, dreams, bad ideas, and "not-done-yet"s. Needless to say, that didn't happen but I still think it'll be a very valuable exercise at some point (and as soon as possible).

geoffjentry:
I'm going to put this on the "cleanup" fast track path since we already decided `while` should be removed. If I don't see any objection by the time I get around to merging this later today, I'll just merge in.

cjllanwarne:
👍 on the change and 👍 on fast-tracking.

If this is "it was never meant to be thus", should we also remove these from the `1.0` spec?

geoffjentry:
@cjllanwarne I don't think we should make retroactive changes to a spec.

PR #235, opened 2018-07-25
Clarify type conversions and meanings

EvanTheB:
I have attempted to make some definitions of the behaviour when different types are mixed in expressions. 

- `Int` are converted to `Float`
- `Int` and `Float` are converted to `String`
- `File` act like `String` everywhere

Many of these choices are controversial, and I have not thought through the repercussions or investigated the cromwell implementation, but I think a starting point is needed for discussion. 

I think the most important is formatting of floats. The approach I have used is that printf 'g' is used, and any more difficult requirements should call out to a real language. Maybe printf 'f' is more usable, what does cromwell do? I would be happy to make WDL even more restricted than this, but not more expressive. 

I also removed the rarer boolean operations, most languages make do with, `and`, `or`, and `not`, and maybe `xor`. Especially for a 'simple' language. It is straightforward to derive the rest. 

I removed floating point remainder, and unary `+`. 

To finish this PR a pass over the serializations, library functions, and json prose is probably also required.

### DIFF
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -302,10 +302,10 @@ For more details on the postfix quantifiers, see the section on [Optional Parame
 
 For more information on type and how they are used to construct commands and define outputs of tasks, see the [Data Types & Serialization](#data-types--serialization) section.
 
-#### Numeric Behaviour
+#### Numeric Behavior
 
 `Int` and `Float` are the numeric types.
-`Int` can be used to hold any integer, there is no size limit.
+`Int` can be used to hold a signed Integer in the range \[-2^63, 2^63). 
 `Float` is a finite 64-bit IEEE-754 floating point number.
 
 #### Custom Types
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -489,6 +489,8 @@ Below are the valid results for operators on types.  Any combination not in the 
 |`Boolean`|`!=`|`Boolean`|`Boolean`||
 |`Boolean`|`||`|`Boolean`|`Boolean`||
 |`Boolean`|`&&`|`Boolean`|`Boolean`||
+|`File`|`==`|`File`|`Boolean`||
+|`File`|`!=`|`File`|`Boolean`||
 |`Float`|`+`|`Float`|`Float`||
 |`Float`|`-`|`Float`|`Float`||
 |`Float`|`*`|`Float`|`Float`||
@@ -533,16 +535,14 @@ Below are the valid results for operators on types.  Any combination not in the 
 |`String`|`>=`|`String`|`Boolean`||
 |`String`|`<`|`String`|`Boolean`||
 |`String`|`<=`|`String`|`Boolean`||
-|`File`|`==`|`File`|`Boolean`||
-|`File`|`!=`|`File`|`Boolean`||
 ||`-`|`Float`|`Float`||
 ||`-`|`Int`|`Int`||
 ||`!`|`Boolean`|`Boolean`||
 
 Note: In expressions using an operator with mismatched numeric types,
 the `Int` will be cast to `Float` and the result will be `Float`.
-This will could cause loss of precision. The `Float` can be
-converted to `Int` with the `ceil`, `round` or `floor` functions if needed.
+This will cause loss of precision if the `Int` is too large to be represented exactly by the `Float`.
+The `Float` can be converted to `Int` with the `ceil`, `round` or `floor` functions if needed.
 
 #### If then else
 
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -302,35 +302,11 @@ For more details on the postfix quantifiers, see the section on [Optional Parame
 
 For more information on type and how they are used to construct commands and define outputs of tasks, see the [Data Types & Serialization](#data-types--serialization) section.
 
-#### Numeric Types
+#### Numeric Behaviour
 
 `Int` and `Float` are the numeric types.
-Int can be used to hold any integer, there is no size limit.
-Float is used to hold a finite 64-bit IEEE-754 floating point number.
-Note that WDL itself is not intended to be used for heavy numeric computation.
-There are no number formatting utilities, many mathematical functions are absent,
-and engine behaviour at extreme values is not defined and should not be relied upon.
-
-The mathematical operators are defined only for identical types.
-In expressions using an operator with mismatched numeric types,
-the `Int` will be cast to `Float` and the result will be `Float`.
-Note that this will cause loss of precision. The `Float` can be
-converted to `Int` with the `ceil`, `round` or `floor` functions if needed.
-
-In expressions involving a numeric type and a String, the numeric type
-will be converted to a string. This is straigthforward for `Int`.
-For `Float`, behaviour will be similar to the 'g' conversion specifier
-defined by the C language.
-If you have stricter requirements for `Float` formatting, you should use `String`
-and call out to your language or tool of choice for numerical manipulation.
-
-#### String Types
-
-`String` and `File` are the string types.
-`File` type can be used in place of `String` in expressions.
-The engine may fail if the value assigned to a `File` type cannot be handled.
-`File` values will normally be unix absolute paths, or URIs.
-Engines should document the supported values for `File`.
+`Int` can be used to hold any integer, there is no size limit.
+`Float` is a finite 64-bit IEEE-754 floating point number.
 
 #### Custom Types
 
@@ -523,6 +499,22 @@ Below are the valid results for operators on types.  Any combination not in the 
 |`Float`|`>=`|`Float`|`Boolean`||
 |`Float`|`<`|`Float`|`Boolean`||
 |`Float`|`<=`|`Float`|`Boolean`||
+|`Float`|`+`|`Int`|`Float`||
+|`Float`|`-`|`Int`|`Float`||
+|`Float`|`*`|`Int`|`Float`||
+|`Float`|`/`|`Int`|`Float`||
+|`Float`|`>`|`Int`|`Boolean`||
+|`Float`|`>=`|`Int`|`Boolean`||
+|`Float`|`<`|`Int`|`Boolean`||
+|`Float`|`<=`|`Int`|`Boolean`||
+|`Int`|`+`|`Float`|`Float`||
+|`Int`|`-`|`Float`|`Float`||
+|`Int`|`*`|`Float`|`Float`||
+|`Int`|`/`|`Float`|`Float`||
+|`Int`|`>`|`Float`|`Boolean`||
+|`Int`|`>=`|`Float`|`Boolean`||
+|`Int`|`<`|`Float`|`Boolean`||
+|`Int`|`<=`|`Float`|`Boolean`||
 |`Int`|`+`|`Int`|`Int`||
 |`Int`|`-`|`Int`|`Int`||
 |`Int`|`*`|`Int`|`Int`||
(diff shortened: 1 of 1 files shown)


versions/development/SPEC.md
```diff
@@ -302,7 +302,37 @@ For more details on the postfix quantifiers, see the section on [Optional Parame
 
 For more information on type and how they are used to construct commands and define outputs of tasks, see the [Data Types & Serialization](#data-types--serialization) section.
 
-#### Custom  Types
+#### Numeric Types
+
+`Int` and `Float` are the numeric types.
+Int can be used to hold any integer, there is no size limit.
```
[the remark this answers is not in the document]
EvanTheB: It makes C implementation harder, every other language can handle bigints fine. It means i never have to think about integer overflow which is one of the many things that make programming in python great for scripting and prototyping.

versions/development/SPEC.md
```diff
@@ -302,7 +302,37 @@ For more details on the postfix quantifiers, see the section on [Optional Parame
...
+will be converted to a string. This is straigthforward for `Int`.
+For `Float`, behaviour will be similar to the 'g' conversion specifier
+defined by the C language.
+If you have stricter requirements for `Float` formatting, you should use `String`
+and call out to your language or tool of choice for numerical manipulation.
+
+#### String Types
+
+`String` and `File` are the string types.
+`File` type can be used in place of `String` in expressions.
```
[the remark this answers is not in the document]
EvanTheB: So that this : 
```
File input ...
File output = "${input}.md5"
``` 
would be valid, but not this : 
```
File output = input + ".md5"
```

I think it makes great sense to me. Is the wordage then something like; 

```String interpolation (${}) expressions must evaluate to a non-collection type, ints floats strings bools formatted as normal blah, File is substituted to the URI depending on the engine.

Only boolean operators are defined for File (Maybe not even them?). 

Files can only be declared (=) from a string or a file. If string, it must make sense to the engine (URI or path)
```

versions/development/SPEC.md
```diff
@@ -538,20 +534,15 @@ Below are the valid results for operators on types.  Any combination not in the
...
-|`String`|`+`|`String`|`String`||
+|`String`|`+`|`String`|`String`|Concatenation|
 |`String`|`==`|`String`|`Boolean`||
 |`String`|`!=`|`String`|`Boolean`||
 |`String`|`>`|`String`|`Boolean`||
 |`String`|`>=`|`String`|`Boolean`||
 |`String`|`<`|`String`|`Boolean`||
 |`String`|`<=`|`String`|`Boolean`||
 ||`-`|`Float`|`Float`||
-||`+`|`Float`|`Float`||
```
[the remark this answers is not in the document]
EvanTheB: For literals, '+' is part of the token: `('+'|'-')?[0-9]+` .

versions/development/SPEC.md
```diff
@@ -302,7 +302,37 @@ For more details on the postfix quantifiers, see the section on [Optional Parame
...
+
+The mathematical operators are defined only for identical types.
+In expressions using an operator with mismatched numeric types,
+the `Int` will be cast to `Float` and the result will be `Float`.
+Note that this will cause loss of precision. The `Float` can be
+converted to `Int` with the `ceil`, `round` or `floor` functions if needed.
+
+In expressions involving a numeric type and a String, the numeric type
+will be converted to a string. This is straigthforward for `Int`.
+For `Float`, behaviour will be similar to the 'g' conversion specifier
```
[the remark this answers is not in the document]
EvanTheB: I was wary to say 'g' as defined by C, as I was not sure how Java and Python would be compatible with that.

versions/development/SPEC.md
```diff
@@ -302,7 +302,37 @@ For more details on the postfix quantifiers, see the section on [Optional Parame
...
+Float is used to hold a finite 64-bit IEEE-754 floating point number.
+Note that WDL itself is not intended to be used for heavy numeric computation.
+There are no number formatting utilities, many mathematical functions are absent,
+and engine behaviour at extreme values is not defined and should not be relied upon.
+
+The mathematical operators are defined only for identical types.
+In expressions using an operator with mismatched numeric types,
+the `Int` will be cast to `Float` and the result will be `Float`.
+Note that this will cause loss of precision. The `Float` can be
+converted to `Int` with the `ceil`, `round` or `floor` functions if needed.
```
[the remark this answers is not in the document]
EvanTheB: The table lacks the statement 'ints are cast to floats then they are added'. That could be added to the table. I enjoyed making the table so much shorter.
````
Int + Int
Float + Float 
Float + Int (int is cast to  float then added)
Int + Float (int is cast to float then added)
````

File: versions/development/SPEC.md [hunk 6]
```diff
@@ -302,7 +302,37 @@ For more details on the postfix quantifiers, see the section on [Optional Parame
...
+
+In expressions involving a numeric type and a String, the numeric type
+will be converted to a string. This is straigthforward for `Int`.
```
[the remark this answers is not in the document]
EvanTheB: I think this idea is great, it allows a format mini language to be defined ala python {} format strings, which could solve the gross issues of `%g` float formatting, and allow things like 000 prefixed ints. And it means the only mixed types on operators would be int and float.

rhpvorderman:
Whoops. I made a typo. I meant to say that `File + String` makes perfect sense to me. Say `File` refers to a directory that you want to place a new file named by the `String` in.
`File + File` is indeed not very intuitive.

As to the removal of `Float + Int` operations from the table. I agree with @cjllanwarne that they should stay there and that your contribution is indeed valuable for a string interpolation or preferred implementation section. It is at least nice to have all possible operations stated in a table so you can use it as a reference when writing WDL. Things will probably work as expected, and I prefer to only look up specific implementation details when things do not work as expected and errors are thrown.

versions/development/SPEC.md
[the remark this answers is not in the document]
DavyCats: I am not sure that referring to a certain behavior in a programming language at all is appropriate for the WDL specs. WDL was intended to usable by non-programmers, asking them to look things up about C, therefore, seems out of place/inappropriate.

It might be more appropriate and clearer to describe the expected behavior in full in the WDL specs themselves.

versions/development/SPEC.md
```diff
@@ -597,7 +587,6 @@ runtime {
 |            | Inequality            | left-to-right | x!=y                 |
 | 3          | Logical AND           | left-to-right | x&&y                 |
 | 2          | Logical OR            | left-to-right | x\|\|y               |
-| 1          | Assignment            | right-to-left | x=y                  |
```
DavyCats: I don't see the purpose of removing this. This might be obvious, but it is still part of the precedence order. If this does end up being removed, the precedence column should be adjusted for the other rows as well.

EvanTheB: In WDL, '=' is not an operator. It is part of the syntax of declaration. `name = expression` Most languages do not have '=' as an operator, C and C++ (Java?) being the well known exceptions.

versions/development/SPEC.md
```diff
@@ -538,20 +534,15 @@ Below are the valid results for operators on types.  Any combination not in the
...
-|`Int`|`+`|`String`|`String`||
-|`String`|`+`|`Float`|`String`||
-|`String`|`+`|`Int`|`String`||
-|`String`|`+`|`String`|`String`||
+|`String`|`+`|`String`|`String`|Concatenation|
 |`String`|`==`|`String`|`Boolean`||
 |`String`|`!=`|`String`|`Boolean`||
 |`String`|`>`|`String`|`Boolean`||
 |`String`|`>=`|`String`|`Boolean`||
 |`String`|`<`|`String`|`Boolean`||
```
EvanTheB: Which sort order is this? WDL says a string is unicode. Python default sort is not the 'official' unicode sort order IIRC.
https://www.unicode.org/reports/tr10/#Introduction

versions/development/SPEC.md
```diff
@@ -481,52 +511,18 @@ Below are the valid results for operators on types.  Any combination not in the
...
-|`Float`|`>=`|`Int`|`Boolean`||
-|`Float`|`<`|`Int`|`Boolean`||
-|`Float`|`<=`|`Int`|`Boolean`||
-|`Float`|`+`|`String`|`String`||
-|`Int`|`+`|`Float`|`Float`||
-|`Int`|`-`|`Float`|`Float`||
-|`Int`|`*`|`Float`|`Float`||
-|`Int`|`/`|`Float`|`Float`||
-|`Int`|`%`|`Float`|`Float`||
-|`Int`|`==`|`Float`|`Boolean`||
```
EvanTheB: It is probably always a mistake to be using == / != on mixed ints and floats.

versions/development/SPEC.md
DavyCats: Okay, the precedence column will still need adjustment in that case :wink:.

versions/development/SPEC.md
```diff
@@ -538,22 +526,24 @@ Below are the valid results for operators on types.  Any combination not in the
...
+|`File`|`!=`|`File`|`Boolean`||
 ||`-`|`Float`|`Float`||
-||`+`|`Float`|`Float`||
 ||`-`|`Int`|`Int`||
-||`+`|`Int`|`Int`||
 ||`!`|`Boolean`|`Boolean`||
 
+Note: In expressions using an operator with mismatched numeric types,
+the `Int` will be cast to `Float` and the result will be `Float`.
+This will could cause loss of precision. The `Float` can be
```
DavyCats: This should be "This will cause..." or "This could cause..."

versions/development/SPEC.md
```diff
@@ -538,22 +526,24 @@ Below are the valid results for operators on types.  Any combination not in the
...
-|`String`|`+`|`String`|`String`||
+|`String`|`+`|`String`|`String`|Concatenation|
 |`String`|`==`|`String`|`Boolean`||
 |`String`|`!=`|`String`|`Boolean`||
 |`String`|`>`|`String`|`Boolean`||
 |`String`|`>=`|`String`|`Boolean`||
 |`String`|`<`|`String`|`Boolean`||
 |`String`|`<=`|`String`|`Boolean`||
+|`File`|`==`|`File`|`Boolean`||
+|`File`|`!=`|`File`|`Boolean`||
```
DavyCats: The table is alphabetical, so these should be higher up, where they originally were.

versions/development/SPEC.md
```diff
@@ -1087,7 +1075,28 @@ task example {
...
+"${"abc"}" == "abc"
+
+File def = "hij"
+"${def}" == "hij"
+
+"${5}" == "5"
+
+"${3.141}" == "3.141000"
+"${3.141 * 1E-10}" == "0.000000"
+"${3.141 * 1E10}" == "31410000000.000000"
```
EvanTheB: Actually 1E10 syntax is not allowed for floats.

versions/development/SPEC.md
```diff
@@ -302,7 +302,13 @@ For more details on the postfix quantifiers, see the section on [Optional Parame
 
 For more information on type and how they are used to construct commands and define outputs of tasks, see the [Data Types & Serialization](#data-types--serialization) section.
 
-#### Custom  Types
+#### Numeric Behaviour
```
geoffjentry: Just for consistency sake, `Behavior`

versions/development/SPEC.md
```diff
@@ -302,7 +302,13 @@ For more details on the postfix quantifiers, see the section on [Optional Parame
...
+
+`Int` and `Float` are the numeric types.
+`Int` can be used to hold any integer, there is no size limit.
```
geoffjentry: None? 

Imagine in the Java world, even using `BigInteger` you're still limited by the maximum size of a `String`.

EvanTheB: I mean, none of our languages are turing complete for the same reason. `If the size would be unrepresentable by the machine, the program must crash` - but this kind of behaviour is assumed. I guess I am trying to say: authors of wdl, do not worry about that machine. Authors of engine, do not make your implementation overly limited.

geoffjentry: TBH based on what your'e trying to convey, I'd just either stick with keeping it simple and declaring `Int` to be an integer type or go with something fairly standard and declaring it to be a 64 bit integer (i.e. what most ppl would call a long)

geoffjentry:
This PR has passed and is waiting for implementation

PR #236, opened 2018-07-27
Add singularity to spec

rhpvorderman:
Since docker is already in the specification, why not add [Singularity](https://www.sylabs.io/)? 
[Singularity can run the same docker containers, but without root](https://www.sylabs.io/guides/2.5.1/user-guide/singularity_and_docker.html). On our cluster root is reserved for admins only. Scientists do not have root access (and no docker access as a result). 
WDL being developed for use by the scientific community I think it is even a better companion to WDL than docker is. 
And since you can use the same docker containers no functionality is lost, with the gain of not requiring root access.

### DIFF
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -1152,6 +1152,25 @@ task docker_test {
 }
 ```
 
+#### singularity
+
+Location of a Singularity image for which this task ought to be run. This can have a format like `patrickvdb/fastqc-singularity`,
+`shub://patrickvdb/fastqc-singularity:latest` or `docker://biocontainers/fastqc:latest`.
+
+```wdl
+task docker_test {
+  input {
+    String arg
+  }
+  command {
+    python process.py ${arg}
+  }
+  runtime {
+    singularity: "docker://ubuntu:latest"
+  }
+}
+```
+
 #### memory
 
 Memory requirements for this task.  Two kinds of values are supported for this attributes:


geoffjentry:
Hi @rhpvorderman thanks for this. I've been interested in singularity support for quite some time, but have always dodged it as I feel like getting the syntax correct will be tricky.

To use the examples you provide in this PR - why would this WDL be **requiring** singularity to be used on `docker://biocontainers/fastqc:latest` - if you write this WDL and then publish it, and I'm running on e.g. Azure, what if I wanted to use actual Docker for that?

What about instead changing the `docker` field in `runtime` to simply say `container` and rqequire the combination of engine smarts & engine-level parameters (e.g. `workflow options` and/or configuration  in Cromwell) to work out the correct technology to use for the container?

To provide pushback to my own suggestion, I'm not certain that would always make sense, but I'd much prefer a world where `runtime` wasn't being so prescriptive on this stuff.

sooheelee:
Two votes for Singularity support. Two GATK workshop hosts said they'd be using Singularity or are already using Singularity.

rhpvorderman:
Hi @geoffjentry thanks for your feedback. Using `container:` makes much more sense than using either `docker:` or `singularity:`. This is a very good suggestion. I will make a separate pull request for this, as it is different from the topic of this PR.

rhpvorderman:
I will leave this open in case #237 is rejected. But I personally feel that renaming docker to container is much, much better. Thans @geoffjentry for this excellent suggestion.

geoffjentry:
Hi @rhpvorderman - Cool! I'm actually going to close this just to avoid confusion, we can always re-open it if necessary. Just don't let me forget if that comes to pass :)

PR #237, opened 2018-07-30
replace docker runtime attribute with more generic container attribute

rhpvorderman:
Thanks to @geoffjentry for this excellent suggestion in #236 . 

`docker` is an implementation of a technology called `containers`. Therefore this PR proposes to ditch the runtime attribute `docker:` and replace it with the more generic and more descriptive `container:`. 
This allows for much easier choice of container engines in the execution engine of wdl.

```WDL
task hello {
  input {
    String addressee
  }
  command {
    echo "Hello ${addressee}!"
  }
  runtime {
      container: "docker://ubuntu:latest"  # Instead of docker: ubuntu:latest
  }
  output {
    String salutation = read_string(stdout())
  }
}
```

I used the singularity way of defining image locations (i.e. `docker://` or `shub://`) because it is easy to understand and can be parsed relatively easily by a execution engine.  Also this allows for having a settings file like this in the execution engine:
```HOCON
container_runtimes {
  default: docker
  shub: singularity
  docker: singularity
  foo: bar_container_solution
}
```
With in the above case choosing singularity for images that are explicitly located on the docker hub and choosing docker for images that have no location specified. 

This PR will decouple WDL from docker and make it more useful for use with other container engines, while still allowing a lot of flexibility. Looking forward to your feedback!

### DIFF
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -778,7 +778,7 @@ Some examples of correct import resolution:
 
 ## Task Definition
 
-A task is a declarative construct with a focus on constructing a command from a template.  The command specification is interpreted in an engine and backend agnostic way. The command is a UNIX bash command line which will be run (ideally in a container (such as a Docker image)).
+A task is a declarative construct with a focus on constructing a command from a template.  The command specification is interpreted in an engine and backend agnostic way. The command is a UNIX bash command line which will be run (ideally in a container such as a Docker or Singularity container).
 
 Tasks explicitly define their inputs and outputs which is essential for building dependencies between tasks.
 
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -1163,7 +1163,7 @@ task container_test {
 
 Since there are multiple container solutions out in the wild it is best to make the location of the image clear. In case the image should be pulled from DockerHub `docker://ubuntu:latest` should be used. 
 In case it should be pulled from the Singularity hub `shub://ubuntu:latest` should be used. 
-In case a container solution named `foo` comes along `foo://bar:baz` should be used. Since `docker://` and `shub://` images can both be pulled by singularity, the execution engine should allow 
+In case a container solution named `foo` comes along with its own FooHub `foo://ubuntu:latest` should be used. Since `docker://` and `shub://` images can both be pulled by singularity, the execution engine should allow 
 to set which prefix (`docker://`, `shub://` or `foo://`) should be matched to which container solution.
 ```
 
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -45,7 +45,7 @@
         * [Task portability and non-standard BaSH](#task-portability-and-non-standard-bash)
     * [String Interpolation](#string-interpolation)
     * [Runtime Section](#runtime-section)
-      * [docker](#docker)
+      * [container](#container)
       * [memory](#memory)
     * [Parameter Metadata Section](#parameter-metadata-section)
     * [Metadata Section](#metadata-section)
@@ -179,7 +179,7 @@ task hello {
   }
 
   runtime {
-    docker: "broadinstitute/my_image"
+    container: "docker://dbroadinstitute/my_image"
   }
 
   output {
@@ -778,7 +778,7 @@ Some examples of correct import resolution:
 
 ## Task Definition
 
-A task is a declarative construct with a focus on constructing a command from a template.  The command specification is interpreted in an engine and backend agnostic way. The command is a UNIX bash command line which will be run (ideally in a Docker image).
+A task is a declarative construct with a focus on constructing a command from a template.  The command specification is interpreted in an engine and backend agnostic way. The command is a UNIX bash command line which will be run (ideally in a container (such as a Docker image)).
 
 Tasks explicitly define their inputs and outputs which is essential for building dependencies between tasks.
 
@@ -1059,7 +1059,7 @@ output {
 }
 ```
 
-The array of `File`s returned is the set of files found by the bash expansion of the glob string relative to the task's execution directory and in the same order. It's evaluated in the context of the bash shell installed in the docker image running the task.
+The array of `File`s returned is the set of files found by the bash expansion of the glob string relative to the task's execution directory and in the same order. It's evaluated in the context of the bash shell installed in the container running the task.
 
 In other words, you might think of `glob()` as finding all of the files (but not the directories) in the same order as would be matched by running `echo <glob>` in bash from the task's execution directory.
 
@@ -1076,9 +1076,9 @@ Then running `echo a*` in the execution directory would expand to `a1.txt`, `ab.
 
 ##### Task portability and non-standard BaSH
 
-Note that some specialized docker images may include a non-standard bash shell which supports more complex glob strings. These complex glob strings might allow expansions which include `a_inner.txt` in the example above.
+Note that some specialized container images may include a non-standard bash shell which supports more complex glob strings. These complex glob strings might allow expansions which include `a_inner.txt` in the example above.
 
-Therefore to ensure that a WDL is portable when using `glob()`, a docker image should be provided and the WDL author should remember that `glob()` results depend on coordination with the bash implementation installed on that docker image.
+Therefore to ensure that a WDL is portable when using `glob()`, a container image (i.e. singularity or docker) should be provided and the WDL author should remember that `glob()` results depend on coordination with the bash implementation installed on that container image.
 
 ### String Interpolation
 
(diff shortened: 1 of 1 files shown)


versions/development/SPEC.md
```diff
@@ -778,7 +778,7 @@ Some examples of correct import resolution:
 
 ## Task Definition
 
-A task is a declarative construct with a focus on constructing a command from a template.  The command specification is interpreted in an engine and backend agnostic way. The command is a UNIX bash command line which will be run (ideally in a Docker image).
+A task is a declarative construct with a focus on constructing a command from a template.  The command specification is interpreted in an engine and backend agnostic way. The command is a UNIX bash command line which will be run (ideally in a container (such as a Docker image)).
```
cjllanwarne: Not a big fan of the nested parens here...

versions/development/SPEC.md
```diff
@@ -179,7 +179,7 @@ task hello {
   }
 
   runtime {
-    docker: "broadinstitute/my_image"
+    container: "docker://dbroadinstitute/my_image"
```
cjllanwarne: Just a thought, maybe "encoding into a single string" is not the nicest way to do this any more, especially if if different technologies might use different formats for their strings?

Eg why not have a standard map format like:
```wdl
runtime {
 container: {
  "organization": "broadinstitute"
  "image_name": "myimage"
  "floating_tag": "latest"
  "technologies": ["singularity", "docker", "foocontainertech"]
 }
}
```

The nice thing about this is that the WDL author doesn't need to worry about different string formatting styles in docker, singularity and foocontaintertech (which uses `\` instead of `/`).

geoffjentry: I get the sentiment here but do you really see people willingly typing all of that out?

EvanTheB:
Re: 
```
container_runtimes {
  default: docker
  shub: singularity
  docker: singularity
  foo: bar_container_solution
}
```

It seems like this won't be enough, as the different tools expect the value of `container` to be pre-processed by the engine differently. From `docker://ubuntu, docker expects `ubuntu`, and singularity expects `docker://ubuntu`. Not the worst for 2 options, but expecting to be future proof for 'foo' is unrealistic. I don't have any ideas for fixing this. Maybe just remove mention of foo.

versions/development/SPEC.md
rhpvorderman: Isn't that something that can be handled by the execution engine given the single string format gives all the necessary information?
Furthermore a container string should always refer to a *single* image. Otherwise it is not reproducible. Given that it should refer to a single image, a single identifier string should be sufficient. Also more convenient. But I agree how this single string should like is still up for debate.

versions/development/SPEC.md
rhpvorderman: Ugly indeed. Fixed.

patmagee:
Tbh I like the idea of switching to using the term `container` but anything more then that really should be engine functionality. The wdl itself should not necessarily have the concept of docker ot singularity or vagrant etc.

rhpvorderman:
I partly agree with @patmagee. I agree fully that WDL should not have a concept of any container implementations.
I disagree that how container images should be defined should be left up to the execution engine. This will lead to changing implementations in execution engines, making WDL files not portable anymore. How a container image should be defined should be a part of the spec.

Thanks to @patmagee I am starting to think that putting this all in one string was a bad idea. But how about this:

```WDL
task example {
  inputs {
    String someInput
  }
  command {
    echo ~{someInput}
  }
  runtime {
    container: broadinstitute/echo:43
    container_engine: docker
  }
}
```

Whatever is in container does not matter, this will be handled by the container engine. The execution engine can determine a default container engine when no container_engine is supplied. But it is better to enforce container_engine when container is defined to enforce reproducibility.

In this way:
* Container image string does not have to be parsed by the execution engine. It can be in native format of whatever container engine is used.
* It is immediately apparent to a reader of the wdl what the container engine is and where the image can be found.
* Portability and reproducibility ensured in one fell swoop!

I am curious what you all think about this. Any thoughts @geoffjentry, @cjllanwarne, @EvanTheB  and @patmagee?

geoffjentry:
@rhpvorderman As I've dug into Singularity support for Cromwell (Thanks @vsoch!) how I view this entire problem has changed a fair amount.

There are really two independent things here which are masquerading as one:

- What container "driver" (for lack of a better word) to use for a standard Docker container. For instance, one might want to use normal `docker` but one might want to use `udocker` or `singularity`. I **don't** think that this should be specified in the WDL at all because that's going to vary from user to user. For instance if I take a workflow and run it on Google Pipelines API I definitely want to be using `docker` but I'll want to take that same WDL and run it on my HPC set up where we use `udocker`. In Cromwell we've started addressing that via Cromwell configuration since it's nearly always the case that those are backend-wide (i.e. not user choice) decisions.
- The case where the container itself is a specialized container type, i.e. a native Singularity container. This obviously is trickier.

Almost always, when I hear people say that they want to use an alternative container tech in a workflow they're referring to the first issue, which means that since that's not even a WDL problem we can all ignore it (well, until I put my Cromwell hat back on ....).

The second one is trickier. Singularity is the only alternative container tech I've seen people request which has its own native format. I don't have a great suggestion at the moment but we should be cognizant that this is by far the minority use case in this discussion as almost always people mean the first bullet point.

patmagee:
@DavyCats That kind of falls under the `hints` idea I guess. where we use the `hints` to define what the "preferred" thing to use is.  Siimilar to your suggestion,I wonder if instead of adding a separate field we rely on the engine to define this and use a driver string at the start of the docekr string. ie:

```wdl
runtime {
  container: "docker:broadinstitaute/echo:43"
}
```
or 
```wdl
runtime {
   container: "singularity:broadinstitute/echo:43"
}
```

In this model, the `driver` is defined directly in the container string, but at the same time it's up to the engine to properly resolve it. The engine would additonally define the  default behaviour if there is no `driver` defined in the container string. Furthermroe, if the engine does not support a `driver` I would if it would fail fast or try to default to the supported engines. Maybe that is an implementation detail we do not need to define

DavyCats:
@patmagee I'm not sure that it should be considered a hint on what engine to use, rather it's just an indication of what engines could be used (if the format is docker, it could use docker, udocker, singularity, etc., but if it's singularity it can only(?) be singularity). It doesn't give any indication of which engine is preferred, but just what format the container is in.

vsoch:
Two quick comments - I think we shouldn't forget that all of these image formats, binary or not, come down to compressed things that (in most cases) can be [uncompressed](https://github.com/vsoch/singularity2docker). The new singularity image format is a variation of squashfs (with a special header) that can also be unsquashed. Docker layers are just .tar.gz, and can be dumped around into different container runtimes, as you know.

I think of this a bit differently I think. A singularity container is a binary, it's more like a command that I might want to run, and maybe even within an orchestration system that uses docker. If you try and "hard code" these particular backends in an effort to abstract away complexity and make it easier for the user, you may in fact be making it harder by using the technology is a very specific (limited?) way.  You also have to be careful with inventing new specifications because it would be pretty confusing to see either of these:

```bash
runtime {
   container: "docker:broadinstitute/echo:43",
   command: "singularity run docker://broadinstitute/echo:43"
}
```
```bash
runtime {
   container: "singularity:docker://broadinstitute/echo:43",
}
```
The reason that singularity is easy to use is because you treat a container just like a binary. You move it around and you run it, and that's why it's flexible to run easily in technologies that already exist. A lot of the other "containers for HPC" have additional dependencies (gateways, storages, wrappers, etc.) and because it's not transparent that you just run it like a command, it's much harder. So we can apply this same logic to something like cromwell - cromwell should be optimized to understand a small set of things (inputs, outputs, executables), and do that well, and expect for those things to change over time (and still work because they aren't hard coded into it).

patmagee:
My hesitation around adding WDL support things like singularity is that at the moment it is unclear how many people are actually going to need? Is it enough to simply state that the Backend which supports singularity can define this within its own configuration? At the moment, most of the labs and clients that I work worth are still just starting to move towards docker and have not as of yet requested singularity.

This is not to say of course that we should not implement support for singularity. But I think we should question whether or not we should change the WDL spec in a way that might obfuscate the standard (docker) use case without a significant demand for it

vsoch:
I would say as long as singularity doesn’t have the same kind of support for instances that Docker has for managing services, singularity should be treated as an executable and just go under the command section. And in this framework we would provide just guidance about how to specify a singularity backend in the configuration examples file.  When singularity is further developed to have better communication between instances and similar functionality to Docker, then this conversation can be started up again to discuss some kind of further integration. But imho it’s not ready for that yet.

rhpvorderman:
Given the discussion above. I think it is still better to replace `docker` with `container` but just to leave it at that. How the container string should be parsed, and with what container engine it is used should be left to the execution engine and ideally be user configurable.
Then all what is required is just a word change in the spec.

geoffjentry:
Honestly the more I think about it the more my $0.02 would be to just stick with `docker` for now. The extreme majority of use cases we've seen from the Cromwell side have been people trying to use docker containers via a more HPC-friendly technology and that knowledge needs to be instance level (e.g. a specific Cromwell instantiation) and not at the WDL level

rhpvorderman:
> Honestly the more I think about it the more my $0.02 would be to just stick with docker for now. The extreme majority of use cases we've seen from the Cromwell side have been people trying to use docker containers via a more HPC-friendly technology and that knowledge needs to be instance level (e.g. a specific Cromwell instantiation) and not at the WDL level.

I think these is a very good reason to drop the PR and close the thread. 

Everyone thanks a lot for your input!

PR #238, opened 2018-08-01
[1.0/development grammar.hgr]: Allow empty strings as static strings

cjllanwarne:


geoffjentry:
I don't think we should be retroactively changing things, even when they're incorrect.

versions/development/parsers/grammar.hgr
```diff
@@ -108,6 +108,7 @@ grammar {
 
     mode<awaiting_version_name> {
       r'1.1-draft' -> :version_name %pop
+      r'biscayne' -> :version_name %pop
```
geoffjentry: I don't think that `biscayne` should be in this at all, unless you'd like to propose that as the official name for `not-yet-1.1`

EvanTheB: I'm happier with 'biscayne' than 1.1-draft. biscayne forever in /dev, real numbers in /!dev

cjllanwarne: I'm happy to advocate for something other than 1.1 (since who knows what the version number will be if any of the breaking changes make it in)

geoffjentry: It's more that at the moment `biscayne` is meaningless in WDL-world

I agree that `1.1` is not the right value.

cjllanwarne:
I like with "don't change things retrospectively" but in this case I'm not happy with tying our hands and being unable to address silly typos and bugs *at all* until the next full version.

What if I split these grammar change off as a 1.0.1 version - no real changes to 1.0 allowed, only typos and bug fix changes allowed?

versions/development/parsers/grammar.hgr
patmagee: I dont really know what biscayne is.. afaik it's a cromwell thing. I do not think it belongs in this pr. I would probably move to a patch version in this instance rather then introduce some weird code name that makes  no sense

geoffjentry:
This is just a bug fix on the grammar file, merging

PR #240, opened 2018-08-02
[1.0/development grammar.hgr]: Allow values called 'version'

### LINKED ISSUE #239: [1.0/development grammar.hgr]:  Allow 'version' as a variable name
Right now the word `version` is treated like a keyword throughout a WDL 1.0 (and development) file by the hermes grammar.

We should indeed treat `version 1.0` at the top of the file specially, but elsewhere using `String version` as an input for example, it should be able to validate just fine:

```wdl
version 1.0

task detect_java_version {
  command <<< ... >>>
  output {
    # This 'version' should NOT be considered a keyword:
    String version = read_string(stdout())
  }
```

cjllanwarne:
Fixed #239 

The PR changes probably look more dramatic than they are. What this does is:

- Move the previous "default" lexing into a new mode called 'main'
- The new 'default' lexing mode allows only whitespace, comments and the `:version` token.
- The new main mode no longer recognized the string `version` token as anything special, allowing it to be used in value names and expressions again

---

As in:

```wdl
version 1.0

workflow version_variables {
  input {
    String version = "hi"
  }
  call foo { input: version = version}
}

task foo {
  input {
    String version
  }
  command { ... }
  output {
    String version_out = ""
  }
}
```

### DIFF
--- a/versions/1.0/parsers/grammar.hgr
+++ b/versions/1.0/parsers/grammar.hgr
@@ -7,79 +7,84 @@ grammar {
       r'{%_identifier_start%}({%_identifier_follow%})*' -> _identifier
     }
 
-    r'\s+' -> null
-
-    enum {
-      python: r'/\*(.*?)\*/' (DOTALL)
-      c: r'/\*(.*?)\*/' (PCRE_DOTALL)
-      java: r'/\*(.*?)\*/' (DOTALL)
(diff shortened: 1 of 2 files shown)


geoffjentry:
Same comments I made on the other PR re `1.0` and `biscayne`

geoffjentry:
Merging this as it's just a bug fix to the grammar

PR #241, opened 2018-08-06
WDL Directory Type (redux)

cjllanwarne:
Apologies for the long wait between me saying "I want to prototype this before going any further with the PR" and actually doing so.

This is an updated version of PR #173 which moves the Directory changes to the development version of the SPEC and includes the grammar changes too.

I can also link to a real, working prototype in https://github.com/broadinstitute/cromwell/pull/3980 that I'd encourage people to try out and see how this would work for real, if adopted.

### DIFF
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -280,6 +280,7 @@ Float f = 27.3             # A floating point number
 Boolean b = true           # A boolean true/false
 String s = "hello, world"  # A string value
 File f = "path/to/file"    # A file
+Directory d = "/path/to/"  # The contents of a directory (including sub-directories)
 ```
 
 In addition, the following compound types can be constructed, parameterized by other types. In the examples below `P` represents any of the primitive types above, and `X` and `Y` represent any valid type (even nested compound types):
@@ -803,6 +804,7 @@ task t {
   input {
     Int i
     File f
+    Directory d
   }
 
   # [... other task sections]
@@ -810,13 +812,19 @@ task t {
 ```
 
 #### Task Input Localization
-`File` inputs must be treated specially since they require localization to within the execution directory:
-- Files are localized into the execution directory prior to the task execution commencing.
+`File` and `Directory` inputs must be treated specially since they require localization to within the execution directory:
+- Files and directories are localized into the execution directory prior to the task execution commencing. 
 - When localizing a `File`, the engine may choose to place the file wherever it likes so long as it accords to these rules:
   - The original file name must be preserved even if the path to it has changed.
   - Two input files with the same name must be located separately, to avoid name collision.
   - Two input files which originated in the same storage directory must also be localized into the same directory for task execution (see the special case handling for Versioning Filesystems below).
+- When localizing a `Directory`, the engine follows the same set of rules as for Files:
+  - The original directory name must be preserved even if the path to it has changed.
+  - Two input directories with the same name must be located separately, to avoid name collision.
+  - Two input directories which originated in the same parent storage directory must be localized into the same parent directory for task execution.
+  - The internal structure of files and sub-directories in the directory is preserved within the new directory location.
 - When a WDL author uses a `File` input in their [Command Section](#command-section), the fully qualified, localized path to the file is substituted into the command string.
+- When a WDL author uses a `Directory` input in their [Command Section](#command-section), the fully qualified localized path to the directory (with no trailing '/') is substituted into the command string.
 
 ##### Special Case: Versioning Filesystems
 Two or more versions of a file in a versioning filesystem might have the same name and come from the same directory. In that case the following special procedure must be used to avoid collision:
@@ -1044,6 +1052,43 @@ output {
   String a = "a"
   String ab = a + "b"
 }
+```
+
+#### Directory Outputs
+
+Directories can be declared as outputs just like files. For example:
+```
+output {
+  Directory d = "created/directory/"
+}
+```
+
+Notes:
+
+ - When a directory is output, the contents (including subdirectories) are considered part of the output but the path is not. 
+ - When a directory is output, its contents *do not overwrite* any pre-existing directory content. For example:
+
+```wdl
+workflow directory_manipulation {
+  # Make a directory:
+  call make_directory
+  Directory d = make_directory.made_directory
+  
+  # List the contents before the "update_directory" task:
+  call list_contents as list_contents_before { input: d = d }
+  String d_contents_before = list_contents_before.list
+
+  # An update task:
+  call update_directory { input: d = d }
+  
+  # List the contents again:
+  call list_contents as list_contents_after { input: d = d }
+  String d1_contents_after = list_contents_after.list
+
+  # No matter what update directory does, the following is *always* true:
+  Boolean match = d1_contents_before == d1_contents_after
+}
+
 ```
 
 #### Globs
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -272,7 +272,9 @@ $float = (([0-9]+)?\.([0-9]+)|[0-9]+\.|[0-9]+)([eE][-+]?[0-9]+)?
 
 ### Types
 
-In WDL *all* types represent immutable values. Even types like `File` and `Directory` represent snapshots of a file or directory at the time when the value is created
+In WDL *all* types represent immutable values. 
+  - Even types like `File` and `Directory` represent logical "snapshots" of the file or directory at the time when the value was created.
+  - It's impossible for a task to change an upstream value which has been provided as an input: even if it makes changes to its local copy the original value is unaffected.
 
 All inputs and outputs must be typed. The following primitive types exist in WDL:
 
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -1086,6 +1086,29 @@ Notes:
  - When a directory is output, the name and contents (including subdirectories) are considered part of the output but the path is not.
  - Any hard- or soft- links within the execution directory shall be resolved into separate, regular files in the WDL `Directory` value produced as the output.
 
+##### Soft link resolution example
+
+For example imagine a task which produces:
+```
+dir/
+ - a           # a file, 10 MB
+ - b -> a      # a softlink to 'a'
+```
+
+As a WDL directory this would manifest as:
+```
+dir/
+ - a           # a file, 10 MB
+ - b           # another file, 10 MB
+```
+
+And therefore if used as an input to a subsequent task:
+```
+dir/
+ - a           # a file, 10 MB
+ - b           # another file, 10 MB
+```
+
 #### Globs
 
 Globs can be used to define outputs which might contain zero, one, or many files. The glob function therefore returns an array of File outputs:
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -272,6 +272,8 @@ $float = (([0-9]+)?\.([0-9]+)|[0-9]+\.|[0-9]+)([eE][-+]?[0-9]+)?
 
 ### Types
 
+In WDL *all* types represent immutable values. Even types like `File` and `Directory` represent snapshots of a file or directory at the time when the value is created
+
 All inputs and outputs must be typed. The following primitive types exist in WDL:
 
 ```wdl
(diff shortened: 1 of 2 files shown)


versions/development/SPEC.md
```diff
@@ -1049,6 +1057,43 @@ output {
...
+  
+  # List the contents before the "update_directory" task:
+  call list_contents as list_contents_before { input: d = d }
+  String d_contents_before = list_contents_before.list
+
+  # An update task:
+  call update_directory { input: d = d }
+  
+  # List the contents again:
+  call list_contents as list_contents_after { input: d = d }
```
cjllanwarne: Actually this example isn't great - since this `call` doesn't depend on any outputs of `update_directory`, it would all just run in parallel 🤦‍♂️

versions/development/SPEC.md
```diff
@@ -1049,6 +1057,43 @@ output {
...
+Directories can be declared as outputs just like files. For example:
+```
+output {
+  Directory d = "created/directory/"
+}
+```
+
+Notes:
+
+ - When a directory is output, the contents (including subdirectories) are considered part of the output but the path is not.
```
EvanTheB: There seems to be a minor inconsistency between input and output.
```input: Directory = "/home/evan/bin" -> ".../workflow_wd/bin/..."
output: Directory = "bin" -> ".../workflow_output/?/..."```

I think I may be misreading. Is the intent that the name of the directory is lost (bin), or only the 'path' to the directory that contains the directory? It is confusing. I think the intent is the same behaviour as for Files, the name is kept, and the 'contents', but not anything else: path-to, permissions, whether it is a link or not etc. As it is not specified for Files, do we need to specify it for directories? or do we also need to specify it for Files?

File: versions/development/SPEC.md [hunk 3]
```diff
@@ -1049,6 +1057,43 @@ output {
...
+```
+output {
+  Directory d = "created/directory/"
+}
+```
+
+Notes:
+
+ - When a directory is output, the contents (including subdirectories) are considered part of the output but the path is not. 
+ - When a directory is output, its contents *do not overwrite* any pre-existing directory content. For example:
```
EvanTheB: Similar behaviour is not specified for Files, and cromwell uses that to allow different localisation strategies; soft-link, hard-link, copy. Depending on the cromwell settings, modifying an input to a task may modify the original File, and an Output File could be the same as an Input (hard link to the same data).

I am interested in the reasoning for Directories being different to Files.

versions/development/SPEC.md
```diff
@@ -806,20 +807,27 @@ task t {
...
-- Files are localized into the execution directory prior to the task execution commencing.
+`File` and `Directory` inputs must be treated specially since they require localization to within the execution directory:
+- Files and directories are localized into the execution directory prior to the task execution commencing. 
 - When localizing a `File`, the engine may choose to place the file wherever it likes so long as it accords to these rules:
   - The original file name must be preserved even if the path to it has changed.
   - Two input files with the same name must be located separately, to avoid name collision.
   - Two input files which originated in the same storage directory must also be localized into the same directory for task execution (see the special case handling for Versioning Filesystems below).
+- When localizing a `Directory`, the engine follows the same set of rules as for Files:
+  - The original directory name must be preserved even if the path to it has changed.
+  - Two input directories with the same name must be located separately, to avoid name collision.
```
LeeTL1220: Is there any guidance as to how name collision is resolved?

versions/development/SPEC.md
```diff
@@ -1049,6 +1057,43 @@ output {
 }
 ```
 
+#### Directory Outputs
+
+Directories can be declared as outputs just like files. For example:
+```
+output {
+  Directory d = "created/directory/"
```
LeeTL1220: So if I my workflow/task produces `created/directory` and `created/directory2`, only `created/directory` would appear.  If I wanted both, I would make my output `created/`, correct?

LeeTL1220:
Maybe for the next pull request...

- What about `size(d)`?  We will need a way to get the total space consumption of directories.
- Can we have optional directories?  This is not a request, per se, I just noticed that it was not in the PR

versions/development/SPEC.md
cjllanwarne: The entire execution model of WDL relies on all values being immutable one created. That assumption probably should be explicit in the spec if it isn't already (for *all* value types) - it's **the** core, underlying assumption used for task scheduling and cloud portability of the same WDL to different backends.

The localization strategies in Cromwell should really be thought of as an optimization to allow you to avoid the file-copying overheads, and *not* as a way to circumvent the immutable value assumptions of WDL.

versions/development/SPEC.md
cjllanwarne: if by "appear" you mean "would be saved as an output", then yes, correct!

versions/development/SPEC.md
cjllanwarne: I think this is right - you want the directory name and everything inside it. Thanks for pointing this out, I'll try to reword things

versions/development/SPEC.md
EvanTheB: I think we agree, either; this should also be stated explicitly for Files, it should be stated explicitly once as a global, general rule for WDL, or it should not be stated here alone as that implies it is only for Directory.

cjllanwarne:
@geoffjentry as input and output values, this is intended to work in exactly the same was as CWL's directories. The "contrast" will hopefully mostly be "CWL gives you a ton more options". One example (at least at first) none of the "metadata" fields that CWL expects an engine to find are accessible - as @LeeTL1220 pointed out, a way to get the `size` would probably be nice at some point!

  - In command string interpolations, we're writing what CWL calls `path` (not location)
  - Outside of command strings, this PR isn't adding anything by way of "interactions" with directories.

versions/development/SPEC.md
patmagee: IMO no two input / output directories (except for the case of returning an unchanged input directory) should really ever be the same Object in the file system that is storing the inputs /outputs. 

Alot of people like the idea of mounting a single directory and making modifications to it and then those modifications magically become present in another jobs workspace. While I can see the draw from this (especially when working in a local computing environment) it is both fragile, and destined to cause more problems then its worth. Once A directory is created it 100% should be an immutable copy. Any modificaitons would result in a NEW directory

versions/development/SPEC.md
```diff
@@ -1049,6 +1057,43 @@ output {
...
+
+  # An update task:
+  call update_directory { input: d = d }
+  
+  # List the contents again:
+  call list_contents as list_contents_after { input: d = d }
+  String d1_contents_after = list_contents_after.list
+
+  # No matter what update directory does, the following is *always* true:
+  Boolean match = d1_contents_before == d1_contents_after
```
patmagee: Directory equalities represent a bit of an issue and I am not 100% certain they really should be implemented. They really will depend on the engine Implementation. IE if the engine is not packaging the directories into any format like a `tar` ball or `zip` folder, do we expect it to recursively dive into the directory and compute md5 sums on all of the content?

EvanTheB: This is a string equality though.

versions/development/SPEC.md
antonkulaga: @patmagee I do not think you should force it on Cromwell level, better just put warnings. Sometimes bioinformatic tools are so buggy that you have to do very crazy tricks to make them work

orodeh:
I think this is good. Here are a few points:  
1) Directories should be immutable (same as files). If a task localizes directory `D`, modifies `D`, and outputs `D`, the result is a new directory `D'`. 
2)  Watch out for coercions of the directory type to string and file. I would rather both were disallowed in workflow contexts. 
3) I believe a WDL `Directory` corresponds to a Unix directory without softlinks, device special files, fifos, etc. Only subdirectories and regular files are permitted.   

👍

versions/development/SPEC.md
patmagee: @EvanTheB Ah, I see., I interpreted this as Directory Equality

patmagee:
Thanks, @orodeh for those points. I was trying to come up with the exact same list but you captured it pretty clearly.

👍

versions/development/SPEC.md
geoffjentry: @cjllanwarne Is the trailing `/` purely superfluous here?

versions/development/SPEC.md
geoffjentry: @antonkulaga We're not discussing the Cromwell level, this is the WDL specification

patmagee: @antonkulaga putting wgat implementers would need to do aside to get this to work, I still think directories MUST be immutable constructs. Once a directory is created there should be no change that does not result in a new directory

versions/development/SPEC.md
cjllanwarne: hmm, this example is turning out to be problematic in a number of ways (see my first comment), so I might just remove it.

The intention was to have this demonstrate that the directories were immutable - meaning that they are still identical even after a task "changes" one - but I agree that that's going to be really hard to actually implement in a reasonably efficient way. 

So maybe for this first implementation we disallow equality checks and maybe revisit this if there's a huge clamor for it?

patmagee: @cjllanwarne it makes sense to me to disallow equality checks for directories. To me, it is not that informative to know that some content in the directory changed (this might be different for other people), since that change does not really reflect a meaningful state necessarily.

geoffjentry: Regarding equality, it looks like the spec doesn't say what `File == File` is comparing but the intention was certainly path equality and not content equality. Thus the same should hold for Directory types.

Whether or not these operators should exist at all is another question, but File and Directory should at least be managed the same

geoffjentry:
@orodeh Why would a softlink not be allowed, if that path eventually leads to a Unix directory structure?

orodeh:
Softlinks are well defined and work well on Unix filesystems. They have no equivalent in Windows or Amazon S3.  The problem arises if the directory `D` includes a softlink, if you localize it on a windows machine, how do you represent the softlink? If you delocalize it to a cloud object-store, say S3, how is the softlink represented? 

I don't know if Azure or GCP have softlinks, but I suspect they don't.

geoffjentry:
@orodeh My point is that it if the original system has softlinks that doesn't mean they need to still be softlinks on the target system. This is true both for task inputs as well as task outputs -> e.g. what if a tool generates a directory structure containing symlinks and that directory structure is an output of the task via a `Directory`? I don't think we can say "Sorry, those don't get delocalized".

It might be that I misunderstood your comment and that's precisely what you're saying.

orodeh:
@geoffjentry An example may illustrate the problem I am seeing. 

1. Directory `D` is local to a worker machine running task `T1`. The directory includes a softlink.
2. Task `T1` ends, and `D` is delocalized to the cloud. 
3. Task `T2` starts running, and `D`  is an input. `D` is localized to the `T2` machine. 

What happens to the original softlink?

geoffjentry:
@orodeh It'd lose its softlink-ness when pushed to the cloud, but the underlying path would still get copied to the cloud and to the T2 machine

orodeh:
@geoffjentry However, `T2` is expecting to see the exact same `D` that `T1` saw. But it won't, because the softlink has been replaced with something else. The directory has changed during the manipulations, and it is visible to standard bash scripts.

orodeh:
@geoffjentry This scenario was worrisome to me. My suggestion is to avoid it, by limiting the `Directory` type to only deal with subdirectories and regular files.

geoffjentry:
@orodeh that's what i mean. Any solution which explicitly forbids delocalization of symlinks I think is going to lead to outrage downstream. I think you **could** get away with something which provides a well defined scheme for de-symlink-ifying those files.

orodeh:
I think this it is a good practical solution. I suggest documenting it well, so as not to catch users by surprise.

geoffjentry:
@cjllanwarne Right, that's what I was picturing

geoffjentry:
@cjllanwarne also, agree with @orodeh that it'd be good to make this clear - perhaps even using the example you had above or something similar?

versions/development/SPEC.md
```diff
@@ -803,20 +808,27 @@ task t {
...
+`File` and `Directory` inputs must be treated specially since they require localization to within the execution directory:
+- Files and directories are localized into the execution directory prior to the task execution commencing. 
 - When localizing a `File`, the engine may choose to place the file wherever it likes so long as it accords to these rules:
   - The original file name must be preserved even if the path to it has changed.
   - Two input files with the same name must be located separately, to avoid name collision.
   - Two input files which originated in the same storage directory must also be localized into the same directory for task execution (see the special case handling for Versioning Filesystems below).
+- When localizing a `Directory`, the engine follows the same set of rules as for Files:
+  - The original directory name must be preserved even if the path to it has changed.
+  - Two input directories with the same name must be located separately, to avoid name collision.
+  - Two input directories which originated in the same parent storage directory must be localized into the same parent directory for task execution.
```
pshapiro4broad: What does this mean? Does it mean if you have input directory `/a/b/c` and input directory `/a/b/d` then the localized paths must look like `"/some/prefix/b/c"` and `/some/prefix/b/d`? If so, does this only apply to a directory's immediate parent? Some examples might help to make required behavior clear.

geoffjentry: Just a reminder that there'll be no changes as part of this PR at this stage. If the answers from @cjllanwarne to your questions lead you towards a yes or no vote, then great.

It's also AOK to suggest followup changes (e.g. more examples)

orodeh:
👍

versions/development/SPEC.md
vdauwera: I would like to see a few more examples please -- particularly toward documenting behavior wrt the point raised by @pshapiro4broad

geoffjentry:
@vdauwera there is none. You can approve or disapprove as-is

dinvlad:
How well will it scale - e.g. if I'd like to localize 100K files - will it have to check hash for all of them? :)

geoffjentry:
@dinvlad wouldn't that depend on the implementation?

versions/development/SPEC.md
cjllanwarne: 1. As written, I think that this line is a pretty weak guarantee: that the directories would end up in `/some/prefix/c` and `/some/prefix/d`. Of course an engine is free to include more of the original path within `/some/prefix/` if it chooses to (as Cromwell does in the prototype https://github.com/broadinstitute/cromwell/pull/3980).
1. This was added to mirror the guarantees that exist for files (eg so that `.bam` and `.bai` inputs remain co-located if they start off next to each other).
1. I agree that it is a little ambiguous as-is, so (assuming this PR merges) I can make a clarification/alteration/examples PR to follow with.

cjllanwarne: (or, if you think this is gating, you can vote against this as-is and request I resubmit with that change made)

antonkulaga:
+1

pshapiro4broad:
👍

versions/development/SPEC.md
pshapiro4broad: Thanks for the clarification. It doesn't seem worth rejecting based on this, but making the clarification part of a future change to the spec would be nice.

patmagee:
:+1

aprabhak2:
+1

aednichols:
👍

geoffjentry:
Sorry for the delay on this, this passes (even without the absentee ballot of @aednichols ) and is now awaiting an implementation

cjllanwarne:
This is now merged and available in Cromwell develop (nb: use `version development` in your WDL to try it out): https://github.com/broadinstitute/cromwell/pull/3980

@geoffjentry I believe that this PR can now be merged.

PR #244, opened 2018-08-15
Adding keys function

ffinfo:
No explanation needed, see the code ;)

### DIFF
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -121,6 +121,7 @@
   * [Array\[Pair(X,Y)\] cross(Array\[X\], Array\[Y\])](#arraypairxy-crossarrayx-arrayy)
   * [Array\[Pair(X,Y)\] as_pairs(Map\[X,Y\])](#arraypairxy-as_pairsmapxy)
   * [Map\[X,Y\] as_map(Array\[Pair(X,Y)\])](#mapxy-as_maparraypairxy)
+  * [Array\[X\] keys(Map\[X,Y\])](#arrayx-keysmapxy)
   * [Map\[X,Array\[Y\]\] collect_by_key(Array\[Pair(X,Y)\])](#mapxarrayy-collect_by_keyarraypairxy)
   * [Integer length(Array\[X\])](#integer-lengtharrayx)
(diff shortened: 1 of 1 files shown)


versions/development/SPEC.md
```diff
@@ -3091,6 +3092,20 @@ Map[String,Int] xmap = as_map(x) # {"a": 1, "b": 2, "c": 3}
 Map[String,Pair[File,File]] ymap = as_map(y) # {"a": ("a.bam", "a.bai"), "b": ("b.bam", "b.bai")}
 ```
 
+## Array[X] keys(Map[X,Y])
+
+Given an Map ,the `keys` function returns a Array of the keys of map. The order of the keys in the resulting Array is the same as the order of the Pairs in the Map.
```
DavyCats: `Map ,the` should be `Map, the`.

I also feel this first sentence could read a bit awkwardly, maybe something like this: "Given a Map, the `keys` function returns an Array consisting of the keys in the Map."

ffinfo: :+1:

patmagee: I think some engine implementations (maybe most) used OrdererdDicts I am not sure that it is always going to be the case, ie if an implementation is using a hash Dict.

versions/development/SPEC.md
```diff
@@ -3091,6 +3092,20 @@ Map[String,Int] xmap = as_map(x) # {"a": 1, "b": 2, "c": 3}
...
+
+In cases where multiple Pairs would produce the same key, the workflow will fail.
```
patmagee: Can you clarify what circumstances would lead to the same key being produced? The nature of a dict would seemingly restrict this from happening

orodeh:
👍

rhpvorderman:
:+1:

cjllanwarne:
The line `In cases where multiple Pairs would produce the same key, the workflow will fail.` looks like copy and paste since it doesn't seem to apply.

But that's not a blocker: 👍 from me

geoffjentry:
👍

geoffjentry:
Voting passes and this is now waiting for implementation

PR #247, opened 2018-08-24
Specify UTF-8 encoding and tidy up string definition

EvanTheB:
de-obfuscate the description of strings. I think this is fast-trackable.

However I do not know the intended behaviour of '\n', It should probably read `any character not in set ... '<newline>' `. But are newlines really not allowed while other whitespace and strange characters are (carriage return?)

### DIFF
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -247,12 +247,14 @@ The inputs to this workflow would be `example.files` and `example.hello.pattern`
 
 ## Global Grammar Rules
 
+WDL files are encoded in UTF-8, with no BOM.
+
 ### Whitespace, Strings, Identifiers, Constants
 
 These are common among many of the following sections
 
 ```
-$ws = (0x20 | 0x9 | 0xD | 0xA)+
+$ws = (0x20 | 0x09 | 0x0D | 0x0A)+
 $identifier = [a-zA-Z][a-zA-Z0-9_]+
 $boolean = 'true' | 'false'
 $integer = [1-9][0-9]*|0[xX][0-9a-fA-F]+|0[0-7]*
@@ -262,10 +264,17 @@ $float = (([0-9]+)?\.([0-9]+)|[0-9]+\.|[0-9]+)([eE][-+]?[0-9]+)?
 `$string` can accept the following between single or double-quotes:
 
 * Any character not in set: `\`, `"` (or `'` for single-quoted string), `\n`
-* An escape sequence starting with `\`, followed by one of the following characters: `\nrbtfav?`
-* An escape sequence starting with `\`, followed by 1 to 3 digits of value 0 through 7 inclusive.  This specifies an octal escape code.
-* An escape sequence starting with `\x`, followed by hexadecimal characters `0-9a-fA-F`.  This specifies a hexadecimal escape code.
-* An escape sequence starting with `\u` or `\U` followed by either 4 or 8 hexadecimal characters `0-9a-fA-F`.  This specifies a unicode code point
+* An escape sequence starting with `\`, followed by one of the following characters: `\nt"'`
+* An escape sequence starting with `\`, followed by 3 digits of value 0 through 7 inclusive.  This specifies an octal escape code.
+* An escape sequence starting with `\x`, followed by 2 hexadecimal digits `0-9a-fA-F`.  This specifies a hexadecimal escape code.
+* An escape sequence starting with `\u` followed by 4 hexadecimal characters or `\U` followed by 8 hexadecimal characters `0-9a-fA-F`.  This specifies a unicode code point.
+
+|Escape Sequence|Meaning|\x Equivalent|
+|`\\`|`\`|`\x5C`|
+|`\n`|newline|`\x0A`|
+|`\t`|tab|`\x09`|
+|`\'`|single quote|`\x22`|
+|`\"`|double quote|`\x27`|
 
 ### Types
 
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -254,8 +254,6 @@ These are common among many of the following sections
 ```
 $ws = (0x20 | 0x9 | 0xD | 0xA)+
 $identifier = [a-zA-Z][a-zA-Z0-9_]+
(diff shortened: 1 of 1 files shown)


DavyCats:
@EvanTheB  Perhaps it should read:
```
Any printable (non-control) character not in set `\`, `"` (or `'` for single-quoted string).
```

Maybe even specifically only ASCII characters? Unicode characters can still be encoded using `\u`, anyway.

EvanTheB:
Yes I think you are right, maybe this is the time to specify that wdls are utf-8 encoded?

patmagee:
@EvanTheB +1 to UTF8, although do you think there should be flexibility in this?

versions/development/SPEC.md
```diff
@@ -254,20 +254,18 @@ These are common among many of the following sections
...
 
 `$string` can accept the following between single or double-quotes:
 
-* Any character not in set: `\\`, `"` (or `'` for single-quoted string), `\n`
-* An escape sequence starting with `\\`, followed by one of the following characters: `\\`, `"`, `'`, `[nrbtfav]`, `?`
-* An escape sequence starting with `\\`, followed by 1 to 3 digits of value 0 through 7 inclusive.  This specifies an octal escape code.
-* An escape sequence starting with `\\x`, followed by hexadecimal characters `0-9a-fA-F`.  This specifies a hexadecimal escape code.
-* An escape sequence starting with `\\u` or `\\U` followed by either 4 or 8 hexadecimal characters `0-9a-fA-F`.  This specifies a unicode code point
+* Any character not in set: `\`, `"` (or `'` for single-quoted string), `\n`
+* An escape sequence starting with `\`, followed by one of the following characters: `\nrbtfav?`
```
cjllanwarne: Can we break this into separate sub-bullets indicating what each escape sequence is interpreted as? 

... and we should probably double check that all of these actually make sense...

versions/development/SPEC.md
```diff
@@ -254,20 +254,18 @@ These are common among many of the following sections
...
 ```
 
 `$string` can accept the following between single or double-quotes:
 
-* Any character not in set: `\\`, `"` (or `'` for single-quoted string), `\n`
-* An escape sequence starting with `\\`, followed by one of the following characters: `\\`, `"`, `'`, `[nrbtfav]`, `?`
-* An escape sequence starting with `\\`, followed by 1 to 3 digits of value 0 through 7 inclusive.  This specifies an octal escape code.
-* An escape sequence starting with `\\x`, followed by hexadecimal characters `0-9a-fA-F`.  This specifies a hexadecimal escape code.
-* An escape sequence starting with `\\u` or `\\U` followed by either 4 or 8 hexadecimal characters `0-9a-fA-F`.  This specifies a unicode code point
+* Any character not in set: `\`, `"` (or `'` for single-quoted string), `\n`
```
cjllanwarne: We should also mention that `~{` indicates a placeholder

versions/development/SPEC.md
```diff
@@ -254,20 +254,18 @@ These are common among many of the following sections
...
+* An escape sequence starting with `\`, followed by one of the following characters: `\nrbtfav?`
+* An escape sequence starting with `\`, followed by 1 to 3 digits of value 0 through 7 inclusive.  This specifies an octal escape code.
+* An escape sequence starting with `\x`, followed by hexadecimal characters `0-9a-fA-F`.  This specifies a hexadecimal escape code.
+* An escape sequence starting with `\u` or `\U` followed by either 4 or 8 hexadecimal characters `0-9a-fA-F`.  This specifies a unicode code point
```
cjllanwarne: If WDL **is** set to unicode, would we remove these code point options?

versions/development/SPEC.md
EvanTheB: these are the same as the c89 escape sequences. Are we happy to refer to other documents or do we want to define here?

versions/development/SPEC.md
EvanTheB: I don't want to be pedantic, but in a spec pedantry is kind of required; the proposal was to specify UTF-8 encoding of wdl files. unicode is not an encoding. On your point, \u would always be useful, because most users would find it hard or impossible to enter non-ascii code points using their keyboard. 

Python did: https://www.python.org/dev/peps/pep-0263/ 
Which allows the encoding of the file to be specified in the file. That decision predates the dominance of UTF-8. Golang more recently specifies source files to be utf-8 encoded. I think there are 2 sensible options: specify ascii, or specify utf-8.

EvanTheB:
@patmagee I don't see any reason for choice, choice means a mechanism for communicating the choice and making all subsequent decisions abstract enough to deal with all possibilities. Python is gross: `# -*- coding: latin-1 -*-`

versions/development/SPEC.md
DavyCats: Just as a note here: the `~{` placeholder syntax is not mentioned under the string interpolation section. This syntax (as per the current SPEC) only applies to the command section.

versions/development/SPEC.md
cjllanwarne: I'd be happy to reference out to other docs or redefine. One reason for listing out is to make it explicit which ones we're actually supporting since some of these do seem unnecessary (eg IIRC `b` is "backspace", which IMO we could probably do without)

EvanTheB:
I have specified UTF-8, explained each escape, and removed many. Still to do; explain ~{} and ${}, and unify string literals with the 'command' thing. It seems like every concept is invented 3 times in WDL ;).


https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals
http://www.asciitable.com/
https://unicode-table.com/en/#07C8
https://www.python.org/dev/peps/pep-0263/

orodeh:
👍

DavyCats:
:+1:

LeeTL1220:
:+1: UTF-8, excellent!

cjllanwarne:
👍 but a warning that UTF-8 and "hermes parser" may be incompatible, so 🤞 but I'm afraid this might be in "awaiting implementation" quite a while until we can get away from hermes...

patmagee:
👍

aednichols:
👍

dheiman:
:+1:

geoffjentry:
@EvanTheB Please rollback the commit you made after voting went live. If you want to apply that it'll need to be done separately

geoffjentry:
@EvanTheB Just looping back around to remind about removing that extraneous commit at the end

EvanTheB:
I think I have done it, it is hard to tell.

geoffjentry:
Implemented in cromwell

PR #253, opened 2018-09-21
Properly unescape strings in 1.0

aednichols:
Between draft-2 and 1.0, we stopped calling `wdl_unescape` for some reason (but the function stuck around).

This PR restores a tweaked version of it that is compatible with other 1.0 changes.


geoffjentry:
@aednichols to be clear, this is purely a bug fix to the grammar file, correct?

aednichols:
Yes, that's the idea. Purely a regression relative to `draft-2`, the spec says the same thing about strings/escapes/regex.

geoffjentry:
@aednichols Great. This doesn't require voting in that case. Merging.

PR #256, opened 2018-09-24
Switch to 'version development' from 'version 1.1-draft'

cjllanwarne:
I propose using `development` in our "in development" grammar in place of "1.1-draft", since we may want or need to use a different version number by the time we perform the next "official release".


geoffjentry:
This isn't a spec change and is in line with both a) how things are described in github and b) the fact that `1.1` is likely to be `2.0` (yay semver) so just merging as-is

PR #257, opened 2018-09-25
Rework of 'Scopes' and 'Variable Resolution' Sections

cjllanwarne:
A rework of two old, mostly out of date sections, `Scope` and `Variable Resolution`, including a bunch of new and amended examples.

---

**Previous Title**: "Amend Example: Declarations don't exist in call inputs" 
**Previous Description**:

Amending what was, in my opinion, a bad example:

```wdl
workflow wf {
...
  call my_task {
    String s = "my_task_s"
    input: in0 = s+"-suffix", in1 = t+"-suffix"
  }
...
}
```

The `Call Statement` section does not allow this syntax and nor does the official grammar file.


mlin:
The diff here contains a few important spec clarifications which go beyond the current PR title! (see also #265)

suggestions:

- further explicate that within tasks and workflows, Expressions may contain forward variable references so long as they do not imply circular dependencies.
- "value names are unique" <-- is the appropriate term "value" or "variable"?
- [a previous example under Scope](https://github.com/openwdl/wdl/blame/master/versions/development/SPEC.md#L2226) needs revision as it has both variable name collisions, and a within-call declaration

Let me know if you'd like me to riff on this @cjllanwarne

cjllanwarne:
Thanks @mlin, I'm very happy to! I'll try to make the changes you suggest in #265 too and probably rename this PR too to be more representative of what it's trying to do. I'll let you know when I've got something for you to look over.

And FWIW I always try to remember to say "value" instead of "variable" since they're set-once and can't be updated once their value is assigned.

patmagee:
@cjllanwarne @mlin where do we stand with this PR? Is it still under discussion, or should it be sent to voting?

versions/development/SPEC.md
```diff
@@ -78,7 +78,7 @@
       * [Struct Member Access](#struct-member-access)
       * [Importing Structs](#importing-structs)
 * [Namespaces](#namespaces)
-* [Scope](#scope)
+* [Value Scopes and Reachability](#value-scopes-and-reachability)
```
pshapiro4broad: I'm not used to seeing the term `Reachability` in this context. Most language descriptions I've seen use `Visibility` to describe this concept.

versions/development/SPEC.md
```diff
@@ -2138,44 +2138,280 @@ Now everything inside of `tasks.wdl` must be accessed through the default namesp
...
+# Value Scopes and Reachability
 
-Scopes are defined as:
+## Top Level Scopes
 
-* `workflow {...}` blocks
-* `call` blocks
-* `if(expr) {...}` blocks
-* `scatter(x in y) {...}` blocks
+The top level scopes which can contain value declarations are:
```
pshapiro4broad: This is a minor point, but I think where the spec says `declaration` what's actually meant is `definition`. Conventionally, a declaration binds a name in a scope but doesn't provide the name's value, or require that a value is defined. E.g. in C/C++, you put function declarations in header files and function definitions in source (.c/cpp) files. Some languages don't support declarations separate from definitions, and I think WDL falls into this category. Therefore all named values are definitions (as are tasks and workflows).

versions/development/SPEC.md
```diff
@@ -2138,44 +2138,280 @@ Now everything inside of `tasks.wdl` must be accessed through the default namesp
...
 
-Inside of any scope, variables may be [declared](#declarations).  The variables declared in that scope are visible to any sub-scope, recursively.  For example:
+* `workflow {...}` definitions
+* `task {...}` definitions
+
+Within these top-level scopes, various sub-sections exist which can also contain value declarations:
+
+* `scatter` blocks
+* `if` blocks
+* `input` sections
```
pshapiro4broad: This is probably not important for this PR, but it's unclear what the difference between blocks and sections are. The text also uses `section` to refer to sections of the document itself, so using `block` everywhere might be better. Unless there's a semantic difference between blocks and sections.

versions/development/SPEC.md
```diff
@@ -2138,44 +2138,280 @@ Now everything inside of `tasks.wdl` must be accessed through the default namesp
...
+
+Values declared within a `task` are either inputs or outputs. Workflows have inputs and outputs as well, and also allow declaration of "graph" values.
+
+* Inputs are evaluated before anything else in the task or workflow is started.
+* Outputs are evaluated after everything else in the task or workflow completes.
+* Graph values are evaluated as part of the normal execution of the workflow DAG. This means that they can be evaluated as soon as all their dependencies are available.
+* Examples of dependencies which graph values may have include:
+    * Inputs
+    * `call` outputs
+    * Other graph values
```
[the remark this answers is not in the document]
pshapiro4broad: I agree that the use of `graph` and `DAG` don't make things clearer and, to me, are implementation details. I think the important point to make is that values that are defined at the top level of a workflow are evaluated in an order determined by their dependencies, which is implied by the acyclic description below. The evaluation order could be made explicit here as well.

Maybe a less loaded name for these values would be top-level values, or workflow values. (vs workflow input values and workflow output values)

versions/development/SPEC.md
```diff
@@ -2138,44 +2138,280 @@ Now everything inside of `tasks.wdl` must be accessed through the default namesp
...
+* `input` sections
+* `output` sections
+
+Some sub-sections provide scoped values:
+
+* `scatter(x in ...)` provides a value `x` **only within the scatter's `{...}` block**
+
+## Input, Output and Graph Values
+
+Values declared within a `task` are either inputs or outputs. Workflows have inputs and outputs as well, and also allow declaration of "graph" values.
```
pshapiro4broad: I think it would be clearer to avoid the use of `graph` here, e.g.
```suggestion
Values declared within a `task` are either inputs or outputs. Workflows can define values in an input block, an output block, or at the top level.
```

versions/development/SPEC.md
geoffjentry: At a quick glance (IOW: he's shooting from the hip and not validating what he says) this is likely an artifact of "The spec matches terminology to how Hermes describes things", which isn't really a great policy going forward

mlin:
@cjllanwarne do you have bandwidth on this in the next month, or would it help for me or someone else to drive it in that timeframe? (cc @patmagee)

mlin:
@cjllanwarne gentle ping

cjllanwarne:
@mlin sorry for being radio silent for much of the last month - yes, hopefully things are looking a little more normal over here now and I'll be able to pick this up again and start moving on some of the remaining "awaiting implementation" tickets too.

If that ends up not being true (eg if something else comes in from the side and/or my attention gets derailed again) - I'll definitely remember the offer and maybe ping you then to ask for help. Thanks!

mlin:
@cjllanwarne awesome let me know how i can help! Somewhat related to this PR topic, I noticed we need to explicate name collision rules (e.g. a value and a call mustn't have the same name...value/task? call/task? import/struct? etc.) and provide a list of keywords which are illegal as names (e.g. `command` `task` `left` ...)

versions/development/SPEC.md
cjllanwarne: I'll happily go along with `definition`.

versions/development/SPEC.md
cjllanwarne: Fair enough. In my head I use the two interchangeably but I'm happy to prefer `block` over section for the sake of clarity

versions/development/SPEC.md
cjllanwarne: I'll rework this section and rename "graph value" to "intermediate value" since that I think has been used elsewhere in the spec

versions/development/SPEC.md
```diff
@@ -2138,44 +2138,280 @@ Now everything inside of `tasks.wdl` must be accessed through the default namesp
...
+
+* Inputs are evaluated before anything else in the task or workflow is started.
```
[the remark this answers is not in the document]
cjllanwarne: I've made a mention of this - and reworded this to be more explicit - all values can really be evaluated as soon as their inputs are available, and there's not really much special about inputs other than that they can be supplied by users, or outputs except that they get collected for users at the end.

versions/development/SPEC.md
cjllanwarne: I'll replace with visibility

versions/development/SPEC.md
```diff
@@ -2138,44 +2138,280 @@ Now everything inside of `tasks.wdl` must be accessed through the default namesp
...
+* Outputs are evaluated after everything else in the task or workflow completes.
```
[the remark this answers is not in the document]
cjllanwarne: I believe they would be considered "intermediate values" even if they follow the `command`. I'll make it clearer that that means they get evaluated before the command runs.

That said - I can foresee a value (perhaps in a more targeted future PR) in allowing some way - perhaps positionally - of specifying "post-command, non-output values"

versions/development/SPEC.md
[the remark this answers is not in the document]
cjllanwarne: I'll try rewording and replacing "graph value" with "intermediate value" since that's the term used elsewhere in the spec.

versions/development/SPEC.md
```diff
@@ -2138,44 +2138,280 @@ Now everything inside of `tasks.wdl` must be accessed through the default namesp
...
+
+## Value Reachability Rules
+
+* Within a top-level (`workflow` or `task`) scope, each value name is unique and always reachable.
+    * Exception: `scatter` value names 
+        * For example the value named `x` in the line `scatter(x in ...) { ... }`
+        * Scatter values are only accessible within the scope of the scatter block
+        * Scatter value names can be reused as scatter value names by other scatter blocks
+    * Exception: `output` values
+        * Because they must be calculated last, `output` values are *not* reachable from outside the `output` section.
```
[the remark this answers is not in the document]
cjllanwarne: I think so - it should be possible to determine an order on outputs just as easily as for inputs/intermediate values.

versions/development/SPEC.md
```diff
@@ -2138,44 +2138,280 @@ Now everything inside of `tasks.wdl` must be accessed through the default namesp
...
+* A graph value may be used as a depenency higher up in the WDL file than the location where it is declared.
+    * This is OK because graph execution order is defined by the DAG, not the order in the page.
+    * See the [Forward Reference](#forward-reference) example.
+* Despite values being reachable from outside sub-sections, the types of values may change if accessed from outside:
+    * Using a value which was declared within a `scatter` from outside that `scatter` implicitly collects the results, turning any X into an `Array[X]`.
+    * Using a value which was declared within a `if` from outside the `if` implicitly optionalizes the results, turning any `X` into an `X?`.
+* References must always be acyclic:
+    * If "`a` depends on `b`" then it cannot also be true that "`b` depends on `a`".
+    * References must also not cause cyclic dependencies between sub-sections. See the [Acyclic Sub-Sections](#acyclic-sub-sections) for an example of what this means.
```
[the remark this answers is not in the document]
cjllanwarne: I'll make that explicit

cjllanwarne:
@mlin @pshapiro4broad thanks for the feedback - I think I responded to all of it in my latest commit.

Unfortunately I got a little bit ahead of myself with my previous comment since I'm actually heading out on vacation now for 2 weeks...

I don't want to make that a blocker for you however - if you have more feedback and/or feel inclined or motivated to make the changes yourself in a fork of this branch, please do go ahead - I promise I won't be offended!

One other comment - I believe this is a clarification of existing behavior (at least wrt Cromwell) rather than a change - so if the PR does get voted on, I believe it would be immediately ready for a merge.

versions/development/SPEC.md
```diff
@@ -2250,44 +2254,287 @@ Now everything inside of `tasks.wdl` must be accessed through the default namesp
...
+  output {
+    Int z = read_int(stdout())
+    Int z_plus_one = z + 1
   }
 }
+```
 
-workflow wf {
+* `x` and `f` are `input` values which will be evaluated before the task begins.
+* `y` is a graph value with an upstream depenency on the input `x`.
```
pshapiro4broad:
```suggestion
* `y` is an intermediate value with an upstream dependency on the input `x`.
```

versions/development/SPEC.md
```diff
@@ -2250,44 +2254,287 @@ Now everything inside of `tasks.wdl` must be accessed through the default namesp
...
+* The `command` section is able to access all `input`s and graph values. It is not able to reference `output` values.
```
pshapiro4broad:
```suggestion
* The `command` section is able to access all `input`s and intermediate values. It is not able to reference `output` values.
```

orodeh:
👍 

This certainly covers #316, and can replace it; it does cover a lot more ground though.

patmagee:
@cjllanwarne how do we feel about this PR? I would like to send it to voting if you are confident there are no more changes needed

cjllanwarne:
@patmagee sorry, I missed your question earlier. Yes, I think this is ready for voting.

pshapiro4broad:
👍 Thanks for all your work on this @cjllanwarne

orodeh:
👍

DavyCats:
:+1:

rhpvorderman:
:+1:

cjllanwarne:
Voting succeeds and this PR is now approved pending implementation.

cjllanwarne:
Because the PR is clarifying and disambiguating existing behavior - rather than proposing changes per se - I'm considering this PR as implemented and will go ahead and immediately merge the PR.

PR #261, opened 2018-10-24
Update wdl_runner to support pipelines v2

kdbinder:


runners/cromwell_on_google/wdl_runner/file_util.py
```diff
@@ -15,7 +15,7 @@
 
 from googleapiclient import discovery
 from googleapiclient.errors import HttpError
-from oauth2client.client import GoogleCredentials
+#from oauth2client.client import GoogleCredentials
```
kemp-google: Why drop the credentials?

runners/cromwell_on_google/wdl_runner/wdl_pipeline.yaml
```diff
@@ -1,24 +1,9 @@
...
-    /wdl_runner/wdl_runner.sh
+actions:
+- name: WDL_Runner
+  commands: [ '/wdl_runner/wdl_runner.sh' ]
+  imageUri: gcr.io/google.com/kbinder-sandbox/wdl_runner 
 
 resources:
-  minimumRamGb: 3.75
+   virtualMachine:
+     machineType: n1-standard-2
```
kemp-google: Why the change here?  I would have thought we were getting a n1-standard-1 before?  Sorry if I missed something

runners/cromwell_on_google/wdl_runner/wdl_runner.sh
```diff
@@ -15,9 +15,9 @@ readonly INPUT_PATH=/pipeline/input
 # the pipeline as environment variables - write them out as
 # files.
 mkdir -p "${INPUT_PATH}"
-echo "${WDL}" > "${INPUT_PATH}/wf.wdl"
-echo "${WORKFLOW_INPUTS}" > "${INPUT_PATH}/wf.inputs.json"
-echo "${WORKFLOW_OPTIONS}" > "${INPUT_PATH}/wf.options.json"
+cp "${WDL}"  "${INPUT_PATH}/wf.wdl"
```
kemp-google: Extra space between the two quoted expressions (here and below)

runners/cromwell_on_google/wdl_runner/wdl_runner.sh
```diff
@@ -15,9 +15,9 @@ readonly INPUT_PATH=/pipeline/input
 # the pipeline as environment variables - write them out as
```
kemp-google: Can you update this comment?  The filenames are passed now, not content

runners/cromwell_on_google/wdl_runner/file_util.py
kdbinder: we deprecated oauth2client in favor of google-auth in our python libraries so its not needed

kemp-google: Awesome, let's remove it entirely then

runners/cromwell_on_google/wdl_runner/wdl_pipeline.yaml
kdbinder: You're right.  Should be n1-standard-1

kemp-google:
LGTM once we get the docker image published and the URL changed (to not point at your sandbox image).

kdbinder:
Readme needs slight modification as args require comma separation now

geoffjentry:
@kdbinder @kemp-google Can you all let me know when you think this is ready? Also, to go back to my trello comment - at some point we're going to be moving this "somewhere else" (location TBD). There were some unexpected hurdles on the Broad side which we're working through (which is also the reason for the delay on upgrading the Cromwell version).

ruchim:
👍

kdbinder:
@geoffjentry @kemp-google @ruchim @moschetti 

I believe we are good to go now that the image has been updated

geoffjentry:
Perfect, thanks @kbinder. I’ll take care of this tomorrow and then will coordinate with @cjllanwarne on a plan to migrate this

moschetti:
@kdbinder @geoffjentry Thanks for the update. Once this is in master I'll push the changes to the Google Cloud [tutorial docs](https://cloud.google.com/genomics/docs/tutorials/gatk)

kemp-google:
Hooray! Thanks everybody!

Issue #209, opened 2018-05-04
Some literal for "Null" is needed.


rhpvorderman:
I cannot find it anywhere in the spec. 
But suppose you have a task with the following inputs:
```WDL
task someTask {
  File read1
  File? read2
  String read1output
  String? read2output
  command {
    echo "some task"
  }
}
```
Now I want to run a workflow that processes a read1 and an optional read2
 Most programs will crash when a read2output is defined while no paired-end input is given. 
So in a workflow I want to do this:
```WDL
workflow someWf {
  File read1
  File? read2
  File outputDir
  call someTask {
     input:
       read1 = read1
       read2 = read2
       read1output = outputDir + "/processedRead1.fastq.gz"
       read2output = if (defined(read2)) then outputDir + "/processedRead2.fastq.gz" else NULL
  } 
}
```
But this is not possible so I do `else read2` (since read2 should evaluate to NULL if it is not present). But I don't think this is a proper way of doing this.

cjllanwarne:
I think this is a great idea.

My only comment is that I think I'd prefer something other than `null` because of its connotations from other languages.

What I mean is, in most languages, you can assign `null` anything to anything, but in WDL you should only be able to assign this value to an optional. Maybe something like `NONE`, `EMPTY` or `UNSET` would be better?

patmagee:
I would favor either `NONE` or `UNSET`. `EMPTY` to me implies an array or map type of object.

pshapiro4broad:
> What I mean is, in most languages, you can assign null anything to anything, but in WDL you should only be able to assign this value to an optional. Maybe something like NONE, EMPTY or UNSET would be better?

This is not true of languages that have nullable types. In those languages (such as `C#`, which uses the same syntax to declare a nullable type) it's a compile time error to assign null (or any nullable typed expression) to a non-nullable type.

rhpvorderman:
I would prefer `NONE` that is what is used in python and therefore  most recognizable to most scientific users, and frankly most other users as well. Also `NONE` is used in scala as well, so I presume it is a nice thing for those who write the execution engine as well.

patmagee:
I do like `NONE`. in terms of the ease of use, I think most people would understand what `NONE` meant, without much explanation. 

I am not sure though if this needs to stand up to equality tests. ie whether we need to do a `x == NONE` or `x  == null`. IMO thats captured by `defined(x)`. `NONE` to me is a placeholder to state taht a variable does not have a value

[...]

PR #263, opened 2018-11-09
Add "None" to spec

rhpvorderman:
After running into yet another case where `None` is definitely needed (or otherwise causing hideous workarounds) I decided to at it as an example in this pull request. 

This aims to implement things that were discussed in #209

### DIFF
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -3280,9 +3280,32 @@ Array[String] env2_param = prefix("-f ", env2) # ["-f 1", "-f 2", "-f 3"]
 
 Given an array of optional values, `select_first` will select the first defined value and return it. Note that this is a runtime check and requires that at least one defined value will exist: if no defined value is found when select_first is evaluated, the workflow will fail.
 
+```WDL
+version 1.0
+workflow SelectFirst {
+  input {
+    Int? maybe_integer = 5
+    Int? maybe_number = None
+    Int? maybe_numeral = 3
+  }
+  Int five = select_first([maybe_integer, maybe_number, maybe_numeral]) # This evaluates to 5
+  Int five = select_first([maybe_number, maybe_integer, maybe_numeral]) # This also evaluates to 5
+}
+
 ## Array[X] select_all(Array[X?])
 
 Given an array of optional values, `select_all` will select only those elements which are defined.
+
+```WDL
+version 1.0
+workflow SelectFirst {
+  input {
+    Int? maybe_integer = 5
+    Int? maybe_number = None
+    Int? maybe_numeral = 3
+  }
+  Array[Int] fivethree = select_all([maybe_integer, maybe_number, maybe_numeral]) # This evaluates to [5, 3]
+}
 
 ## Boolean defined(X?)
 
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -2334,6 +2334,7 @@ task test
     File? logFile = if logFile then "test.log" else None
   }
 }
+```
 If `logFile` is true, the command will evaluate to `test --log inputFile outputPath` and a log will be created.
 If `logFile` is false, the command will evaluate to `test inputFile outputPath` and log will not be created.
 
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -698,7 +698,8 @@ Pair values can also be specified within the [workflow inputs JSON](https://gith
 =======
 ### Optional literals
 
-Any non-optional value can be  turned into an optional value by assigning it to an optional variable:
+Any non-optional value can be stored into an optional variable by assigning it to an optional variable.
+This does not change the value, but it changes the type of the value:
 
 ```WDL
 Int? maybe_five = 5
@@ -3294,12 +3295,12 @@ Given an array of optional values, `select_first` will select the first defined 
 version 1.0
 workflow SelectFirst {
   input {
-    Int? maybe_integer = 5
-    Int? maybe_number = None
-    Int? maybe_numeral = 3
+    Int? maybe_five = 5
+    Int? maybe_four_but_is_not = None
+    Int? maybe_three = 3
   }
-  Int five = select_first([maybe_integer, maybe_number, maybe_numeral]) # This evaluates to 5
-  Int five = select_first([maybe_number, maybe_integer, maybe_numeral]) # This also evaluates to 5
+  Int five = select_first([maybe_five, maybe_four_but_is_not, maybe_three]) # This evaluates to 5
+  Int five = select_first([maybe_four_but_is_not, maybe_five, maybe_three]) # This also evaluates to 5
 }
 
 ## Array[X] select_all(Array[X?])
@@ -3310,11 +3311,11 @@ Given an array of optional values, `select_all` will select only those elements 
 version 1.0
 workflow SelectFirst {
   input {
-    Int? maybe_integer = 5
-    Int? maybe_number = None
-    Int? maybe_numeral = 3
+    Int? maybe_five = 5
+    Int? maybe_four_but_is_not = None
+    Int? maybe_three = 3
   }
-  Array[Int] fivethree = select_all([maybe_integer, maybe_number, maybe_numeral]) # This evaluates to [5, 3]
+  Array[Int] fivethree = select_all([maybe_five, maybe_four_but_is_not, maybe_three]) # This evaluates to [5, 3]
 }
 
 ## Boolean defined(X?)
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -698,7 +698,7 @@ Pair values can also be specified within the [workflow inputs JSON](https://gith
 =======
 ### Optional literals
 
-Any non-optional value can be wrapped into an optional value by assigning it to an optional variable:
+Any non-optional value can be  turned into an optional value by assigning it to an optional variable:
 
 ```WDL
 Int? maybe_integer = 5
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -727,8 +727,8 @@ task test
   }
 }
 ```
-If `log` is true, the command will evaluate to `test --log inputFile outputPath` and `logFile` will be a defined optional.
-If `log` is false, the command will evaluate to `test inputFile outputPath` and `logFile` will be an undefined optional.
+If `log` is true, the command will evaluate to `test --log inputFile outputPath` and `logFile` will be `test.log`.
+If `log` is false, the command will evaluate to `test inputFile outputPath` and `logFile` will be `None`.
 
 ## Document
 
@@ -2282,7 +2282,7 @@ task test {
     Array[File]+ b
     Array[File]? c
     #File+ d <-- can't do this, + only applies to Arrays
-    Array[File]+? e  # An optional array that, if defined, must contain more than one element
+    Array[File]+? e  # An optional array that, if defined, must contain at least one element
   }
   command {
     /bin/mycmd ${sep=" " a}
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -712,11 +712,11 @@ Int? maybe_integer = None
 Int? maybe_numeral = 5
 Int certain_integer = 3
 # maybe_numeral is a defined optional
-is_defined(maybe_integer) # Evaluates to false
-is_defined(maybe_numeral) # Evaluates to true
-maybe_integer == None # Evaluates to true, same as !is_defined(maybe_integer)
-maybe_integer != None # Evaluates to false, same as is_defined(maybe_integer)
-certain_integer == None # This will cause an error. Since None only makes sense when dealing with optionals.
+Boolean test_defined = defined(maybe_integer) # Evaluates to false
+Boolean test_defined2 = defined(maybe_numeral) # Evaluates to true
+Boolean test_is_none = maybe_integer == None # Evaluates to true, same as !defined(maybe_integer)
+Boolean test_not_none = maybe_integer != None # Evaluates to false, same as defined(maybe_integer)
+Boolean compare_int_to_none = certain_integer == None # This will cause an error. Since None only makes sense when dealing with optionals.
 ```
 
 This is useful in a number of cases, for example when an output of a command depends on a certain flag:
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -79,6 +79,7 @@
 * [Scope](#scope)
 * [Optional Parameters & Type Constraints](#optional-parameters--type-constraints)
   * [Prepending a String to an Optional Parameter](#prepending-a-string-to-an-optional-parameter)
+  * [The None argument](#the-none-argument)
 * [Scatter / Gather](#scatter--gather)
 * [Variable Resolution](#variable-resolution)
   * [Task-Level Resolution](#task-level-resolution)
@@ -2319,7 +2320,7 @@ python script.py ${"--val=" + val}
 
 ## The `None` argument
 
-Sometimes you need to define an output that is optional, because it depends on a command parameter for example:
+Sometimes you need to define an output that is optional, because it depends on a command parameter. For example:
 ```wdl
 task test
   input {
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -721,7 +721,7 @@ certain_integer == None # This will cause an error. Since None only makes sense 
 
 This is useful in a number of cases, for example when an output of a command depends on a certain flag:
 ```wdl
-task test
+task test {
   input {
     File inputFile
     String outputPath
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -24,7 +24,6 @@
     * [Object Literals](#object-literals)
     * [Pair Literals](#pair-literals)
     * [Optional Literals](#optional-literals)
-  * [Document](#document)
   * [Versioning](#versioning)
   * [Import Statements](#import-statements)
   * [Task Definition](#task-definition)
@@ -735,12 +734,6 @@ task test {
 ```
 If `log` is true, the command will evaluate to `test --log inputFile outputPath` and `logFile` will be `test.log`.
 If `log` is false, the command will evaluate to `test inputFile outputPath` and `logFile` will be `None`.
-
-## Document
-
-```
-$document = ($import | $task | $workflow)+
-```
 
 ## Versioning
 
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -23,11 +23,8 @@
     * [Map Literals](#map-literals)
     * [Object Literals](#object-literals)
     * [Pair Literals](#pair-literals)
-<<<<<<< HEAD
-=======
     * [Optional Literals](#optional-literals)
   * [Document](#document)
->>>>>>> process feedback, move chapter on None
   * [Versioning](#versioning)
   * [Import Statements](#import-statements)
   * [Task Definition](#task-definition)
@@ -263,7 +260,7 @@ $float = (([0-9]+)?\.([0-9]+)|[0-9]+\.|[0-9]+)([eE][-+]?[0-9]+)?
 
 ### Types
 
-In WDL *all* types represent immutable values. 
+In WDL *all* types represent immutable values.
   - Even types like `File` and `Directory` represent logical "snapshots" of the file or directory at the time when the value was created.
   - It's impossible for a task to change an upstream value which has been provided as an input: even if it makes changes to its local copy the original value is unaffected.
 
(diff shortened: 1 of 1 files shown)


antonkulaga:
I think that you should have started with Some/None support instead of null-s from the very begining

cjllanwarne:
1. I totally agree with the need for a `None` literal! Thanks for making this PR.
2. Instead of only mentioning `None` all the way down on line 2280 (not your fault - the length of the spec is a known issue...), I think I would add this concept in a new `Optional Literals` section below `Pair Literals` (see line 26 in the ToC) and talk about how any non-optional value can be automatically wrapped into an optional (eg `Int? maybe_integer = 5`), and how `None` can be used as a literal to specify no value.
3. Side comment - for optional `File` outputs specifically (your example) - I think that eventually we'll want a way of expressing this using an engine function like `File? logfile = optional_file("test.log")` - this (a) eliminates the need for coordinating behavior through the command and the outputs using the boolean `logfile`, and (b) works in more general cases where we cannot predict using an `if` expression which `File`s will and won't be produced.
    - But for now, the example is great!

versions/development/SPEC.md
```diff
@@ -2275,6 +2277,30 @@ The latter case is very likely an error case, and this `--val=` part should be l
...
+  }
+  output {
+    File? logFile = if logFile then "test.log" else None
+  }
+}
+```
+If `logFile` is true, the command will evaluate to `test --log inputFile outputPath` and a log will be created.
+If `logFile` is false, the command will evaluate to `test inputFile outputPath` and log will not be created.
+
+Using a non-optional `File` will crash when `logFile` is false. The `None` directive is there for these and
```
pshapiro4broad: > Using a non-optional `File` will crash when `logFile` is false.

Do you mean that `File logFile = if logFile then "test.log" else None` is not valid WDL and will result in a compiler error? That seems correct to me, as the WDL compiler will determine that the type of the `if` expression is `File?` and then disallow the assignment.

> The `None` directive is there for these and other cases where an optional variable is conditionally defined.

Am not sure what you mean by `directive` in this context. I think it would be more general to define `None` as a literal that can be used anywhere an expression with an optional type is valid. E.g. it's natural to think of
```
input {
    File? inputFile
}
```
being shorthand for
```
input {
    File? inputFIle = None
}
```

cjllanwarne: FWIW I also prefer the term `literal` over `directive`

rhpvorderman:
@cjllanwarne and @pshapiro4broad thank you for your feedback. I edited the pull request.

What does everyone think about retroactively adding this to the 1.0 spec? I think that we should do this. The concept is already there. But since there is no literal for it you cannot use it easily yet. The change will be backwards compatible.

geoffjentry:
There are no retroactive changes

rhpvorderman:
@geoffjentry . Okay, that's settled then. This pull request only affects the development spec.

pshapiro4broad:
Thanks for making the changes, it looks better to me. Do you want to include an example that makes it explicit that `xxx = None` is equivalent to `defined(xxx)`?

versions/development/SPEC.md
```diff
@@ -698,6 +699,40 @@ Pair values can also be specified within the [workflow inputs JSON](https://gith
...
+  }
+  command {
+    test ~{true="--log" false="" log} inputFile outputPath
+  }
+  output {
+    File? logFile = if log then "test.log" else None
+  }
+}
+```
+If `log` is true, the command will evaluate to `test --log inputFile outputPath` and `logFile` will be a defined optional.
```
pshapiro4broad: Given that `None` is _the_ undefined optional, the wording of the phrase "will be a defined optional" seems a bit awkward. You could just as easily say "will be `None`". Here you could instead say:

When `log` is `true`, the output `logFile` has the value `"test.log"`. If it's `false`, it has the value `None`.

rhpvorderman:
Thanks @yfarjoun @pshapiro4broad and @cjllanwarne  for your excellent comments. These should now be addressed.

versions/development/SPEC.md
```diff
@@ -698,6 +699,48 @@ Pair values can also be specified within the [workflow inputs JSON](https://gith
...
+
+The `None` literal is the value that an optional has when it is not defined.
+
+```WDL
+Int? maybe_integer = None
+# maybe_integer is not defined
+Int? maybe_numeral = 5
+Int certain_integer = 3
+# maybe_numeral is a defined optional
+is_defined(maybe_integer) # Evaluates to false
```
pshapiro4broad: I think the WDL function is `defined()`, not `is_defined()`. Also would it be clearer to write these lines as if they were inputs or outputs? As it is now they're WDL code fragments. e.g.
```suggestion
Boolean test1 = defined(maybe_integer) # Evaluates to false
```

versions/development/SPEC.md
```diff
@@ -698,6 +699,48 @@ Pair values can also be specified within the [workflow inputs JSON](https://gith
 }
 ```
 
+### Optional literals
+
+Any non-optional value can be wrapped into an optional value by assigning it to an optional variable:
```
yfarjoun: wrapped -> turned into? 

I think that wrapped is too programmer-specific, not english.

versions/development/SPEC.md
```diff
@@ -698,6 +699,48 @@ Pair values can also be specified within the [workflow inputs JSON](https://gith
...
+Int? maybe_integer = None
+# maybe_integer is not defined
+Int? maybe_numeral = 5
+Int certain_integer = 3
+# maybe_numeral is a defined optional
+is_defined(maybe_integer) # Evaluates to false
+is_defined(maybe_numeral) # Evaluates to true
+maybe_integer == None # Evaluates to true, same as !is_defined(maybe_integer)
+maybe_integer != None # Evaluates to false, same as is_defined(maybe_integer)
+certain_integer == None # This will cause an error. Since None only makes sense when dealing with optionals.
```
yfarjoun:
```suggestion
certain_integer == None # This will cause an error, since None only makes sense when dealing with optionals.
```

versions/development/SPEC.md
```diff
@@ -698,6 +699,48 @@ Pair values can also be specified within the [workflow inputs JSON](https://gith
...
+```
+
+This is useful in a number of cases, for example when an output of a command depends on a certain flag:
+```wdl
+task test
```
yfarjoun:
```suggestion
task test {
```

versions/development/SPEC.md
```diff
@@ -698,6 +699,48 @@ Pair values can also be specified within the [workflow inputs JSON](https://gith
...
+  input {
+    File inputFile
+    String outputPath
+    Boolean log
+  }
+  command {
+    test ~{true="--log" false="" log} inputFile outputPath
+  }
+  output {
+    File? logFile = if log then "test.log" else None
```
yfarjoun: is this actually needed? what happens now if you just have 
```wdl
  File? logFile = "test.log"
```

rhpvorderman: Then `logFile` will always be `test.log` even if there is no `test.log` created.

versions/development/SPEC.md
rhpvorderman: That is more clear indeed. I have added this

versions/development/SPEC.md
rhpvorderman: updated

versions/development/SPEC.md
rhpvorderman: Thanks for pointing this out.

versions/development/SPEC.md
pshapiro4broad: Wrapping seems imprecise anyway since this is an implicit type conversion, e.g. "any expression of a non-optional type can be implicitly convert to an optional version of that type."

versions/development/SPEC.md
```diff
@@ -698,6 +699,48 @@ Pair values can also be specified within the [workflow inputs JSON](https://gith
 }
 ```
 
+### Optional literals
+
+Any non-optional value can be  turned into an optional value by assigning it to an optional variable:
```
pshapiro4broad: I think the distinction that's important to make here is that the value doesn't change, but the type of the value does change.

rhpvorderman: Added this

versions/development/SPEC.md
```diff
@@ -3188,10 +3232,33 @@ Array[String] env2_param = prefix("-f ", env2) # ["-f 1", "-f 2", "-f 3"]
...
+
 ## Array[X] select_all(Array[X?])
 
 Given an array of optional values, `select_all` will select only those elements which are defined.
 
+```WDL
+version 1.0
+workflow SelectFirst {
+  input {
+    Int? maybe_integer = 5
```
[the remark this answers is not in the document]
rhpvorderman: Good suggestion! added this.

versions/development/SPEC.md
```diff
@@ -698,6 +699,51 @@ Pair values can also be specified within the [workflow inputs JSON](https://gith
...
+Int? maybe_five_and_is = 5
+# maybe_five_and_is is a defined optional
+Int certainly_five = 5
+# Certainly five is not an optional
+
+Boolean test_defined = defined(maybe_five_but_is_not) # Evaluates to false
+Boolean test_defined2 = defined(maybe_five_and_is) # Evaluates to true
+Boolean test_is_none = maybe_five_but_is_not == None # Evaluates to true, same as !definedmaybe_five_but_is_not)
+Boolean test_not_none = maybe_five_but_is_not != None # Evaluates to false, same as defined(maybe_five_but_is_not )
+Boolean compare_int_to_none = certainly_five == None # This will cause an error, since None only makes sense when dealing with optionals.
```
yfarjoun: is this a runtime error or a validation error?

yfarjoun: I'm not sure that this should be an error. 

What about 
```wdl
Int? maybe_five = 5
Int five = 5
Boolean compare_five_to_maybe_five = maybe_five == five
```
Should this validate?
I think it should and that the answer should be `True`.

but this means that one should be able to compare an optional with a non-optional.

versions/development/SPEC.md
```diff
@@ -698,6 +699,51 @@ Pair values can also be specified within the [workflow inputs JSON](https://gith
...
+# maybe_five_but_is_not is an undefined optional
+Int? maybe_five_and_is = 5
+# maybe_five_and_is is a defined optional
+Int certainly_five = 5
+# Certainly five is not an optional
+
+Boolean test_defined = defined(maybe_five_but_is_not) # Evaluates to false
+Boolean test_defined2 = defined(maybe_five_and_is) # Evaluates to true
+Boolean test_is_none = maybe_five_but_is_not == None # Evaluates to true, same as !definedmaybe_five_but_is_not)
+Boolean test_not_none = maybe_five_but_is_not != None # Evaluates to false, same as defined(maybe_five_but_is_not )
```
yfarjoun:
```suggestion
Boolean test_not_none = maybe_five_but_is_not != None # Evaluates to false, same as defined(maybe_five_but_is_not)
```

versions/development/SPEC.md
```diff
@@ -698,6 +699,51 @@ Pair values can also be specified within the [workflow inputs JSON](https://gith
...
+This does not change the value, but it changes the type of the value:
+
+```WDL
+Int? maybe_five = 5
+```
+
+The `None` literal is the value that an optional has when it is not defined.
+
+```WDL
+Int? maybe_five_but_is_not = None
```
geoffjentry: Curious what the non-scalawags think - is `None` intuitive here? 

It is to me, but I'm used to Scala's `Option`, which is what this construct was modeled after. 

I don't know that I have any suggestion which seems better (going back to "it is intuitive to me"), perhaps `Empty`?

versions/development/SPEC.md
mlin: I have opened #271 for general discussion of typechecking of optional values (i.e. when use of an optional value may be a static validation error, a runtime error in case actually null, or not an error at all). I think the ambiguities involved are broader than this more-focused PR for adding `None`

versions/development/SPEC.md
rhpvorderman: In python, there is also a [`None`](https://stackoverflow.com/questions/19473185/what-is-a-none-value). It also means something like 'empty' and calling a `None` variable, will give a not defined error. So for people who know python, it can be intuitive (it is for me,  but I also know scala and I can not judge for other people).

pshapiro4broad: I think `null` is a better choice as that's how you write this concept in JSON, which is how WDL inputs are represented. That way the same concept looks the same in both paces.

rhpvorderman: I think assigning `null` to something is a contradiction in itself. So, in my preference `null` is not available as a variable in the toolbox.

As for the "then it will look the same as in JSON" argument. You have a point there, but frankly, `None` is different from `null`. Using the same name for a different concept does not sound like a good idea to me.

pshapiro4broad: > One thing worth thinking through is if there's a meaningful difference between "This space intentionally left blank" and "For unknown reasons there's nothing here" in WDL-land. When those are both viable states I think that null should always mean the latter, and only the latter

Can you explain how these two different states could arise in WDL? They sound to me like a states in the mind of the WDL user, not the WDL interpreter.

pshapiro4broad: > I think assigning null to something is a contradiction in itself. So, in my preference null is not available as a variable in the toolbox.

Remember that, in my proposal, `null` is the name WDL has for the concept of an unassigned value. It has no meaning with respect to the implementation of WDL, or pointers, or other programming language concepts.

> As for the "then it will look the same as in JSON" argument. You have a point there, but frankly, None is different from null. Using the same name for a different concept does not sound like a good idea to me.

Are you saying that using `null` in JSON has a different effect than using (as you propose) `None` in WDL? If so, how is this difference detectable from inside WDL?

In my proposal, `null` is not different than `None` as they are just different spellings of the same concept. Maybe the point that you're trying to make is that `null` has baggage due to its meaning in other programming languages. I agree that it could be a hurdle for WDL users that have experience programming in other languages. Is this the argument that you're making here?

geoffjentry: @pshapiro4broad 

> Can you explain how these two different states could arise in WDL?

I don't have an example, just thought it was worth throwing it out there in case it prompted someone to think of one 

> In my proposal, null is not different than None as they are just different spellings of the same concept.

I don't think they necessarily are, going back to the difference I cited above. For instance in R `NA` and `NULL` are two very difference concepts, but they'd both fold into `null` in many languages which only have the one concept.

> Maybe the point that you're trying to make is that null has baggage due to its meaning in other programming languages. 

I certainly fall under that category (I've protested against adding "the billion dollar mistake" to WDL in the past as well), but to go back earlier in the convo I also admit I'm a fairly biased witness here.

rhpvorderman: @geoffjentry @pshapiro4broad Most of the "Null vs None" stuff was already discussed in #209. Most people liked `None` more, so I settled for that.

rhpvorderman:
What needs to be done to get this PR ready for voting? I would like to get this into the spec at some point. There's been a lot of discussion about the PR, but I suggest we get down to voting this thing. If it is insufficient, it is back to the drawing board for me. But then at least there is a decision being made.

versions/development/SPEC.md
```diff
@@ -689,6 +691,59 @@ Pair values can also be specified within the [workflow inputs JSON](https://gith
...
+  input {
+    File inputFile
+    String outputPath
+    Boolean log
+  }
+  command {
+    test ~{true="--log" false="" log} inputFile outputPath
+  }
+  output {
+    File? logFile = if log then "test.log" else None
```
cjllanwarne: This is the line which I think could be trickier to guarantee. I think for this PR that this example should be removed or reworked to not use `File?`, and a new PR specifically guaranteeing support for `File?` outputs should be added (in case there are alternative ways or syntax options of expressing this intention in an easier-for-cloud-support way)

versions/development/SPEC.md
```diff
@@ -689,6 +691,59 @@ Pair values can also be specified within the [workflow inputs JSON](https://gith
...
+  }
+}
+```
+If `log` is true, the command will evaluate to `test --log inputFile outputPath` and `logFile` will be `test.log`.
+If `log` is false, the command will evaluate to `test inputFile outputPath` and `logFile` will be `None`.
+
+## Document
+
+```
+$document = ($import | $task | $workflow)+
```
yfarjoun: Can you really have more than one workflow in a single document?

yfarjoun: also, why are we making these additions in this PR? seems out of place.

cjllanwarne: This looks like an accidental reversal of https://github.com/openwdl/wdl/pull/270 happened during a rebase to me

cjllanwarne: (or maybe the PR changes jut need another rebase on `master` to pick up that missing PR?)

rhpvorderman: I definitely did not put that in :slightly_smiling_face:.  I will try if a rebase fixes it.

geoffjentry:
@rhpvorderman Sorry for the delay - we can certainly move this to the voting process. Just let me know if you'd still like to go forward as-is, I'm not sure if you've already incorporated what you wanted from the most recent round of comments

versions/development/SPEC.md
```diff
@@ -23,6 +23,11 @@
     * [Map Literals](#map-literals)
     * [Object Literals](#object-literals)
     * [Pair Literals](#pair-literals)
+<<<<<<< HEAD
```
cjllanwarne: Something's still not right here

versions/development/SPEC.md
geoffjentry: @rhpvorderman Heads up that I don't think your rebasing did the trick as this line is still here.

versions/development/SPEC.md
```diff
@@ -23,6 +23,8 @@
     * [Map Literals](#map-literals)
     * [Object Literals](#object-literals)
     * [Pair Literals](#pair-literals)
+    * [Optional Literals](#optional-literals)
+  * [Document](#document)
```
cjllanwarne: Still needs to be deleted

versions/development/SPEC.md
```diff
@@ -689,6 +691,57 @@ Pair values can also be specified within the [workflow inputs JSON](https://gith
...
+  }
+  output {
+    File? logFile = if log then "test.log" else None
+  }
+}
+```
+If `log` is true, the command will evaluate to `test --log inputFile outputPath` and `logFile` will be `test.log`.
+If `log` is false, the command will evaluate to `test inputFile outputPath` and `logFile` will be `None`.
+
+## Document
```
cjllanwarne: Still needs to be deleted

cjllanwarne:
Thanks @rhpvorderman! And FWIW I agree with your meta comment that if something is making iteration on these PRs difficult for contributors, that's definitely worth thinking about to see if there's anything in the git flow to make that easier.

@geoffjentry my concerns are met, happy to vote on this PR now.

versions/development/SPEC.md
mlin: As we move to tie off this PR, perhaps we should punt on this example for now?

rhpvorderman: Agree. Not rushing this in through the backdoor. I will remove this.

geoffjentry:
Sorry for the delay on this, I've been out of office for the last few weeks. Opening for a vote now.

orodeh:
👍

versions/development/SPEC.md
```diff
@@ -689,6 +690,32 @@ Pair values can also be specified within the [workflow inputs JSON](https://gith
...
+Int? maybe_five_but_is_not = None
+# maybe_five_but_is_not is an undefined optional
+Int? maybe_five_and_is = 5
+# maybe_five_and_is is a defined optional
+Int certainly_five = 5
+# Certainly five is not an optional
+
+Boolean test_defined = defined(maybe_five_but_is_not) # Evaluates to false
+Boolean test_defined2 = defined(maybe_five_and_is) # Evaluates to true
+Boolean test_is_none = maybe_five_but_is_not == None # Evaluates to true, same as !definedmaybe_five_but_is_not)
```
mwalker174: Missing a left parenthesis after `defined`.

geoffjentry: Hi @mwalker174 - unfortunately once voting is active votes are on the PR **as-is**. I'd encourage either you or @rhpvorderman file a followup PR after this one is finished

mwalker174:
👍

aednichols:
👍

DavyCats:
:+1:

mlin:
:+1:

cjllanwarne:
👍

cjllanwarne:
@geoffjentry - This PR has been [implemented](https://github.com/broadinstitute/cromwell/pull/4639). No implementation issues came up.  This PR can now be merged into the SPEC.

PR #266, opened 2018-11-16
updated example version string in v 1.0 SPEC.md

delagoya:
Just a small example change to guide new users to properly start out their WDL files. 

Keeping in mind that this is the v 1.0 Specification document, I have a follow-on question about the rest of the paragraph: 

Should all versions quoted in the section be `1.0` ? (I think so, but wanted comments from the group).

### DIFF
--- a/versions/1.0/SPEC.md
+++ b/versions/1.0/SPEC.md
@@ -719,7 +719,7 @@ $document = ($import | $task | $workflow)+
 For portability purposes it is critical that WDL documents be versioned so an engine knows how to process it. From `draft-3` forward, the first line of all WDL files must be a `version` statement, for example
 
 ```wdl
-version draft-3
+version 1.0
 ```
 
 Any WDL files which do not have a `version` field must be treated as `draft-2`.  All WDL files used by a workflow must have the same version.


geoffjentry:
@delagoya re your question at the end, can you give an example which shows something you're talking about? I think I misunderstand your question, but in that specific section of the spec all uses of a version string (i.e. `draft-2` and `draft-3`) outside of the one in the example are correct.

cjllanwarne:
I think any sample WDL in the spec currently using `version draft-3` should probably be updated to `version 1.0` (unless they're specifically talking about some peculiarity of draft-3.

Side comment, we should probably also update the `development` branch to be using `version development`, even though that'll of course need re-updating when a version number is chosen for it.

rhpvorderman:
@cjllanwarne  `version development` is a very good thing. I would not consider it `re-updating` when a version number is chosen for it. A new version would simply be a snapshot of the development version at some date. So if you choose `version development` in your WDL you know you are going to run into issues later because the spec keeps developing.

cjllanwarne:
@rhpvorderman good point, the `development` branch will indeed always want to have `version development`!

cjllanwarne:
@delagoya you may want to rebase this onto master since there's now another example right next to this one:
````md
## Versioning

For portability purposes it is critical that WDL documents be versioned so an engine knows how to process it. From `draft-3` forward, the first non-comment statement of all WDL files must be a `version` statement, for example

```wdl
version draft-3
```
or

```wdl
#Licence header

version draft-3
```
````

delagoya:
Yeah, this is a bike shedding PR, and I am not too tied to it. Considering
removing.

patmagee:
@delagoya I am going to close this following @geoffjentry suggestions. I do not think this example needs to explicitly be updated with each version change as its just showing how to write a version

delagoya:
Thats fine.

PR #269, opened 2018-11-20
Remove the 'state of the specification' section

cjllanwarne:
In my opinion, given that the SPEC has progressed a lot since August 2015, this section is more misleading than useful.


geoffjentry:
Couldn't agree more. Nearly did this myself the other day

geoffjentry:
Merging as-is as this is purely textual cleanup and does not affect the specification itself

PR #270, opened 2018-11-20
Remove 'Document' section

cjllanwarne:
This section doesn't seem to add much other than details on how the hermes grammar implementation works.

It's also missing the `version` statement which literally appears in the very next paragraph.


geoffjentry:
Merging this as-is as it's just a spec cleanup and doesn't affect the spec itself. This is an artifact of hermes

PR #273, opened 2018-11-29
[Hermes Grammar] Removal of Object (and Uri)

cjllanwarne:
To satisfy #228, and because Uri was *never* a type in the spec 🤦‍♂️


patmagee:
@cjllanwarne is this ready for voting?

cjllanwarne:
@patmagee this isn't a spec change per se, it's an implementation detail of #283 so shouldn't need voting

patmagee:
@cjllanwarne sounds good to me!

Issue #201, opened 2018-03-13
Struct Literals


cjllanwarne:
Especially when working with structs, needing to type in `object` at struct-construction time is slightly counter-intuitive.

As I was writing some of these out in test cases, I was thinking about more intuitive struct literals. Here are a few options:

* They all use `A` as the constructor indicator instead of `object`.
* 1 and 2 use `{}`, 3 and 4 `()`
* 1 and 3 use variable names (eg `i: ..., f:...`), 2 and 4 do not.

```wdl
struct A {
  Int i
  Float f
}
workflow foo {
  # Today's syntax:
  A a = object { i: 5, f: 5.5 }

  # Proposal 1:
  A a = A { i: 5, f: 5.5 }
  # Proposal 2:
  A a = A { 5, 5.5 }

  # Proposal 3:
  A a = A(i: 5, f: 5.5)
  # Proposal 4:
  A a = A(5, 5.5)
}
```

Any other thoughts?

---
Side note: why do we need anything at all before the `{` at all? Because otherwise the expression evaluates as a map literal and the key is actually an expression eg:
```wdl
String i = "i"
String f = "g"

# oops, the 'f' key is actually "g", but we can't tell that until runtime:
A a = { i: 5, f: 5.5 }
```
NB: I think this side-note could lead to unexpected errors for authors. In the winstanley IDE I intend to flag the above pattern (ie a Struct value being assigned from a map literal) as a warning with the suggestion being "use an object/struct literal instead".

orodeh:
@cjllanwarne I am for 1 or 2.

orodeh:
Having chewed on it some more, I think braces ```(,)``` would be similar to object construction in standard programming languages.  So, proposals 3 or 4. Another possibility: 
```
# Proposal 1+2+3+4
A a = A( {i: 4, f: 5.5} )
```

We might want to allow all (or some) of these syntaxes. One way or the other, I suggest sticking to a known constructor syntax.

patmagee:
Good suggestions @cjllanwarne. I like proposal 3 and proposal 1. leading with the property you are setting is clearer to me in this case and allows some flexibility in terms of order when creating a struct. Additionally, I think these work better for complex type, since it becomes clearer what property is being set. Whereas leaving out the property it looks a bit muddled imo.

```wdl
struct A {
  Array[Pair[String,File]] complex
  Int i
}

workflow foo {
  #Proposal 1
[...]

PR #278, opened 2018-12-14
Struct literals

patmagee:
Currently, there is no such thing as a `Struct Literal`, instead we use the `Object literal` notation and hope that the engine conversion does not mess up any of the types. On top of `objects` being deprecated in the near future, there is also the added problem that there could be a loss of information if an object literal is the first cast to an object, before being cast to the declaring type.

To address this, I propose the idea of `Struct Literals`, which as previously discussed here #201 

I chose to go with a simple annotation for defining struct literals that should both look familiar and also not look too ugly when implementing:

```wdl
Person a = Person( age: 30, name: "Bill" )
```

### DIFF
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -2088,7 +2088,7 @@ Arugments placed within the parethesis are key-value pairs, where the key is the
 There is no need to wrap a key in quotation marks, instead keys are represented in plain text followed by an equals sign `=`. The value follows after the `=`. Multiple arguments can 
 be separated by a comma ',' and arguments do not need to be specified in a specific order.
 
-Values passed to struct literals can be any previously defined declaration, or they themselves can be a literal notation. 
+Values passed to struct literals can be any previously defined declaration, or they themselves can be a literal notation.
 
 
 ```wdl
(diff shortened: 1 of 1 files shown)


rhpvorderman:
> ```wdl
> Person a = Person( age: 30, name: "Bill" )
> ```

This implements the JSON way of assigning values using `:`.  How about using `=` which is used troughout WDL?
```WDL
Person a = Person(age=30, name="Bill", tags=["tag1", tag2], somedict={"key": "value"} )
```
Note that is just a matter of preference. I feel `=` is slightly better than `:` because of consistency with the rest of the spec, but I am happy with either of them.

patmagee:
@rhpvorderman that was actually my initial instinct, partially because it's the "pythonic" way of doing things. I was unsure though if it was too far a departure from the object literal notation, which uses the colons. 

We can see what others suggest, but I am impartial either way as well

cjllanwarne:
I like this a lot.

- My preference (but honestly, not particularly strong) would be an direct replacement of the `object { ... }` syntax with literals which have `object`replaced by the struct name.
  - in other words, `Person p = Person { name: "Bob", age: 55 }`
  - This just feels like the least difference from what we have today, and the least editing of existing WDL files to bring them into line. 
- I do this we should go through and remove any reference to `= object { ... }` style assignments.
- I'm pretty close to an implementation of the "Remove Objects" PR... it might be worth waiting for that to merge (and rebase this onto it) before voting, as I would anticipate a fair number of potential merge conflicts

rhpvorderman:
@mlin I have removed the example that related to the typechecking.

patmagee:
I'm curious to hear what people think about this before making any changes. The old way that `objects` work (ie `object { .... }` ) is one potential method, but it seems a bit like an outlier in the world of constructors to me.  To Paraphrase  @geoffjentry  "When in doubt do what python does" was a bit of my motto. 

 Additionally, I agree with @rhpvorderman in regards to the consistency within the WDL of the usage of `=` instead of `:`. It always throws me off when we switch between the two (as is the case in the `runtime` block). 

As for waiting for the removal of objects, I am open for that, but I do not think we should remove objects and release without the proper struct literal in place to create struct types

patmagee:
@cjllanwarne ah thats what you meant in regards to waiting for the Object PR to close. I assumed you were completely removing the `object` literal with your PR as well

PR #282, opened 2019-01-18
Removed old Broad Scripts from repo

patmagee:
There were a number of scripts that were artifacts from when this repo was owned by the folks at the broad. This PR simply removes those scripts with a note that users should look to dockstore for more recently published ones.

**Reasons for removal:**

- Old and unmaintained. The WDL specification has advanced significantly and these scripts have not changed.
- The scripts themselves highlight outdated and unmaintained pipelines and should not be endorsed by within this repo
- The scripts were very GATK specific
- The broad has recently released many of their production pipelines in dockstore, and users should look there for updated version.

It may be beneficial to have a set of `Tutorial` WDL's which could be published on Dockstore. There are already a number of `hello-world` workflows so whether this is necessary is up for a debate.

@vdauwera @geoffjentry can you confirm that none of these files are needed? if they are, would it be possible to move them to a separate Broad owned repo, and then pass this PR?


geoffjentry:
As this is not a spec change it doesn't need to go through the voting process. Merging now, thanks @patmagee

PR #284, opened 2019-02-12
[Hermes Grammar] Add none literal

cjllanwarne:
- Implementation detail for #263


patmagee:
@cjllanwarne does this need to be voted on?

cjllanwarne:
@patmagee I don't think so, although technical review is always welcome. 

It's just an implementation detail of #263

patmagee:
to be honest I am not expert at hermes, and it LGTM? Maybe if @geoffjentry can weigh in or someone else just to make sure that everything looks good, then we can merge this

geoffjentry:
Yeah, we don't vote on the hermes grammar changes as they're not part of the spec. will review

PR #286, opened 2019-02-13
Struct literal - Revist

patmagee:
I have modified my original `Struct Literal` PR after rebasing on master once the object removal PR had been merged. I have removed all references to `object literal` notation and replaced it with `struct literal` as requested by @cjllanwarne .

## Changes 

Currently, there is no such thing as a `Struct Literal`, instead we use the `Object literal` notation and hope that the engine conversion does not mess up any of the types. On top of `objects` being deprecated in the near future, there is also the added problem that there could be a loss of information if an object literal is the first cast to an object, before being cast to the declaring type.

To address this, I propose the idea of `Struct Literals`, which as previously discussed here #201

I chose to go with a simple annotation for defining struct literals that should both look familiar and also not look too ugly when implementing (since it is based on python constructors)

Here is an example:

```wdl

struct Person {
  Int age
  String name
}

...

Person a = Person( age=30, name="Bill" )
```

## Open Questions for debate:

1. Recent changes to the spec introduced the ability to use Map Literals to create a struct or allow map literals to be cast to a struct. My question is, with the implementation of struct literals, should this be allowed or encouraged? To me it makes sense to allow casting of a map to specific Struct type, but what do you guys think

### DIFF
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -75,9 +75,9 @@ Table of Contents
     * [Struct Declarations](#struct-declarations)
       * [Optional and non Empty Struct Values](#optional-and-non-empty-struct-values)
     * [Using a Struct](#using-a-struct)
-      * [Struct Assignment from Map Literal](#struct-assignment-from-map-literal)
-    * [Struct Member Access](#struct-member-access)
-    * [Importing Structs](#importing-structs)
+      * [Struct Literals](#struct-literals)
+      * [Struct Member Access](#struct-member-access)
+      * [Importing Structs](#importing-structs)
 * [Namespaces](#namespaces)
 * [Scope](#scope)
 * [Optional Parameters &amp; Type Constraints](#optional-parameters--type-constraints)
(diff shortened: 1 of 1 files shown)


cjllanwarne:
Re the open question - IMO I think we can state that it's possible if a map happens to be (a) a valid WDL map and (b) the right shape, but in our examples we should always use struct literals to demonstrate creation of structs

patmagee:
@cjllanwarne I agree, I changed a few map literal casts to simply structs

DavyCats:
:+1:
I did spot one typo, but I guess I'm too late with this now :stuck_out_tongue: 
It isn't a big deal anyway:
`[...] which enforces the typing of all of is assigned parameters.` on line 2096/2097 should be
 `[...] which enforces the typing of all of its assigned parameters`.

geoffjentry:
:+1:

orodeh:
👍

aednichols:
👍 

We should fix `Person(name="Harry", age=11}` soon after; the other examples and prose description make the typo not a blocker.

versions/development/SPEC.md
```diff
@@ -2102,14 +2084,54 @@ task myTask {
 }
 
 workflow myWorkflow {
-    Person harry = object {name: "Harry", age: 11}
+    Person harry = Person(name="Harry", age=11}
```
mlin: `s/}/)/`

geoffjentry:
@patmagee considering some of our more eagle eye voters are starting to pick up on typos and such, it might be worth pulling this back and then re-launching the vote

versions/development/SPEC.md
```diff
@@ -2102,14 +2084,54 @@ task myTask {
...
+be separated by a comma ',' and arguments do not need to be specified in a specific order.
+
+Values passed to struct literals can be any previously defined declaration, or they themselves can be a literal notation.
+
+
+```wdl
+
+#Simple case
+File fastq_1
+File fastq_2
```
cjllanwarne: This is also a typo presumably? I think it should either be (a) assigned a value (even `File fastq2 = ...` as a placeholder) or put into an `input { }` section

cjllanwarne:
One more potential typo for a fixup, but otherwise this looks great - and thanks for taking care of those option literals! 👍

geoffjentry:
@patmagee Voting has passed however it's clear that there are multiple typos here. Would you rather go ahead and merge in & followup w/ a cleanup PR, or would you rather clean it up & resubmit?

patmagee:
@geoffjentry I am notorious for my typos, this is quite bad. I personally think I should create a separate PR, closing this one, and we can merge that one. I would rather not include any errors that might be missed

PR #288, opened 2019-02-14
`~{ }` use commentary in SPEC

infispiel:
Added `~{ }` commentary to the appropriate version of the SPEC.md; I am sorry for editing the wrong one last time around.

Related issue is: [https://github.com/openwdl/wdl/pull/url](url)

### DIFF
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -997,7 +997,7 @@ task heredoc {
 
   command<<<
   python <<CODE
-    with open("${in}") as fp:
+    with open("~{in}") as fp:
       for line in fp:
         if not line.startswith('#'):
           print(line.strip())
@@ -1006,7 +1006,7 @@ task heredoc {
 }
 ```
 
-Parsing of this command should be the same as the prior section describes.
+Parsing of this command should be the same as the prior section describes.  As noted earlier in the [Expression Placeholders](#command-parts) section, it is important to remember that string interpolation within `<<<`...`>>>` blocks must be done using the syntax `~{expression}`.
 
 #### Stripping Leading Whitespace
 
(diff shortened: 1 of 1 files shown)


cjllanwarne:
Thanks - FWIW I think this is such an obviously good change that it can go straight to voting (or even better, just merged as-is)

patmagee:
This looks great too me. Its a good and clear clarification that I think is needed (especially with any confusion around the placeholders)

geoffjentry:
@cjllanwarne IMO this is borderline regarding voting. I **personally** don't think it's necessary, but it's close enough that I don't want to set a precedent. I'd rather keep subjectivity out of the process as much as possible

patmagee:
@geoffjentry I agree this is borderline. I think to err on the side of caution (and not start any new trends) we should just put this through voting, which if everyone is happy, we can lock this where it is and start voting on it right now

geoffjentry:
Thanks @infispiel ! I've opened this issue up for voting

Speaking of, I'm voting 👍

patmagee:
👍

mlin:
:+1: on this, just two notes for visibility (related to but not conflicting with this diff)

1. The "String Interpolation" section begins,

> Within tasks, any string literal can use string interpolation to access the value of any of the task's inputs.

I think it's an oversight to discuss tasks but not workflows.

2. It's left unclear whether `sep`, `default`, and `true/false`are usable in interpolations outside of commands.

geoffjentry:
@mlin Remember that voting is as-is. It's possible that I was a bit overeager on calling this to a vote. Would you prefer we pull back for now or maybe we can have a followup PR

mlin:
@geoffjentry yea my vote is :+1:. the rest was commentary

geoffjentry:
@mlin perfecto

orodeh:
👍

cjllanwarne:
👍

DavyCats:
:+1:

geoffjentry:
Vote passes unanimously. Waiting for implementation

cjllanwarne:
Unless anyone disagrees(?), I think this can be merged as implemented since it's more of a clarification than an active change to behavior

geoffjentry:
Good point

Issue #271, opened 2018-11-26
Clarifying typechecking of optional values


mlin:
Spec gray area:

If I have an optional declaration `T? x`, and an expression refers to `x` in some context where (non-optional) `T` is expected, when should this be statically rejected as a type mismatch error, vs. considered valid albeit liable to fail at runtime?

The spec right now says `x` "can only be used in calls or functions that accept optional values" but this isn't quite clear to the above.

My "first principles" instinct would be that `T? x` should statically fail to unify with (non-optional) `T` absent some explicit coercion, but:

- It's expressly valid and desirable for interpolations in task commands, as described in "Prepending a String to an Optional Parameter." Also relevant #227
- If one writes another declaration `T y = x` what should happen when?
- I infer from some examples in the wild that Cromwell accepts the pattern like `T y = if defined(x) then EXPR_WITH_x else SOME_DEFAULT` ([ex1](https://github.com/gatk-workflows/gatk4-somatic-snvs-indels/blob/28132cbec2fc9178e50cd3a36aa8f94023619428/mutect2.wdl#L469), [ex2](https://github.com/HumanCellAtlas/skylab/blob/29946a3a71f80d250d35e9ab8711d06ca18894b0/library/tasks/Attach10xBarcodes.wdl#L12)).
  - In those examples, `EXPR_WITH_x` is `x * 1000` and `size(x)` respectively.
  - Do we check the optional quantifier statically at all?
  - Alternatively, one could imagine special-case typechecking logic for the `if defined(x) ...` case.
- If one specifies `x` to a required call input of type `T` what should happen when? ([ex](https://github.com/DataBiosphere/topmed-workflows/blob/31ba8a714b36ada929044f2ba3d130936e6c740e/variant-caller/variant-caller-wdl/calculate_contamination.wdl#L74))

Appreciate if someone from Cromwell team could clarify how it approaches this (and of course comment on what would be ideal for the spec). Thanks!

geoffjentry:
This feels an awful lot like a discussion @yfarjoun raised in #263 - noting that here, if I didn't just overaggressively pattern match it'd be useful to unify the conversations.

Also tagging @cjllanwarne to take a first stab at this from the cromwellian perspective

cjllanwarne:
This does indeed come up in discussions a lot!

[...]

Issue #187, opened 2018-01-29
Ease optional access with scatter-like block


cjllanwarne:
Port of Cromwell issue https://github.com/broadinstitute/cromwell/issues/1852

Equivalent to scatter blocks for arrays, I propose a new ifDefined block for optionals:
```wdl
workflow foo {
  String? bar
  ifDefined (concreteBar in bar) {
    call baz { input: x = concreteBar }
  }
  output {
    # Just like outputs from inside a scatter become array outputs, isDefined creates optional outputs:
    String? baz_out = baz.out
  }
}

task baz {
  # Non-optional input:
  String x
  # ...
}
```

orodeh:
I can see the issue, but I can't say I'm crazy about the syntax. Also, I assume the `String` in `isDefined` is superfluous. 
```
ifDefined (concreteBar in bar)
```

geoffjentry:
@cjllanwarne Could you provide context on use cases and such? In general it'd be helpful for proposed language constructs to come attached with some concrete examples of "why"

cjllanwarne:
Sure!

The idea here is to provide syntactic sugar for the boilerplate in the following scenario:
```wdl
workflow foo {
  [...]
  Int? x = ...
  if(defined(x)) {
    Int real_x = select_first([x])
    call bar { input: x = real_x }
  }
}

task bar {
  Int x
  [...]
}
```

This construction is becoming more and more common in WDLs as people start using `if {}` blocks more and getting more and more optional values, and at the same time, Cromwell 30 fixed a bug in WDL parsing and now enforces that optional values cannot be passed into tasks which expect a non-optional input.

cjllanwarne:
In my opinion the `Int real_x = select_first([x])` is an awkward and hard to read line that presents a barrier to entry for WDL readers. The more we can let authors write what they want without having to add awkward hoops, the better.

patmagee:
Non Deterministic workflows (although already supported) is one thing that I always struggle with, as it defeats the underlying purpose of reproducibility. I always wonder If this type of construct is occurring more and more, whether its something that we should encourage, or try to steer users away from. 

That being said if this is something everyone wants, then I will entertain the idea.
When it comes to the specific proposed solution I am with @orodeh on this. I am not really a huge fan of the syntax. the `in` keyterm is more strongly associated with iterable types and it a bit confusing. 

I would not add any additional constructs or execution blocks. I believe this can be solved by properly employing current constructs then adding a better `select` engine function.

[...]

PR #290, opened 2019-02-19
Clarify optional and nonempty type constraints

mlin:
Following discussion on issue #271, this PR is a straw-man resolution to current ambiguity over how/when the optional and nonempty type quantifiers are "enforced." It will definitely need debate and discussion, so fire away!

A couple of comments to start:

* The `select_first` idiom is workable but not exactly loved; there's some discussion on #187 about nicer sugar which could follow on.
* On the other hand I don't think there's an idiomatic way to coerce `Array[T]` to `Array[T]+` right now, which is needed if we want to enforce nonempty in the same way we propose to statically validate optionals. (And it would be a little weird to enforce one of them statically and the other only at runtime, right?)

### DIFF
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -2327,17 +2327,15 @@ Then the command would be instantiated as:
 /bin/mycmd /path/to/c.txt
 ```
 
-In general, it's invalid to supply an expression of optional type `T?` for a non-optional input or variable of type `T`, as the latter cannot accept `None`. (The idiom `select_first([expr, default])` coerces `expr : T?` to `T` by substituting `default : T` when `expr` is undefined; if `default` is omitted, then the coercion fails at runtime.) Similarly, an `Array[T]` expression does not satisfy an `Array[T]+` input.
+It's invalid (a static validation error) to supply an expression of optional type `T?` for a non-optional input or variable of type `T`, as the latter cannot accept `None`. (The idiom `select_first([expr, default])` coerces `expr : T?` to `T` by substituting `default : T` when `expr` is undefined; if `default` is omitted, then the coercion fails at runtime.) This constraint propagates into compound types, so for example an `Array[T?]` expression doesn't satisfy an `Array[T]` input. It also applies to function arguments, with exceptions for string concatenation, described below, and the equality/inequality operators `==` and `!=`, which can compare two values of types differing only in their quantifiers, considering `None` equal to itself but no other value.
 
-These constraints also apply to function arguments, with exceptions for string concatenation, described below, and the equality/inequality operators `==` and `!=`, which can compare two values of types differing only in their quantifiers, considering `None` equal to itself but no other value.
-
-Quantifiers may apply within compound types; for example, if `a : Array[T?]+`, then `a[0] : T?`. An array type can be both non-empty and optional, `Array[T]+?`, such that the value can be `None` or a non-empty array, but not the empty array. (Reliance on this subtle distinction is discouraged, but it arises from other constructs such as a conditional call with an `Array[T]+` output.)
+The nonempty array quantifier `Array[T]+` isn't statically validated in the same way, but rather as a runtime assertion: binding an empty array to an `Array[T]+` input or function argument is a runtime error. An array type can be both non-empty and optional, `Array[T]+?`, such that the value can be `None` or a non-empty array, but not the empty array.
 
 ## Interpolating and concatenating optional strings
 
 Interpolations with `~{}` and `${}` accept optional string expressions and substitute the empty string for `None` at runtime.
 
-String concatenation with the `+` operator has special typing properties. When applied to two non-optional operands, the result is a non-optional `String`. However, if either operand has an optional type, then the concatenation has type `String?`, and the runtime result is `None` if either operand is `None`. These unusual semantics facilitate command-line flag interpolation. Consider this task:
+Within interpolations, string concatenation with the `+` operator has special typing properties to facilitate formulation of command-line flags. When applied to two non-optional operands, the result is a non-optional `String`. However, if either operand has an optional type, then the concatenation has type `String?`, and the runtime result is `None` if either operand is `None`. To illustrate how this can be used, consider this task:
 
 ```wdl
 task test {
@@ -2367,6 +2365,8 @@ The latter case is very likely an error case, and this `--val=` part should be l
 ```
 python script.py ${"--val=" + val}
 ```
+
+The `+` operator cannot accept optional operands outside of interpolations.
 
 # Scatter / Gather
 
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -2266,7 +2266,7 @@ workflow wf {
 
 [Types](#types) can be optionally suffixed with a `?` or `+` in certain cases.
 
-* `?` means that the parameter is optional.  A user does not need to specify a value for the parameter in order to satisfy all the inputs to the workflow.
+* `?` means that the parameter is optional, potentially `None` at runtime. A user does not need to specify a value for the parameter in order to satisfy all the inputs to the workflow.
 * `+` applies only to `Array` types and it represents a constraint that the `Array` value must contain one-or-more elements.
 
 ```wdl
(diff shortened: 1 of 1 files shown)


versions/development/SPEC.md
```diff
@@ -2327,9 +2327,17 @@ Then the command would be instantiated as:
...
+In general, it's invalid to supply an expression of optional type `T?` for a non-optional input or variable of type `T`, as the latter cannot accept `None`. (The idiom `select_first([expr, default])` coerces `expr : T?` to `T` by substituting `default : T` when `expr` is undefined; if `default` is omitted, then the coercion fails at runtime.) Similarly, an `Array[T]` expression does not satisfy an `Array[T]+` input.
 
-Sometimes, optional parameters need a string prefix.  Consider this task:
+These constraints also apply to function arguments, with exceptions for string concatenation, described below, and the equality/inequality operators `==` and `!=`, which can compare two values of types differing only in their quantifiers, considering `None` equal to itself but no other value.
+
+Quantifiers may apply within compound types; for example, if `a : Array[T?]+`, then `a[0] : T?`. An array type can be both non-empty and optional, `Array[T]+?`, such that the value can be `None` or a non-empty array, but not the empty array. (Reliance on this subtle distinction is discouraged, but it arises from other constructs such as a conditional call with an `Array[T]+` output.)
```
DavyCats: I don't think it's valid to discourage using `Array[T]+?`, consider you might have an optional input which may occur multiple times (eg. giving multiple bed/interval files to a variantcaller). It's optional, thus the `?`, and giving an empty array might result in an invalid command.

eg:
There are two valid ways of calling a script: `someScript` or `somScript -i 1.txt -i 2.txt ...`
The command block might looks like this: `someScript ~{true="-i " false="" defined(inputs)}~{sep="-i " inputs}`
- Using `Array[File]` or `Array[File]+` would not allow for the first option.
- Using `Array[File]?` would allow for an empty array which would prduce the invalid command `someScript -i`
- So your only option in to use `Array[File]+?`.

versions/development/SPEC.md
```diff
@@ -2327,9 +2327,17 @@ Then the command would be instantiated as:
...
+
+## Interpolating and concatenating optional strings
+
+Interpolations with `~{}` and `${}` accept optional string expressions and substitute the empty string for `None` at runtime.
+
+String concatenation with the `+` operator has special typing properties. When applied to two non-optional operands, the result is a non-optional `String`. However, if either operand has an optional type, then the concatenation has type `String?`, and the runtime result is `None` if either operand is `None`. These unusual semantics facilitate command-line flag interpolation. Consider this task:
```
DavyCats: :+1:

versions/development/SPEC.md
```diff
@@ -2327,9 +2327,17 @@ Then the command would be instantiated as:
 /bin/mycmd /path/to/c.txt
 ```
 
-## Prepending a String to an Optional Parameter
+In general, it's invalid to supply an expression of optional type `T?` for a non-optional input or variable of type `T`, as the latter cannot accept `None`. (The idiom `select_first([expr, default])` coerces `expr : T?` to `T` by substituting `default : T` when `expr` is undefined; if `default` is omitted, then the coercion fails at runtime.) Similarly, an `Array[T]` expression does not satisfy an `Array[T]+` input.
```
DavyCats: A way of coercing `Array[T]` to `Array[T]+` will need to be added.

mlin: Agreed, it will be problematic to statically enforce `+` without also adding such a coercion. We can also contemplate leaving `+` as more of a / runtime assertion, with the downside of nonuniformly treating `?` and `+`. Discussion on this welcome. cc @orodeh

versions/development/SPEC.md
mlin: Sure, I think its fair to say the parentheses statement is more of my editorial opinion and can be stricken.

pshapiro4broad: Is the only valid form `Array[T]+?` or can it also be written `Array[T]?+`? The spec isn't explicit about this as far as I can see.

versions/development/SPEC.md
orodeh: @mlin, thanks for straightening out these issues. 

My personal opinion is that having `+` as a non-empty array qualifier is fine, but I prefer that it's implementation be done as a runtime assertion. If it is part of the type system, then we run into a combinatorial explosion of types. For example:

```
Array[String], Array[String]+
Array[Array[String]], Array[Array[String]]+, Array[Array[String]+], Array[Array[String]+]+
```

However, I would argue that most people find two types sufficient here:  `Array[String]` and `Array[Array[String]]`.

versions/development/SPEC.md
orodeh: This was not obvious to me, thank you for pointing it out. For ease of reference, this is the task:

```wdl
task test {
  input {
    String? val
  }
  command {
    python script.py ${"--val=" + val}
  }
}
```

I would much rather not overload the `+` operator to handle optional strings in a special way. Perhaps the die is already cast, because `${"--val=" + val}` is a much loved, common idiom. If at all possible, I would rather use a different symbol than `+`, to avoid overloading.

cjllanwarne: This is technically new, FWIW - right now it's *only* permissible as a special case in string interpolation, and an error elsewhere.

versions/development/SPEC.md
cjllanwarne: I think that technically you can only apply `+` to arrays. Thus, since `Array[T]?` is not an array (it's an optional), the form `Array[T]?+` would be invalid

versions/development/SPEC.md
cjllanwarne: I don't mind (/actively like) the fact that you can nest this as a type constraint inside other types. It makes the logic much easier to implement, at least...

I agree though that it should be a coercion (and thus enforced via a runtime check) to go from `Array[X]` to `Array[X]+`. Anything else (eg engine functions) is going to be a nightmare to apply correctly in the more complex cases here

mlin: @cjllanwarne just clarifying -- do you want to see it enforced as a runtime assertion *rather than* statically? (I'm on the fence at this point)

versions/development/SPEC.md
mlin: I'm fine with the special case although I speculate it's actually a bit trickier to implement, propagating the flag of whether we're inside an interpolation or not when typechecking expressions...

mlin:
@cjllanwarne @orodeh @DavyCats @pshapiro4broad to try the feedback on for size, I changed nonempty-array to be a runtime-only check, constrained the special string concatenation behavior to apply inside interpolations only, (and struck my editorial comment about `Array[T]+?`). I am personally on the fence about these changes (could change them back), the nice thing will be just to get to something unambiguous. See what you think!

orodeh:
Looks good to me. I am still not a fan of allowing `+` inside interpolation to take a combination of `String?` and `String` types.

pshapiro4broad:
@mlin This looks good to me. As you've done here, it would be nice to make the spec explicit everywhere about run-time vs compile-time behaviors; you shouldn't have to define terminology here for that. That would something for another PR though.

> I am still not a fan of allowing `+` inside interpolation to take a combination of `String?` and `String` types.

I agree that it's messy but am not sure what a better solution would be. Maybe a different syntax for string interpolation that makes it clear that a list of fields must all not `None` for the interpolation to be evaluated.

versions/development/SPEC.md
patmagee: This also seems to open the Pandora's box for operators (although that might not be a bad thing), IE should we allow the same logic to be applied to other Types and operators? 
```wdl
Int? one
Int? two

Int? subtract = one - two
Int? add = one + two
Int? multiply = one * two
Int? all = ((one * two) + one) - two
```
To me, it seems like it could be a bad idea to overload the `+` operator and not do the same overloading on all other operators. I think it will likely cause confusion, and I am sure people will inevitably ask for it if in other contexts if we introduce it to string concatenation.

versions/development/SPEC.md
patmagee: `+` to me has always seemed like a strange type. To me its a more of a runtime decorator that is an assertion check as opposed to a `Type` check.  I for one do not even mind if it is removed from the specification entirely, however that likely would cause people to raise their pitchforks in protest. 

Realistically, the only guarantee that `Array[X]+` provides is that `Array[0]` will not be `None`. This is not a strong enough argument to assert that it is a separate type completely IMO, and it SHOULD be treated as a normal `Array[]` type. Ie the following should be valid with no need to coercion:
```wdl
Array[X]+ myArray
Array[X] otherArray = myArray
```

The one thing we need to be cognisant of however, is that be relegating the `+` type to a `decorator` and not a type, we essentially are introducing a new concept to WDL. That of `Decorator` expressions, which serve as runtime `hints` that should be enforced by the engine, but should not change the typeing of the variable they are on

geoffjentry:
@mlin where does this stand?

mlin:
@geoffjentry I think it's ready for a 'last call' for comments before voting -- I'm happy to address anything that comes in

pshapiro4broad:
+1

orodeh:
👍

TMiguelT:
:+1:

geoffjentry:
The ayes have it and this is moving to `waiting for implementation`. I'll lean on @cjllanwarne and @mlin to speak up when that point has been hit.

geoffjentry:
Cromwell conforms to this behavior, making it implemented. Merging

PR #292, opened 2019-02-24
Remove `while(expr)` scope from specs

azzaea:
From [here](https://github.com/openwdl/wdl/pull/232) it seems  `while` blocks are not allowed in wdl version 1.0

### DIFF
--- a/versions/1.0/SPEC.md
+++ b/versions/1.0/SPEC.md
@@ -2113,7 +2113,6 @@ Scopes are defined as:
 
 * `workflow {...}` blocks
 * `call` blocks
-* `while(expr) {...}` blocks
 * `if(expr) {...}` blocks
 * `scatter(x in y) {...}` blocks
 


geoffjentry:
This is another one of those borderline cases where there's no **real** reason it should be voted upon, but it **is** changing the spec.

I'm going to make the executive decision to go ahead and merge this as a previous vote already removed the `while` construct from the language, and this is obviously just a dangling bit which wasn't caught in that process. Further, unlike #288 it's referencing syntax which is currently impossible.

Thanks @azzaea !

geoffjentry:
Oof, for those who saw my previous comment (including @azzaea who likely saw it in notifications) - please disregard. I thought it seemed weird that that was still hanging around.

I'm closing this as it's against the `1.0` version of the spec, which is currently viewed as immutable (warts and all). You want to be making any edits to the [development](https://github.com/openwdl/wdl/blob/master/versions/development/SPEC.md) version, which has already incorporated this fix.

Thanks for the eagle eyes however, @azzaea

Issue #289, opened 2019-02-18
File Bundles and Secondary / Accessory Files


illusional:
Dear WDL team!

I was hoping to source the community opinion on file bundles, or CWL's secondary files. This issue was spawned from: https://github.com/broadinstitute/cromwell/issues/2269 and also within the [WDL forum](https://gatkforums.broadinstitute.org/wdl/discussion/9299/secondary-index-files-and-directories-in-wdl) where there was general support for accessory files.

My main input is that I'd like the associated files to be attached within a explicit type, and not annotated on a specific task. Even though it might be easy to determine at runtime whether the list of index files is different, I think it would be clearer to have a bundle type (that might sit alongside the [same importing rules as structs](https://github.com/openwdl/wdl/pull/160#issuecomment-345261117)). 

CWL annotates each input with a list of [secondaryFile](https://www.commonwl.org/v1.0/CommandLineTool.html#CommandInputParameter)s that uses a simple query syntax:
> 1. If string begins with one or more caret ^ characters, for each caret, remove the last file extension from the path (the last period . and all following characters). If there are no file extensions, the path is unchanged.
> 2. Append the remainder of the string to the end of the file path. 

(They also allow for an expression that should resolve to a filename or an array of files) 

## My suggestions
To take with a grain salt, I've got a few suggestions to get the ball rolling. I'll use a modified indexed BAM, that has three files ($base.bam, $base.bam.bai, $base.txt) to show how the examples hold up:

1. Create a `bundle` that has an implicit base file type (like CWL), and references secondary files using this base file with a secondary files selector syntax, probably the same as CWL for consistency and as far as I know it does the job. If coerced into a File, it should just resolve to the base file (in the following case, a Bam).

```
bundle NamedModifiedIndexBam {
  bai = ".bai"
  txt = "^.txt"
}
```
OR (just two suggestions for the same thing, not that WDL should accept both)
```
bundle AnonymousModifiedIndexBam = [".bai", "^.txt"]
```

2. Create a more explicit bundle that has a basename, and must explicitly reference each _associated_ file. It might be a good idea to have a _base_ property, that would be the resolver when passing to the command line, or potential coersion into a file. Benefits are it doesn't require the query language and makes it clearer.

```
bundle ExplicitModifiedIndexBam {
  bam = ".bam"	# or base = ".bam"
  bai = ".bam.bai"
  txt = ".txt"
}
```

Then you could reference them in the same way you do primitives or structs:

```wdl
task my_task {
[...]

PR #297, opened 2019-03-27
Struct literal Revisit - Revisit

patmagee:
I have fixed the identified typos. @DavyCats @cjllanwarne @mlin @geoffjentry please feel free to scrutinize this before it goes to voting again!

### DIFF
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -75,9 +75,9 @@ Table of Contents
     * [Struct Declarations](#struct-declarations)
       * [Optional and non Empty Struct Values](#optional-and-non-empty-struct-values)
     * [Using a Struct](#using-a-struct)
-      * [Struct Assignment from Map Literal](#struct-assignment-from-map-literal)
-    * [Struct Member Access](#struct-member-access)
-    * [Importing Structs](#importing-structs)
+      * [Struct Literals](#struct-literals)
+      * [Struct Member Access](#struct-member-access)
+      * [Importing Structs](#importing-structs)
 * [Namespaces](#namespaces)
 * [Scope](#scope)
 * [Optional Parameters &amp; Type Constraints](#optional-parameters--type-constraints)
@@ -2109,6 +2109,39 @@ workflow myWorkflow {
     }
 }
 ```
+
+#### Struct Literals
+Structs can be created and assigned using the `Struct Literal` notation. Struct literal notation enables the creation of typed struct objects which enforces the typing of all of is
+assigned parameters. It vaguely resembles object literal notation ('{ "foo":"bar" }'), however it attempts to be more declarative to help engine implementations apply the proper
+type conversions to nested structs, as well as remove any ambiguity over what the object being constructed represents.
+
+A `Struct Literal` declaration looks like an engine function call, where instead of a function the name of the struct is used followed by parenthesis. For example:
+
+```wdl
+Person( ... )
+```
+
+Arugments placed within the parethesis are key-value pairs, where the key is the name of one of the struct declarations, and the value is the value to set the argument to.  
+There is no need to wrap a key in quotation marks, instead keys are represented in plain text followed by a colon. The value follows after the colon. Multiple arguments can 
+be separated by a comma ',' and arguments do not need to be specified in a specific order.
+
+Values passed to struct literals can be any previously defined declaration, or they themselves can be a literal notation. 
+
+
+```wdl
+
+#Simple case
+File fastq_1
+File fastq_2
+Sample sample_1 = Sample( type: "Blood", sequencing_info: "WGS", fastq: fastq_1 )
+Sample sample_2 = Sample( type: "Liver", sequencing_info: "WES", fastq: fastq_2 )
+Person a = Person(name: "John", age: 30, samples: [sample_1,sample_2])
+
+#You can also use Struct literals wihtin another Struct literal
+Person c = Person( name: "Bob", age: 45, samples: [ Sample( type: "Oral", sequencing_info: "WES", fastq: fastq_3 )] )
+
+```
+
 
 ### Struct Member Access
 
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -328,7 +328,7 @@ struct BamAndIndex {
     File bam
     File bam_index
 }
-BamAndIndex b_and_i = BamAndIndex(bam="NA12878.bam",bam_index="NA12878.bam.bai")
+BamAndIndex b_and_i = BamAndIndex { bam: "NA12878.bam", bam_index: "NA12878.bam.bai" }
 ```
 
 Some examples of types:
@@ -2084,7 +2084,7 @@ task myTask {
 }
 
 workflow myWorkflow {
-    Person harry = Person(name="Harry", age=11)
+    Person harry = Person { name: "Harry", age: 11 } 
     call myTask {
         input:
             a = harry
@@ -2097,18 +2097,14 @@ Structs can be created and assigned using the `Struct Literal` notation. Struct 
 assigned parameters. Struct literal notation attempts to be more declarative to help engine implementations apply the proper
 type conversions to nested structs, as well as remove any ambiguity over what the object being constructed represents.
 
-A `Struct Literal` declaration looks like an engine function call, where instead of a function the name of the struct is used followed by parenthesis. For example:
+A `Struct Literal` declaration is similar to a map literal except it is preceded by a reference to the struct name. For example, the basic syntax for the `Person` struct would look like the following:
 
 ```wdl
-
-
-
-
-Person( ... )
+Person {...}
 ```
 
-Arguments placed within the parenthesis are key-value pairs, where the key is the name of one of the struct declarations, and the value is the value to set the argument to.  
-There is no need to wrap a key in quotation marks, instead keys are represented in plain text followed by an equals sign `=`. The value follows after the `=`. Multiple arguments can 
+Arguments placed within the brackets are key-value pairs, where the key is the name of one of the struct declarations, and the value is the value to set the argument to.  
+There is no need to wrap a key in quotation marks, instead keys are represented in plain text followed by a colon `:`. The value follows after the `:`. Multiple arguments can 
 be separated by a comma ',' and arguments do not need to be specified in a specific order.
 
 Values passed to struct literals can be any previously defined declaration, or they themselves can be a literal notation.
@@ -2123,17 +2119,19 @@ input {
     File fastq_2
 }
 
-Sample sample_1 = Sample( type="Blood", sequencing_info="WGS", fastq=fastq_1 )
-Sample sample_2 = Sample( type="Liver", sequencing_info="WES", fastq=fastq_2 )
-Person person_1 = Person( name="John", age=30, samples=[sample_1,sample_2] )
+Sample sample_1 = Sample { type: "Blood", sequencing_info: "WGS", fastq: fastq_1 }
+
+Sample sample_2 = Sample { type: "Liver", sequencing_info: "WES", fastq: fastq_2 }
+Person person_1 = Person { name: "John", age: 30, samples: [sample_1,sample_2] }
 
 #Example representing using different literal notations
-SomeStruct struct_1 = SomeStruct(someDict={"key":"value"},someArray=[1.0,2.3,1.5])
+SomeStruct struct_1 = SomeStruct { someDict: {"key":"value"}, someArray: [1.0,2.3,1.5] }
 
 #You can also use Struct literals within another Struct literal
-Person person_2 = Person( name="Bob", age=45, samples=[Sample( type="Oral", sequencing_info="WES", fastq=fastq_3 )] )
+Person person_2 = Person { name: "Bob", age: 45, samples: [Sample { type:"Oral", sequencing_info: "WES", fastq: fastq_3 }] )
 
 ```
+
 
 
 ### Struct Member Access
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -2118,11 +2118,15 @@ type conversions to nested structs, as well as remove any ambiguity over what th
 A `Struct Literal` declaration looks like an engine function call, where instead of a function the name of the struct is used followed by parenthesis. For example:
 
 ```wdl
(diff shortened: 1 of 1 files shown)


versions/development/SPEC.md
```diff
@@ -2102,14 +2084,58 @@ task myTask {
...
+
+```wdl
+
+
+
+
+Person( ... )
+```
+
+Arugments placed within the parethesis are key-value pairs, where the key is the name of one of the struct declarations, and the value is the value to set the argument to.
```
DavyCats: I found one more :stuck_out_tongue: 
```suggestion
Arguments placed within the parethesis are key-value pairs, where the key is the name of one of the struct declarations, and the value is the value to set the argument to.  
```

illusional:
Can you construct Struct literals in the output? Something like: 

```wdl
task output_struct_collection {
    input {
        String firstName = "Peter"
    }
    output {
        Person harry = Person(name="Harry", age=5+7)
        Person constructed = Person(name=firstName, age=88)
    }
}
```

As far as I can see, the documentation doesn't address this.

patmagee:
@illusional I am actually not sure if that is supported at the moment by any of the implementations. To me it would be good eventually to get to the point where constructing struct literals in the output is supported, however it will require a bit of work from the workflow engine. I would be interested to know from others if they this for now this should be allowed or labelled as future support
@cjllanwarne @mlin @geoffjentry

geoffjentry:
I agree that I don't think it's currently possible. I think I think it'd be a good idea in the future.

patmagee:
@geoffjentry as it is not explicitly stated in the spec whether we allow this or not, do you think  it would be safe to assume that it is not allowed, or do you think we need to expressly forbid it for the time?

geoffjentry:
I'd double check with @cjllanwarne and @mlin  that I haven't misspoken but if indeed it's not supported we should explicitly disallow it

mlin:
I don't quite see the argument to forbid using the new struct literal syntax in output sections. Cromwell doesn't support it there, but nor does it implement the new literal syntax anywhere, right?

illusional:
I think initialisation in the output would be a useful feature of the language, and not just because I'm [biased](#289 ), is there a way it can be supported?

geoffjentry:
@illusional anything can be passed via vote, but to be folded in requires implementation. I can't speak for @mlin but from the perspective of Cromwell it's more likely to see implementation of things which require less work (disclaimer: I don't actually know how hard it'd be to implement ... @cjllanwarne ?)

DavyCats:
If I'm not mistaken, currently cromwell does allow for the initialization of structs in the outputs using the (in develop removed) `object` notations.
eg.:
```wdl
output {
    IndexedBamFile outputBam = object {
        file: bamFile,
        index: bamIndex
    }
}
```

mlin:
With the new syntax we may want some new constraints on name collisions between struct types and standard library function names (#307), eg my empty `struct stdout {}` appears to have the literal `stdout()`. There may be a concern about introducing new standard library functions that break existing WDLs with colliding struct names.

mlin:
Could someone familiar with Hermes comment on whether the above & related cases complicate language parsing? @cjllanwarne @geoffjentry

cjllanwarne:
* On struct literals v. library functions, I *think* we'll be ok so long as standard library calls look like `x(y, z)` and structs look like `x(a=y, b=z)`. But for user sanity a blanket "don't use standard library names as struct names (or task names, workflow names, value names, ...)" might be appropriate?
  * To be absolutely sure, we could try to check this works in Hermes *before* making a final choice - assuming we're fairly sure the vote will pass, it doesn't seem too additionally painful to try it out before - rather than after - voting?

* On allowing struct literals in outputs: I don't see any reason why we wouldn't allow this. To me a struct literal is just an special case of an expression which evaluates to a struct - and therefore valid anywhere that an expression is valid - runtime attributes, outputs, inputs, value declarations, even String placeholders like `String s = "~{Person(name="Harry", age=17).name}"` if you *really* wanted to .... In fact I think it would be harder and require more special logic for us (Cromwell at least) to *disallow* this.

cjllanwarne:
I had a go at implementing this [in the grammar](https://github.com/openwdl/wdl/pull/308) - I had a few thoughts:

* Mainly: it was tricky to distinguish function calls and struct literals in the grammar. You'll notice that I had to introduce `identifier_being_assigned` token to allow the grammar to distinguish the two parameter list types. This is partially a hermes thing - and partially a "I could probably do better if I looked at it harder"... but to me this is raising a small warning flag that if it's tricky for the grammer, maybe it'll also be tricky for users to distinguish function calls from structs. These suggestions would hopefully make it easier for both:
  * Maybe we could add a keyword to make it obvious that this is a struct being created, eg `Person p = new Person(...)`?
  * Maybe we could make the syntax a little more distinct, eg `Person p = Person((...))`
  * Maybe we don't care
* I don't think we want to allow structs and function calls to share the same names. In fact I'd be in favor of making function names their own token type in the grammar so that we could validate at parsing time whether people are using the right option out of constructor syntax vs function call syntax. But to disclose my own bias: this would simplify the parsing logic too!
* For the sake of flexibility - and if we can distinguish function calls from struct literals in another way, what do people think about (a) allowing function calls to name parameters too, and/or (b) allowing struct literals to *not* name parameters if they're simple?

versions/development/SPEC.md
```diff
@@ -2102,14 +2084,58 @@ task myTask {
...
+assigned parameters. Struct literal notation attempts to be more declarative to help engine implementations apply the proper
+type conversions to nested structs, as well as remove any ambiguity over what the object being constructed represents.
+
+A `Struct Literal` declaration looks like an engine function call, where instead of a function the name of the struct is used followed by parenthesis. For example:
+
+```wdl
+
+
+
+
```
cjllanwarne: That's a lot of empty lines...?

patmagee:
@cjllanwarne thanks for taking a stab at this. In regards to the struct literals in the output section, it appears that the general consensus is that this is already supported and there should be no strict reason to disallow it. If this is the case, I will go ahead and add an example to demonstrate this.

As for the naming collisions, this definitely will represent a problem. Engine functions should be considered key words in my opinion. This has the added side effect of course of making the addition of an engine function a breaking change that must wait for a major version release. But I think this is okay and would be accepted by the community. Therefore there should be no collisions within a `WDL` for a specific version. But if the user wanted to update to the latest `WDL` syntax, it is possible they will have to alias some of their structs.

I am also now concerned about complicating the `struct` literal with an expression from what you described. I am not sure what the best way forward? If anyone is a `grammar guru` and has a clean suggestion that would be appreciated. Otherwise, we might be stuck with updating the syntax of the literal annotation.

A few suggestions:

1. `new Person(name = "john", age = 23)` - This is a bit verbose and is not consistent with any other literal implementation
2. `Person< name = "john", age = 23 >` - Might as well consider this? I dont really like it necessarily... but others might.
3. `Person { name = "john", age = 23 }` - go back to the old object literal syntax... maybe this was not doing anything wrong except having the `object` name.
4. `Person::(name = "john", age = 23)` - I actually kind of like this, as to me It indicates I am calling a function belonging to the `Person` struct.

patmagee:
After looking at the example, I think I personally lean towards using `Person::(name = "john", age=23)` syntax to completely avoid any collisions. Additionally, to me this notation marks this explicitly as a `struct` literal and prevents any confusion with other syntax  (ie other literals, engine functions, call blocks etc)

mlin:
One cheesy trick is to formally constrain struct type names to start with an uppercase letter, while function names start with a lowercase letter. :cheese: :cheese: :cheese:

I also quite like Pat's #3 keeping the existing `object { [k: v ...] }` syntax just replacing the word `object` with the struct type name...

patmagee:
@mlin the one concern I had about repurpossing the object literal syntax was that it might get confused with other types of blocks, either by users or at a grammar level.  But if you do not see any issues arising from that, it might be our easiest way forward

cjllanwarne:
FWIW from a grammar point of view, if we had a defined set of `function_name` keywords then the ambiguity *in the grammer* will hopefully disappear - because we could key off of the differing token types:
```
      (*:left) $e = :function_name :lparen $fcall_param_list :rparen -> FunctionCall(name=$0, params=$2)
      (*:left) $e = :identifier <=> :lparen $struct_param_list :rparen -> StructLiteral(name=$0, params=$2)
```

cjllanwarne:
FWIW2: I also kind of like "Idea 3: repurpose the old `object` syntax but with struct names" idea

patmagee:
@cjllanwarne @mlin the benefit to option 3 imo, is that it would help for users to migrate away from objects, since it would be quite easier. 

@geoffjentry  @illusional @DavyCats any specific thoughts on this before I try implementing something

geoffjentry:
I prefer Option #3

DavyCats:
Options 1, 3 and 4, as well as @mlin's suggestion of making capitalization mandatory for structs and prohibited for functions (for some reason I thought this was already the case), all seem fine to me. Option 2 just looks kinda weird to me and might give grammer ambiquity with `Person` and `name` both being valid variable names and `<` being an operator.

mlin:
The capitalization approach to disambiguation tends to invite :man_facepalming: :woman_facepalming: reactions but does have an illustrious pedigree in ML & Scala! :)

pshapiro4broad:
> The capitalization approach to disambiguation tends to invite 🤦‍♂️ 🤦‍♀️ reactions but does have an illustrious pedigree in ML & Scala! :)

Unless I'm missing something, the use of case to disambiguate functions vs classes in Scala (and Java, and C++, etc) is entirely conventional and not part of the language. Using a different syntax seems like a better solution to me.

FWIW I like #3 from @patmagee 's list above.

orodeh:
I am fine with option #3. I am also in favor of requiring structs to start with an uppercase letter, and functions to start with a lowercase letter. It would make code easier to read, and perhaps, make the grammar easier to write.

mlin:
I know it's hard to believe but...

> [In many places, OCaml makes a distinction between capitalized identifiers and identifiers that begin with a lowercase letter.](https://caml.inria.fr/pub/docs/manual-ocaml/lex.html)

In Scala there is this [less-pervasive case](https://stackoverflow.com/questions/4479474/scala-pattern-matching-with-lowercase-variable-name) of free variables vs scoped identifiers in pattern matching.

:cheese: :cheese: :cheese: 

Hopefully it's clear I'm not without reservations about this solution, just noting precedent!

geoffjentry:
I would prefer to avoid too many syntactic cleanliness type of rules, e.g. enforced capitalizations, etc. As a programmer I'm fine with these things, but many users aren't really programmers and could just get frustrated by seemingly arbitrary rules

orodeh:
In my opinion, requiring structs to start with a capital letter would make WDL code a lot easier to read. For example, under this rule, the variants below, are all struct instances: 
```
Person{ name=x, age=y}
Person{ x, y }
Person(x, y)
```

This syntax (any of the three) is very similar to a standard library call (`Person(x, y)`). The only sure way to distinguish them, is to know all the stdlib functions by heart. As we grow the library, this will become harder.

geoffjentry:
@orodeh I'm all for promoting stuff like that as best practices and would also be for official linters (if we had such a thing) calling stuff like that out. I'm just not particularly stoked about requiring it, but that's really just my own $0.02, YMMV

orodeh:
The reason I think this discussion is important, as it is related to a proposal I have for [extending the standard library](https://github.com/openwdl/wdl/issues/312). If we have many more library functions, it will become hard to distinguish between `struct` construction and a function call. Therefore, In my opinion, there are essentially two practical ways to disambiguate: 

(A) Using the `new` keyword for struct construction 
(B) Requiring upper case for struct definitions, and lowercase for functions.

Version A:
```
new person{ name = "Gordon", age = 41 }
```
It is legal to have a function named `person`.

Version B:
```
Person{ name = "Gordon", age = 41 }
```
It is illegal to have a function named `Person`.

patmagee:
@orodeh @geoffjentry @cjllanwarne @mlin  I am fan of Convention >> Forcing users to go down a specific route. For that reason I am not a huge fan of forced capitalization, but rather linting tools that let users know of `struct` conventions.

Additionally, the `new` keyword would make sense, but If we went that route, I would suggest we normalize literals all across the board with the `new` keyword. I am not a huge fan of having edge case behaviour to satisfy a single requirement. Since there are alot of WDL's which use the literal notation already, I think the best approach is `Person { name = "Bob", age = 23 }` and the `P` is not REQURIED to be a capital.

cjllanwarne:
Thanks @patmagee - since nobody has spoken out against that version of the syntax it seems like a good candidate for my to cycle back round to - I'll try to prototype that in hermes for us within the next few days

patmagee:
Great thanks @cjllanwarne!

vdauwera:
Great discussion, folks. From the perspective of community support wrt capitalization, I’m very much for promoting conventions and against enforcing them with requirements. Case is a tricky thing and can easily cause unwarranted difficulty, especially for people from cultures whose language does not include a concept of case.

rhpvorderman:
I like the overall idea, and I do not have an opinion on syntactic details. It looks quite reasonable and understandable to me. So :+1: and lots of thanks to @patmagee and everyone involved in this discussion.

patmagee:
It looks like general discussion has died down for this PR in its current iteration. I am going to give it 1 more day for any last minute opinions, and then put it to a vote

DavyCats:
:+1:

orodeh:
👍

geoffjentry:
👍

illusional:
👍

geoffjentry:
:+1:

patmagee:
Voting passes 4-0

mlin:
The new style struct literals are [implemented](https://github.com/chanzuckerberg/miniwdl/pull/453) in [miniwdl v0.9.0](https://github.com/chanzuckerberg/miniwdl/releases/tag/v0.9.0) for `version development` documents. The struct type name in the literal definitely helps with typechecking.

I think we can merge this if @patmagee would like to rebase on main and unless Cromwell team has any late-breaking concerns (cc @cjllanwarne)

jdidion (changes requested):

jdidion:
This is implemented in wdlTools

jdidion:

PR #300, opened 2019-04-01
Grammar bugfix: allow empty { } for calls in 1.0 and up

aednichols:
Based on the [grammar in the 1.0 spec](https://github.com/openwdl/wdl/blob/master/versions/1.0/SPEC.md#call-statement), a call block like
```
call my_task {}
```
should be allowed based on the line
```
$call_body = '{' $ws* $inputs? $ws* '}'
```
On that basis I am submitting this grammar change as a bugfix rather than a spec change.

I am aware that [the spec for version `development`](https://github.com/openwdl/wdl/blob/master/versions/development/SPEC.md#call-statement) does not have an equivalent grammar example. However, nothing there seems to _disallow_ this implementation so I am submitting this change for `development` as well.


cjllanwarne:
If the development spec is now ambiguous, I would ask to include a simple "btw, you can also do X" to remove that ambiguity.

Otherwise, this looks great to me.

aednichols:
It seems like the development spec change should be in a separate PR because it requires voting, yes?

aednichols:
Created https://github.com/openwdl/wdl/pull/302

geoffjentry:
Merging as this isn't a spec change

Issue #298, opened 2019-03-27
Override runtime parameters at runtime


illusional:
Is there a way to override runtime parameters at runtime, except for explicitly porting them from the inputs?

eg:
```json
{ "myWorkflow.myTask.runtime.disks": "100 GB" }
```

Or if I did explicitly port them, does (/ should) a value of None in the runtime be override by a backend's default.

geoffjentry:
@illusional One of the many, many, many issues I have with `runtime` is that there's no way to officially do this. It's possible to wire in an input there as any valid WDL expression is viable there, but that requires the WDL to be written that way. You can't do it the way you're suggesting.

One of the other many, many, many issues I have with this is that not everyone sees the current situation as bad :)

I'd personally wholeheartedly support suggested changes to the spec which start to overhaul `runtime`

illusional:
Can I suggest `Specifying / overriding runtime attributes from JSON` as a feature to be added to the openwdl spec. Something similar to [Specifying Workflow Inputs in JSON](https://github.com/openwdl/wdl/blob/master/versions/development/SPEC.md#specifying-workflow-inputs-in-json), and would follow the same resolution order as #141 proposes (backed up by https://github.com/broadinstitute/cromwell/issues/2446). 

Maybe something like:
___

## Specifying / overriding runtime attributes from JSON

Workflow runtime attributes may additionally be specified as key/value pairs within the JSON input. The mapping from JSON or YAML values to WDL values is codified codified similar to the [serialization of task inputs section](#).

In JSON, the user may be able to specify task-specific runtime attributes similar to the following:

{
  "wf.t1.runtime.memory": "16 GB",
  "wf.t2.runtime.cpus": 2,
  "wf.t2.runtime.disks": "100 GB",
  "wf.t3.runtime.arbitrary_key": ["arbitrary value"]
}

As the runtime section consists of key/value pairs, it's up to user to ensure they provide the correct coercible type for the backend they are targeting. See the section on [Type Coercion](#) for more details.

### Resolution Order

Similar to how inputs are resolved, these runtime parameters are resolved in the following order:

1. Inputs provided by the inputs JSON
2. The value provided inline on the task, whether that's an expression or literal value.

rhpvorderman:
> One of the other many, many, many issues I have with this is that not everyone sees the current situation as bad :)

@geoffjentry  I think the current situation with runtime makes it a liability in a cluster environment.
@illusional great addition I would support this if it were a pull request to the spec.

[...]

Issue #262, opened 2018-11-02
Allow passthrough of inputs from the `inputs.json` to a subworkflow


TMiguelT:
Currently, all inputs to a subworkflow and any tasks inside it must be specified in the workflow that calls it. However, this results in a lot of duplication of input metadata. Each input must be listed in:
* The subworkflow task
* The subworkflow 
* The parent workflow
* The `input.json` file

A much cleaner syntax would allow the `inputs.json` to directly specify inputs into the subworkflow and its tasks. Ideally, this WDL should run: https://github.com/TMiguelT/WdlSubworkflowPassthrough

patmagee:
There is nothing prohibiting this in the specification as far as I can tell. I would think this is an implementation detail.  @cjllanwarne @geoffjentry what do you think? If this is more an implementors issue then we can go ahead and close this

illusional:
Hey @geoffjentry / @cjllanwarne, just wondering if Cromwell currently does this. I can't get it to work, and the spec definitely implies that you can reference a workflow's task's input ([development/SPEC.md#specifying-workflow-inputs-in-json](https://github.com/openwdl/wdl/blob/master/versions/development/SPEC.md#specifying-workflow-inputs-in-json)).

I am relying on it for Runtime attribute overrides (#301). If this is matter of the spec SHOULD weigh in, I'm happy to put a PR together. It's the only way you could have a 'runtime.json' specifying all the runtimes attributes required for a platform.

Some additional thoughts anyway, you should specify the call alias if one exists (so you could distinguish different calls of the same tool), and then it falls back to the task / subworkflow name.

eg: 
- `myworkflow.task1name.inputOfTask1Name`
- `myworkflow.mysubworkflowalias.taskname.inputName`

geoffjentry:
@illusional to be clear, are you putting the task name or the call name (assuming they're not the same). That should work.

illusional:
I put the call name, eg: call B.Bwamem as TheCallAliasForBwaMem

Then I’d put:
`myworkflow.TheCallAliasForBwaMem.myinput’ which would give me an error.

geoffjentry:
@illusional eagle eye @cjllanwarne pointed out to me that you're using subworkflows and that's not supported. I don't see the spec being opinionated in one direction or the other, and that should be clarified. 

This is related to #217 

IMO this **should** be supported

illusional:
Was going back through the WDL issue / PR history and found #192 as well.

[...]

Issue #217, opened 2018-06-01
Mapping inputs for sub-workflow calls


wdesouza:
Is it possible to map inputs for calls inside a sub-workflow? Example:

subworkflow.wdl:

```wdl
task A {
  String message

  command {
    echo ${message}
  }

  output {
    File messageFile = stdout()
  }
}

workflow B {
  call A

  output {
    File outputFile = A.messageFile
  }
}
```

```wdl
import "subworkflow.wdl" as sub

workflow C {
  call sub.B {
    input:
      A.message = "Hello, World!"
  }

  output {
    Array[String] message = read_lines(B.outputFile)
  }
}
```

WOMtool validation raises error:

```
ERROR: Unexpected symbol (line 6, col 8) when parsing 'mapping'.

Expected equal, got ..

      A.message = "Hello, World!"
       ^

$mapping = :identifier :equal $e -> IOMapping( key=$0, value=$2 )
```

In real world scenario, I am trying to iterate over samples calling [GATK Best Practices for Data Pre-Processing](https://github.com/gatk-workflows/gatk4-data-processing/blob/master/processing-for-variant-discovery-gatk4.wdl). I would like to map values to call inputs of this workflow. For example, set genome index and dictionary files created by previous calls instead of setting manually in inputs JSON file.

patmagee:
@Welliton309 Try explicitly naming the inputs in the subworkflow which you want to call, instead of only naming them in the task. I do not think you can use a `.` in a input notation according to the existing grammar. 

so change you workflow `B` to:
```wdl
workflow B {
  String message
  call A {
    input: 
      message = message
  }

  output {
    File outputFile = A.messageFile
  }
}
```

FYI there is an open discussion about how better to do imports here: https://github.com/openwdl/wdl/issues/214

Welliton309:
Thank you @patmagee for the recommendation. I think your solution is better than I proposed. However, the problem is how to reuse third-party workflows (as I mentioned) without changing them.

For example, it is not possible to change [this parameter](https://github.com/gatk-workflows/gatk4-data-processing/blob/master/processing-for-variant-discovery-gatk4.hg38.wgs.inputs.json#L12) in a workflow that calls this sub-workflow.

patmagee:
ah yes, @cjllanwarne wdyt?

geoffjentry:
As more of a meta-comment, I agree with @Welliton309 that keeping in mind reuse & composability is key, regardless of the actual implementation

[...]

Issue #141, opened 2017-10-11
Workflow inputs don't override WDL defaults


katevoss:
Truncated from @cjllanwarne 

> I believe the order of precedence should be:
> - Inputs provided by the inputs JSON
> - Inputs specified explicitly by call (i.e. in the workflow doing `call foo {input: inputInt = 6}`)
> - The default value in the task
> 
> Right now the  order appears to be:
> - Inputs specified explicitly by call (i.e. in the workflow doing `call foo {input: inputInt = 6}`)
> - The default value in the task
> - Inputs provided by the inputs JSON

Full context: see the [conversation from the Cromwell repo](https://github.com/broadinstitute/cromwell/issues/1861)

PR #301, opened 2019-04-02
Add Runtime Attribute override to Development Spec

illusional:
I raised an issue (#298) around overriding runtime attributes by JSON input, I've taken the [comment](https://github.com/openwdl/wdl/issues/298#issuecomment-477869635) I left there and turned it into this PR.

I don't believe the JSON runtime override should be able to support expressions, like if someone specifically put in:

```json
{  "wf.my_task.runtime.cpus": "${size(myFileInput) * 4}"  }
```

Ie: I don't think it should resolve myFileInput then perform the computation. I don't know if I should explicitly state this though.


patmagee:
@illusional this looks great. I would err on the side of caution and explicitly ban expressions in the supplied `JSON`.  This is the type of thing that if not banned people will invariably try to do

cjllanwarne:
My only complaint here is the mixing together of runtime and inputs into a single `json` file. Ideally (in my opinion), the runtime would be overridable via a separate `runtime.json` to:

* Avoid name collisions and confusion between inputs and runtime attributes
* Preserve `inputs.json` validation
* Allow separate reuse of inputs and runtime overrides without having to cherry-pick out the values you want to keep.

Today we can validate an `inputs.json` by comparing it to the task. If we say "any unrecognized inputs are runtime overrides" then we lose the ability to detect typos early on.

Another advantage of separate files is easier re-use. With separate files I could take the same `inputs.json` file and swap out the `runtime.json` to run against a different cloud. Or, I could keep the same `runtime.json` and swap out the `inputs.json` to run a different input set on the same cloud. If the files are mixed up together I have to do some tedious cherry-picking of values I want to keep each time.

illusional:
Thanks @patmagee! I've clarified that expressions should not be allowed:

> Expressions are disallowed within the runtime.json, hence no input will be coerced to an expression.

I agree @cjllanwarne that best practices should be to put these in a separate `runtime.json` inputs file, I've more explicitly mentioned this now in my latest commits. Also, my proposal was to add an additional `runtime` tag to transparent that the user is overriding runtime attributes. eg: key: `wf.my_task.runtime.runtime_attribute`.

geoffjentry:
My $0.02 .....

- My initial reaction (from the previous issue) to embedding these in the inputs json was dislike but the more I thought about I moved all the way around to preferring it. There are two reasons for this - one being that my experience has been that many members of the target audience loathe dealing with the extra files, and the other being that Im now starting to picture `runtime` values as just being special inputs (in fact, many Broad based WDLs just use actual inputs to fill in the runtime values)
- Likewise my initial reaction to @patmagee 's suggestion was "Right on!" but the more I think about it I wouldn't mind allowing expression strings in there. TBH I'd be surprised if that wasn't a pretty immediate user request
- Seeing your edit to allow **both** a `runtime.json` and directly embedding in the inputs json is good in the "why not both" manner. I do think that if we're going to go that route that it'll need to be fleshed out a bit more in the current PR, e.g. it's not mentioned in the resolution order. It might also be good to perhaps loop back to [inputs json definition](https://github.com/openwdl/wdl/blob/develop/SPEC.md#specifying-workflow-inputs-in-json) and create a section about companion files in general

illusional:
@geoffjentry @cjllanwarne Do you think there should be a specific type of input called runtime, something like `-r runtime.json` if passing through Cromwell? My intention was just to tell a user that best practice is to separate your inputs into a `inputs.json` and a `runtime.json`, and then pass multiple of those to Cromwell (eg: through `-i`). 

I'm happy to unwind disallowing expressions, my thought was if you can determine your runtime attribute based on an expression, you may as well bake it into the wdl. But thinking about it you may want a specific disk format for GCP, or something else for a different backend.

geoffjentry:
@illusional Ah I see what you mean. I don't have a strong opinion. 

re expressions I also odn't feel super strongly about this either. Happy to go w/ the flow of public opinion

cjllanwarne:
#### Inputs names
I didn't notice that `.runtime.` in the name string before - as long as we disallow `runtime` as a task or value name that should be fine (and I think that's pretty reasonable!).

#### Add `runtime.json` vs allow multiple `input.json`s

I have to say that I do like the idea of considering these to just be multiple input files which can be combined together or left separate per user preference. Cromwell already allows this so (selfishly) it would be nice to just make that formal in the spec.

Multiple compose-able input files would allow you to, for example, break down your submission into:
* Basic inputs (which rarely change between runs. A reference genome for example)
* Run specific inputs (the ones that you're tweaking this time)
* Runtime inputs
* Other as-yet-unforeseen inputs?

Or, if you prefer, just bundle them all together as a single package (which again, someone could surgically override later if they chose to).

The more I think about it, the more that seems to make the most sense to me!

#### Expressions in input files

I'd prefer to be conservative with allowing expressions-in-inputs for now. I'm 50:50 on the merits so my only strong arguments for doing it separately would be (a) seeing whether it's really needed, (b) a background/general preference for iterative changes vs "change everything at once" PRs

geoffjentry:
The more I think about it the more I like just saying that best practice is to provide it as a separate input file. This reinforces that view point of runtime attrs being special inputs instead of a pure first class thing. It also means not having to define a new file type.

However, that **does** imply codifying that there can be multiple input jsons and specifying things like resolution rules.

illusional:
> as long as we disallow runtime as a task or value name that should be fine
@cjllanwarne That's a great point, is there an easy way I can modify my PR to prohibit this, like potentially modifying the validator regex to not accept "runtime"?

Your multiple compose-able inputs is exactly what we're planning do when we run ours. We can specify essentially static files for specific environments (reference genome etc), one for our input files and then (if this proposal passes) one for our runtime for each environment.

@geoffjentry it seems that the spec doesn't do a great job of specifying the order of resolution rules (except maybe [#variable resolution](https://github.com/openwdl/wdl/blob/master/versions/development/SPEC.md#variable-resolution), hence why I added a section to mine (though I think I could clarify the wording). There is an order that has been proposed, specifically #141 (backed up by [broadinstitute/cromwell#2446](https://github.com/broadinstitute/cromwell/issues/2446)) but not implemented yet. I'm also happy to spin up another PR to address this if the community wants that.

versions/development/SPEC.md
```diff
@@ -2646,6 +2648,30 @@ In JSON, the inputs to the workflow in the previous section might be:
 
 It's important to note that the type in JSON must be coercible to the WDL type.  For example `wf.int_val` expects an integer, but if we specified it in JSON as `"wf.int_val": "three"`, this coercion from string to integer is not valid and would result in a coercion error.  See the section on [Type Coercion](#type-coercion) for more details.
 
+## Specifying / Overriding Runtime Attributes in JSON
+
+Workflow runtime attributes may additionally be specified as key/value pairs within a JSON input file. Ideally this should be a separate `runtime.json` file to   The mapping from JSON or YAML values to WDL values is codified similar to the serialization of task inputs section, however with an additional runtime tag to avoid name collisions and allow for unspecified attributes to be set by the runtime input file. Runtime attributes do not need to be specified in the task defintion to be overidden or set by the JSON.
```
pshapiro4broad: > Ideally this should be a separate `runtime.json` file to   The mapping from JSON

typo?

illusional: Nice catch, fixed this to read.

> It's recommended that this should be a separate inputs file called `runtime.json`. The mapping from JSON ....

versions/development/SPEC.md
```diff
@@ -2646,6 +2648,30 @@ In JSON, the inputs to the workflow in the previous section might be:
 
 It's important to note that the type in JSON must be coercible to the WDL type.  For example `wf.int_val` expects an integer, but if we specified it in JSON as `"wf.int_val": "three"`, this coercion from string to integer is not valid and would result in a coercion error.  See the section on [Type Coercion](#type-coercion) for more details.
 
+## Specifying / Overriding Runtime Attributes in JSON
+
+Workflow runtime attributes may additionally be specified as key/value pairs within a JSON input file. It's recommended that this should be a separate inputs file called `runtime.json`. The mapping from JSON or YAML values to WDL values is codified similar to the serialization of task inputs section, however with an additional runtime tag to avoid name collisions and allow for unspecified attributes to be set by the runtime input file. Runtime attributes do not need to be specified in the task defintion to be overidden or set by the JSON.
```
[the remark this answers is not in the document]
geoffjentry: I was getting ready to throw down a giant "you're wrong", but you're correct that the [spec](https://github.com/openwdl/wdl/blob/master/versions/development/SPEC.md#specifying-workflow-inputs-in-json) does not **literally** say this is a different file. However it's heavily implied and as one of the people who are responsible for all the less than correct things in the spec I can assure you that's what we meant.

I like how here it's a suggestion, we don't **need** to make a suggestion but IMO it's not a bad idea to encourage conformist behavior

cjllanwarne:
I'd like to throw in one more "it would be *really* nice to tidy this up at the same time" -

I believe (but it may take some hunting down) that there's a line somewhere that says "you can't provide inputs directly to subworkflows or their tasks". Since we're saying that `runtime` are really just inputs, and since I would hope that runtime could be provided all the way down the call stack, it might be nice to also remove that line (and/or explicitly state somewhere that we can provide any inputs we choose as part of the input set)

dheiman:
I haven't seen this mentioned yet, apologies if it has, but if runtime inputs are to become settable via json, they **should not** be used for call caching. This would eliminate so many headaches where memory is mapped to a regular input and needs to be tweaked for unforeseen reasons

geoffjentry:
@dheiman Just keep in mind that call caching isn't a WDL thing, it's a potential feature of implementations

dheiman:
@geoffjentry fair enough, but isn't an implementation required before the PR can be merged? I'd be happy to move the comment over to the Cromwell PR/issue if this passes.

geoffjentry:
@dheiman Yeah, that's exactly what I was getting at. Also keep in mind that voting **here** can't reasonably be done with a disclaimer of "as long as my preferred implementation does X" since the `Waiting for implementation` phase happens post-voting.

versions/development/SPEC.md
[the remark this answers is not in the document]
TMiguelT: > however with an additional runtime tag
I don't think this is precise enough. I would say "nested inside the `task.runtime` object".

I agree with @pshapiro4broad that the spec shouldn't mention input files, since that's implementation specific. If we want to acknowledge input files in the spec it should be a separate PR. However, this PR might be able to say something like "we recommend that the executor supports providing runtime attributes separately from the workflow inputs, for instance as a separate input JSON file"

versions/development/SPEC.md
```diff
@@ -2646,6 +2648,30 @@ In JSON, the inputs to the workflow in the previous section might be:
...
+```json
+{
+  "wf.t1.runtime.memory": "16 GB",
+  "wf.t2.runtime.cpus": 2,
+  "wf.t2.runtime.disks": "100 GB",
+  "wf.t3.runtime.arbitrary_key": ["arbitrary", "value"]
+}
+```
+
+As the runtime section consists of key/value pairs, it is the user's responsibility to ensure they provide the correct coercible type for the backend they are targeting. Expressions are disallowed within the `runtime.json`, hence no input will be coerced to an expression. See the section on [Type Coercion](#type-coercion) for more details.
```
TMiguelT: It's not that expressions are not *allowed*, it's that they're not evaluated. For instance you can put in `1+1` as a runtime attribute value, it just won't end up as `2` downstream.

illusional: Yeah I see how my initial wording is a little ambiguous, I've modified it (in 26537d5c5058e8278dd157f52759cc5f2c5d135c) to read:

> An input value should not be coerced to an expression, hence values provided within the `runtime.json` should be evaluated.

I think it is clearer to read that what you provide _shouldn't_ be evaluated (like if you provided `"my test ${1+1} expression"`, it would remain), and the language better aligns with the rest of the spec.

versions/development/SPEC.md
```diff
@@ -91,6 +91,8 @@ Table of Contents
   * [Computing Task Inputs](#computing-task-inputs)
   * [Computing Workflow Inputs](#computing-workflow-inputs)
   * [Specifying Workflow Inputs in JSON](#specifying-workflow-inputs-in-json)
+  * [Specifying / Overriding runtime attributes in JSON](#specifying--overriding-runtime-attributes-in-json)
+    * [Resolution Order](#resolution-order)
   * [Optional Inputs](#optional-inputs)
```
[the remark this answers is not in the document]
illusional: Sneaky sneaky change: bc9afaa5526d8627f73794746c68495c71609289

versions/development/SPEC.md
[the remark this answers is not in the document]
illusional: I'm happy to remove my statement in favour of @TMiguelT's:

> We recommend that the executor supports providing runtime attributes separately from the workflow inputs, for instance as a separate `runtime` JSON file.

I'm mixed, I'm very opinionated that you should specify your runtime (such as specific google cloud, or specific SFS) attributes separately to your inputs.  I think it promotes a good practice, and will probably allow people to better share optimised workflows. The more I think about it, the less I believe the spec is the right place to put this, but I'm also not sure if there is another place. 

Our use case is, we want to run the same workflow with varying inputs sizes on different backends. I can create specific optimised runtime jsons for our different targets and give them to other researchers or clinicians who don't need to know about the attributes we decided.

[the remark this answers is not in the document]
TMiguelT: My wording does encourage the implementations to allow this, but it doesn't require it, because as you say it's not exactly something the spec can require

versions/development/SPEC.md
TMiguelT: Yes, I think you should say they shouldn't be evaluated. But right now your wording says "values provided within the runtime.json should be evaluated", which is incorrect.

illusional: ... maybe I shouldn't add attention to detail on my resume anymore (fixed).

patmagee:
It seems as though this PR brings up a bit of a delicate issue that I think we need to resolve before we can consider voting on this.

**Should the Spec define inputs and the specific input files and their *Specific* format, or should this be an implementation detail**

There are actually benefits to both sides of this. being opinionated on the matter leads to greater conformance and reusability of workflows, at the potential cost of strong arming implements. Additionally by defining this right in the specification, open standards such as [WES](https://github.com/ga4gh/workflow-execution-service-schemas) become easier to implement at a language level.

At the same time, requiring inputs in a specific format here (with support for specific files), could be beyond the capabilities of different implementors. Additionally, it might make it more harder for them to support the full feature set of WDL

What are everyone's thoughts?

rhpvorderman:
My thought is that I hate to use JSON for configuration since YAML is so much better for humans to edit. So I think the spec should define the input files, but **not** the format. (Unless it is my preferred format YAML of course :wink: ). But seriously, that bit should be an implementation detail. Who knows what useful formats come along in the future?
You would not want to be stuck on something like XML for configuration because "the spec says so". Any file that can be parsed to some sort of dictionary should be fine.

geoffjentry:
@patmagee My $0.02 is to treat them separately and I don't think a particular order is necssary. We can evaluate this change given the current spec (the one which **does** specify input format) and then if someone wants to remove that portion of the spec (or modify it) afterwards that can be a separate discussion afterwards. If that implies changes to this or anything else in the spec, it's on that proposer to make the appropriate edits.

IOW i think this has been hanging around long enough to judge it on its own merits and let complaints about inputs being codified be handled afterwards.

illusional:
@patmagee It's something I definitely didn't consider when spinning this PR up (hence the repeated use of JSON), but I wholeheartedly agree that the spec should not make an opinion on the format of inputs, eg no use of JSON, YAML, XML or others. Happy for whatever route we go, except that for tracking purposes it would be better to create a separate issues / PR to ensure that the community gets to vote on that.

I do want the spec to be opinionated about best practices for these specific input files, I think it promotes the wider goal of portability. 

---

Side note, these two issues will need to be resolved before this gets implemented, or maybe even voted on. @geoffjentry Is it worth me spinning up a new issue to cover these, or put it in with this (as this feature would require pass through of inputs to subworkflows to work.

- https://github.com/openwdl/wdl/issues/217
- https://github.com/openwdl/wdl/issues/262

illusional:
Not to piggy back on this issue, but adding it as a note here so I don't forget. It would also be great if you could reference a runtime attribute inside the WDL command. For example, many tools require, or allow you to pass a `threads` attribute, where this is intrinsically tied to the runtime.

geoffjentry:
@illusional Sorry, I'm confused - why do the subworkflow PRs need to come first?

illusional:
@geoffjentry Otherwise you couldn't pass runtime attributes to tools under a subworkflow. I guess it's not directly a pre-req, but something I feel should be completed before this feature is useful.

geoffjentry:
Ok. Yeah I was wondering why it was a hard requirement and not a nice to have. Either way, this is your PR so if you want to table it until all of this other stuff is hashed out that's fine. As I said above I don't think these things all need to fit into a particular order, but that's up to you :)

geoffjentry:
Hi everyone. It occurred to me that it is a common pattern in Broad-authored WDLs (and thus the majority of WDLs out in the wild) to already do what @illusional is suggesting and wire their runtime attributes in as workflow inputs.

In effect, the WDL authors have already spoken and this is the model that they would like to see. This PR would clean up that behavior and make it less ad hoc.

The subworkflows issue would be nice to have but as @illusional noted it is not truly a pre-req. The inputs json thing, well, it's already in the spec. We should not be voting against one thing because we dont like another thing already in the spec (and IMO it not only **should** be in the spec, but enhanced).

@illusional - I'd like to move this forward. If you want to hold off let me know and I'll take this over.

illusional:
Thanks @geoffjentry, happy for it to move forward!

geoffjentry:
@illusional I realized it was going to be a pain to make commits to this as it's on your fork. Closing and reopening as #313

TMiguelT:
Could you not have used the "allow edits from maintainers" option?

illusional:
It was off when I had a look, but it's not an option when you create a PR, just tucked away after you've created one. I turned it on now (not that it's any good).

geoffjentry:
I also didn't realize that was a thing :)

PR #306, opened 2019-04-05
Remove + from non-array objects and spec lint fix

ying-w:
Removed some legacy examples from spec where `+` was used to indicate array instead of explicitly declaring as Array (see spec-2 changes)

Partial lint fix for 1.0 Spec, didn't get everything got most of it.

### DIFF
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -1740,6 +1740,8 @@ workflow wf {
 ```
 
 The call inputs block (eg `{ input: x=a, y=b, z=c }`) is optional and specifies how to satisfy a subset of the the task or workflow's input parameters.
+The call inputs block (eg `{ input: x=a, y=b, z=c } `) is optional and specifies how to satisfy a subset of the the task or workflow's input parameters. An empty call inputs block of the form `call no_input_task { }` is valid and has the same meaning as `call no_input_task`.
+>>>>>>> f31dc7c2d1297289d44e7f986bced1c83caffc90
 
 Each variable mapping in the call inputs block maps input parameters in the task to expressions from the workflow.  These expressions usually reference outputs of other tasks, but they can be arbitrary expressions.
 
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -23,6 +23,7 @@ Table of Contents
     * [Array Literals](#array-literals)
     * [Map Literals](#map-literals)
     * [Pair Literals](#pair-literals)
+    * [Optional Literals](#optional-literals)
   * [Versioning](#versioning)
   * [Import Statements](#import-statements)
   * [Task Definition](#task-definition)
@@ -303,7 +304,7 @@ workflow wf {
 
 ### Types
 
-In WDL *all* types represent immutable values. 
+In WDL *all* types represent immutable values.
   - Even types like `File` and `Directory` represent logical "snapshots" of the file or directory at the time when the value was created.
   - It's impossible for a task to change an upstream value which has been provided as an input: even if it makes changes to its local copy the original value is unaffected.
 
@@ -352,7 +353,7 @@ For more information on type and how they are used to construct commands and def
 #### Numeric Behavior
 
 `Int` and `Float` are the numeric types.
-`Int` can be used to hold a signed Integer in the range \[-2^63, 2^63). 
+`Int` can be used to hold a signed Integer in the range \[-2^63, 2^63).
 `Float` is a finite 64-bit IEEE-754 floating point number.
 
 #### Custom Types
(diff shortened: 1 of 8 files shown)


mlin:
Good catches but I think we need to make these changes on the [development](https://github.com/openwdl/wdl/tree/master/versions/development) spec version, leave draft-2 and 1.0 as they are and...feel the burn.

geoffjentry:
As @mlin mentioned, we have a policy (for better or for worse) of not retroactively fixing old versions

ying-w:
I've reverted the changes to 1.0 and draft-2 and put in changes to development.

While going through I noticed something inconsistent in `Struct serialization using write_json()`

```
{
  "name", "John"
  "age", 5
  "friends": ["James", "Jim"]
}
```
Should the last line of the block use comma instead of colon

```
  "friends", ["James", "Jim"]
```

versions/development/SPEC.md
```diff
@@ -802,8 +804,8 @@ Some examples of correct import resolution:
 | Root Workflow Location                                | Imported Path                      | Resolved Path                                           |
 |-------------------------------------------------------|------------------------------------|---------------------------------------------------------|
 | file://foo/bar/baz/qux.wdl                            | some/task.wdl                      | file://foo/bar/baz/some/task.wdl                        |
-| http://www.github.com/openwdl/coolwdls/myWorkflow.wdl | subworkflow.wdl                    | http://www.github.com/openwdl/coolwdls/subworkflow.wdl  |
-| http://www.github.com/openwdl/coolwdls/myWorkflow.wdl | /openwdl/otherwdls/subworkflow.wdl | http://www.github.com/openwdl/otherwdls/subworkflow.wdl |
+| `http://www.github.com/openwdl/coolwdls/myWorkflow.wdl` | subworkflow.wdl                    | `http://www.github.com/openwdl/coolwdls/subworkflow.wdl`  |
```
cjllanwarne: I would leave these in, they're probably there to stop github from rendering these made up paths are links.

versions/development/SPEC.md
```diff
@@ -985,16 +992,16 @@ For example, `${true='--enable-foo' false='--disable-foo' allow_foo}` would eval
...
 Both `true` and `false` cases are required. If one case should insert no value then an empty string literal should be used, eg `${true='--enable-foo' false='' allow_foo}`
 
-> 1.  `true` and `false` values MUST be string literals.
-> 2.  `true` and `false` are only allowed if the type is `Boolean`
-> 3.  Both `true` and `false` cases are required.
-> 4.  Consider using the expression `${if allow_foo then "--enable-foo" else "--disable-foo"}` as a more readable alternative which allows full expressions (rather than string literals) for the true and false cases.
+1. `true` and `false` values MUST be string literals.
+1. `true` and `false` are only allowed if the type is `Boolean`
+1. Both `true` and `false` cases are required.
+1. Consider using the expression `${if allow_foo then "--enable-foo" else "--disable-foo"}` as a more readable alternative which allows full expressions (rather than string literals) for the true and false cases.
```
cjllanwarne: When I've tried to do things similar to this in the past, people have pointed out that a list of `1.`s is harder to read if you're using a text editor (vs reading the rendered MD file). Personally I'm not too fussed either way so I normally let them have their way... 😄

versions/development/SPEC.md
```diff
@@ -1088,49 +1095,54 @@ output {
...
+```sh
 dir/
  - a           # a file, 10 MB
  - b -> a      # a softlink to 'a'
 ```
 
 As a WDL directory this would manifest as:
-```
+
+```sh
```
cjllanwarne: This is more `txt` than `sh`

versions/development/SPEC.md
```diff
@@ -1401,16 +1415,16 @@ task bwa_mem_tool {
   input {
     Int threads
     Int min_seed_length
-    Int min_std_max_min
+    Array[Int] min_std_max_min
```
cjllanwarne: I'd make this `Array[Int]+`

versions/development/SPEC.md
```diff
@@ -1401,16 +1415,16 @@ task bwa_mem_tool {
...
     File reference
-    File reads
+    Array[File] reads
```
cjllanwarne: and `Array[File]+ reads`

versions/development/SPEC.md
```diff
@@ -1421,7 +1435,7 @@ task bwa_mem_tool {
 }
 ```
 
-Notable pieces in this example is `${sep=',' min_std_max_min+}` which specifies that min_std_max_min can be one or more integers (the `+` after the variable name indicates that it can be one or more).  If an `Array[Int]` is passed into this parameter, then it's flattened by combining the elements with the separator character (`sep=','`).
+Notable pieces in this example is `${sep=',' min_std_max_min}` which specifies that min_std_max_min can be one or more integers (the `+` after the variable name indicates that it can be one or more).  If an `Array[Int]` is passed into this parameter, then it's flattened by combining the elements with the separator character (`sep=','`).
```
cjllanwarne: This whole paragraph is a bit suspect. I suggest:
> One notable piece in this example is `${sep=',' min_std_max_min}` which specifies that min_std_max_min will be flattened by combining the elements with the separator character (`sep=','`).

cjllanwarne:
@ying-w I would think the example you found should be using colons and commas, representing a JSON input eg:
```wdl
{
  "name": "John",
  "age": 5,
  "friends": ["James", "Jim"]
}
```

versions/development/SPEC.md
ying-w: wouldn't changing them to code blocks stop the rendering?

versions/development/SPEC.md
ying-w: i agree but felt wanted the text after `#` to show up as comment

ying-w:
@cjllanwarne in the context of where that code block is, I felt like it would be weird to say you pass in a struct as json and you get back a json. I believe the code block is trying to explain what the struct is holding, maybe something like

```txt
name -> "John"
age -> 5
friends -> ["James", "Jim"]
```

versions/development/SPEC.md
TMiguelT: I think if you're going to use code tags here to prevent the hyperlinks, you should do it to all entries in this table so it looks consistent. Or use one of the other workarounds: https://gist.github.com/alexpeattie/4729247

versions/development/SPEC.md
```diff
@@ -1401,16 +1413,16 @@ task bwa_mem_tool {
   input {
     Int threads
     Int min_seed_length
-    Int min_std_max_min
+    Array[Int]+ min_std_max_min
```
TMiguelT: How did this ever work? Could you pass multiple values to a scalar `Int` in the past?

ying-w: i'm not sure, might have just been a typo?

patmagee:
@ying-w is this ready to be moved to voting?

ying-w:
@patmagee yup just resolved some conflicts

ying-w:
can someone merge this?

PR #307, opened 2019-04-05
add keyword list

mlin:
Not exciting, but we need an explicit list of keywords that can't be used as identifiers, because they'd complicate language parsing or cause ambiguities/confusion. 

Not included in this list, but debatable: `default sep version`

### DIFF
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -24,6 +24,7 @@ Table of Contents
     * [Map Literals](#map-literals)
     * [Pair Literals](#pair-literals)
     * [Optional Literals](#optional-literals)
+    * [Keywords](#keywords)
   * [Versioning](#versioning)
   * [Import Statements](#import-statements)
   * [Task Definition](#task-definition)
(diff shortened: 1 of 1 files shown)


patmagee:
:+1:

DavyCats:
Look good to me, I'm just wondering whether this the best way to format this list. It seems like it would be easy to overlook an item in the list. Maybe ordering them alphabetically would be better?

pshapiro4broad:
+1

orodeh:
👍

rhpvorderman:
:+1:

illusional:
+1

TMiguelT:
I think this would ideally be mentioned in the `Whitespace, Strings, Identifiers, Constants` section, because this is restricting the set of valid identifiers already listed there.

However, I'm not allowed to request changes, so this is a +1 so it at least gets into the spec.

geoffjentry:
This passes and is `waiting for implementation`

@TMiguelT and @mlin - if one of you would want to make a followup PR representings Michael's comments that'd be great

cjllanwarne:
Unless anyone disagrees, I think this can be merged as implemented - I don't think any existing engines run counter to this by accidentally allowing these values?

PR #309, opened 2019-04-18
fix escape sequence table

DavyCats:
I just noticed that this table wasn't rendering correctly.


geoffjentry:
Oh good catch! Merging this in as it's a markdown fix & not a spec change

Issue #20, opened 2016-01-05
Optional Outputs


tmdefreitas:
Broad's GDAC often deals with sparsely and inconsistently populated data, and many analyses support a range of acceptable inputs. For example, our methylation preprocessor will correlate methylation data to expression data if it is available for a sample set. However, this means that in addition to optional inputs (already supported by WDL), output files are also optional, or at least conditional upon inputs. Currently cromwell raises an error when an expected output is not present, and I don't believe there is syntax to support it in WDL.

It would be nice to be able to specify that an output is optional. One solution could be to mirror to the optional input syntax, and I could specify optional outputs like this:

```
task my_task {
    ... inputs, commands, etc...

    output {
        required_output="file1"
        optional_output?="file2"
    }
}
```

Having this would eliminate the need to write additional logic to create/handle blank files.

scottfrazer:
Seems legit to me

cjllanwarne:
I think you can get *almost* the same thing already using globs - I believe optional_output = glob("file2") will give you either a 0 or 1 element array of files depending on the existence or otherwise of the file.

tmdefreitas:
Thanks for pointing out that workaround, that would at least allow this:

```
output {
  Array[File] optional_output = glob("file2")
}
```

It would still be nice at some point to not worry about 0/1-element arrays, since that syntax doesn't make it clear that the array will only be size 0 or 1.

davidbenjamin:
I'm running into the same issue writing scripts for Mutect 2.

davidbenjamin:
In my case, it's not that a task may or may not produce an output, it's that some tasks may not be run at all depending on a conditional.

vdauwera:
@kshakir Has this been addressed on the Cromwell side? I remember some chatter but I have no details...

kshakir:
Not yet. Here is the cromwell issue: https://github.com/broadinstitute/cromwell/issues/1838

antonkulaga:
I totally support this issue. We need Option-s to work

jtratner:
See https://github.com/openwdl/wdl/pull/310 for PR adding this to spec (and note that cromwell seems to already support this behavior)

hisplan:
Has this been implemented yet?

jtratner:
It works as described in both dxWDL and in Cromwell.

jtratner:
If there’s a conformance test spec I’d be happy to add to it (or could take
advantage of the tests @orodeh wrote for dxWDL)

[...]

PR #310, opened 2019-04-21
Add section on file outputs including optional outputs

jtratner:
Clarify how optional outputs currently work (which I believe is supported by both Cromwell and dxDL right now).  I also noticed there is no definition of how file outputs are supposed to be resolved, so I included a first pass fo that as well.

Here's running example from this doc.
```
›› miniwdl cromwell example.wdl                                                                                      11:39 Apr21
+ mkdir -p /20190421_113957_optional_output
Cromwell input: {}
+ java -DLOG_LEVEL=warn -DLOG_MODE=pretty -jar /var/folders/fm/sjgpzb856kld04jvq82bxrcw0000gq/T/cromwell-36.jar run /example.wdl -o /20190421_113957_optional_output/cromwell/options.json -i /seq_pipeline/20190421_113957_optional_output/inputs.json [2019-04-21 11:40:10,03] [warn] SingleWorkflowRunnerActor: received unexpected message: Done in state RunningSwraData
{
  "outputs": {
    "optional_output.example_optional": null,
    "optional_output.array_optional": [null, "/20190421_113957_optional_output/cromwell/cromwell-executions/optional_output/092d98b8-f118-4376-9dee-ad83000f76bf/call-optional_output/execution/arr2.exists.txt"],
    "optional_output.example_exists": "/20190421_113957_optional_output/cromwell/cromwell-executions/optional_output/092d98b8-f118-4376-9dee-ad83000f76bf/call-optional_output/execution/example_exists.txt"
  },
  "id": "092d98b8-f118-4376-9dee-ad83000f76bf"
}
{
  "outputs": {
    "optional_output.example_optional": null,
    "optional_output.array_optional": [
      null,
      "/20190421_113957_optional_output/cromwell/cromwell-executions/optional_output/092d98b8-f118-4376-9dee-ad83000f76bf/call-optional_output/execution/arr2.exists.txt"
    ],
    "optional_output.example_exists": "/20190421_113957_optional_output/cromwell/cromwell-executions/optional_output/092d98b8-f118-4376-9dee-ad83000f76bf/call-optional_output/execution/example_exists.txt"
  },
  "id": "092d98b8-f118-4376-9dee-ad83000f76bf",
  "dir": "/20190421_113957_optional_output"
}
```

### DIFF
--- a/versions/1.0/SPEC.md
+++ b/versions/1.0/SPEC.md
@@ -1039,8 +1039,7 @@ output {
 
 #### Files and Optional Outputs
 
-File outputs are described as string paths relative to the initial working
-directory.
+File outputs are represented as string paths. Relative paths are interpreted relative to the execution directory; whereas absolute paths are defined in a container-dependent way.
 
 ```
 File somefile = "my/path/to/something.txt"
@@ -1048,7 +1047,7 @@ File somefile = "my/path/to/something.txt"
 
 All file outputs are required to exist, otherwise the task will fail.
 
-However, outputs may be annotated as `Optional` types (including array types),
+However, outputs may be annotated as `Optional` types (including within arrays as `Array[File?]`)
 in which case they will be null if files do not exist.
 
 E.g., in this example task:
@@ -1070,8 +1069,8 @@ task optional_output {
 If run, will generate the following outputs:
 
 * `optional_output.example_exists` will resolve to a File
-* `optional_output.example_optional` will resolve to `null`
-* `optional_output.array_optional` will resolve to `[<File>, null]`
+* `optional_output.example_optional` will resolve to `None`
+* `optional_output.array_optional` will resolve to `[<File>, None]`
 
 #### Globs
 
--- a/versions/1.0/SPEC.md
+++ b/versions/1.0/SPEC.md
@@ -1048,7 +1048,7 @@ File somefile = "my/path/to/something.txt"
 All file outputs are required to exist, otherwise the task will fail.
 
 However, outputs may be annotated as `Optional` types (including within arrays as `Array[File?]`)
-in which case they will be null if files do not exist.
+in which case they will be `None` if files do not exist.
 
 E.g., in this example task:
 
--- a/versions/1.0/SPEC.md
+++ b/versions/1.0/SPEC.md
@@ -41,6 +41,7 @@
       * [Alternative heredoc syntax](#alternative-heredoc-syntax)
       * [Stripping Leading Whitespace](#stripping-leading-whitespace)
     * [Outputs Section](#outputs-section)
+      * [Files and Optional Outputs](#files-and-optional-outputs)
       * [Globs](#globs)
         * [Task portability and non-standard BaSH](#task-portability-and-non-standard-bash)
     * [String Interpolation](#string-interpolation)
@@ -1035,6 +1036,42 @@ output {
   String ab = a + "b"
 }
 ```
+
+#### Files and Optional Outputs
+
+File outputs are described as string paths relative to the initial working
+directory.
+
+```
+File somefile = "my/path/to/something.txt"
+```
+
+All file outputs are required to exist, otherwise the task will fail.
+
+However, outputs may be annotated as `Optional` types (including array types),
+in which case they will be null if files do not exist.
+
+E.g., in this example task:
+
+```
+task optional_output {
+    command <<<
+        touch example_exists.txt
+        touch arr2.exists.txt
+    >>>
+    output {
+        File example_exists = "example_exists.txt"
+        File? example_optional = "example_optional.txt"
+        Array[File?] array_optional = ["arr1.dne.txt", "arr2.exists.txt"]
+    }
+}
+```
+
+If run, will generate the following outputs:
+
+* `optional_output.example_exists` will resolve to a File
+* `optional_output.example_optional` will resolve to `null`
+* `optional_output.array_optional` will resolve to `[<File>, null]`
 
 #### Globs
 


jtratner:
Would close https://github.com/openwdl/wdl/issues/20

jtratner:
cc @orodeh - does this match your expectations on optional outputs for dxWDL as well?

orodeh:
Thanks @jtratner, this looks good. 

I have some reservations about the general ability to treat a `File` type as a `String` in workflow context, but that is a story for another time.

versions/1.0/SPEC.md
```diff
@@ -1036,6 +1037,42 @@ output {
 }
 ```
 
+#### Files and Optional Outputs
+
+File outputs are described as string paths relative to the initial working
```
DavyCats: Depeding on what you mean by `initial working directory`, this might conlict with another part of the spec:
At [File Outputs](https://github.com/openwdl/wdl/blob/master/versions/development/SPEC.md#file-outputs):
```
Individual File outputs can be declared using a String path relative to the execution directory. For example:
```

Here I'm assuming that the `initial working directory` refers to the directory one runs their execution engine from, whereas `execution directory` refers to a dedicated workspace for each task (ie. the cromwell implementation).

I'm wondering how making paths relative to the initial working directory will affect things like globbing and scattering. This might greatly the risk of tasks overwritting each others outputs in the case of scatters and with globbing the globbed files may end up including those from multiple tasks (or even completely unrelated files, if you already had  files in the working directory to begin with).

versions/1.0/SPEC.md
```diff
@@ -1036,6 +1037,42 @@ output {
...
+directory.
+
+```
+File somefile = "my/path/to/something.txt"
+```
+
+All file outputs are required to exist, otherwise the task will fail.
+
+However, outputs may be annotated as `Optional` types (including array types),
+in which case they will be null if files do not exist.
```
DavyCats: Since there is a literal for null nowadays, it maybe good to use that here.
```suggestion
in which case they will be `None` if files do not exist.
```

versions/1.0/SPEC.md
```diff
@@ -1036,6 +1037,42 @@ output {
...
+File outputs are described as string paths relative to the initial working
+directory.
+
+```
+File somefile = "my/path/to/something.txt"
+```
+
+All file outputs are required to exist, otherwise the task will fail.
+
+However, outputs may be annotated as `Optional` types (including array types),
```
DavyCats: Does `including array types` mean `Array[File?]` or `Array[File]?`?

versions/1.0/SPEC.md
```diff
@@ -1036,6 +1037,42 @@ output {
...
+        Array[File?] array_optional = ["arr1.dne.txt", "arr2.exists.txt"]
+    }
+}
+```
+
+If run, will generate the following outputs:
+
+* `optional_output.example_exists` will resolve to a File
+* `optional_output.example_optional` will resolve to `null`
+* `optional_output.array_optional` will resolve to `[<File>, null]`
```
DavyCats: Here `None` might be more appropriate that `null`, as well.

versions/1.0/SPEC.md
jtratner: I meant `Array[File?]`

jtratner:
@DavyCats - updated to use `None` rather than `null` and specified `execution directory` for relative paths rather than `initial working directory`.

Does this look okay to you?

Are there normative WDL test cases that I should add to as well? (e.g., some kind of programmatic version of the example I put up)

geoffjentry:
@jtratner note that this needs to be against the `development` spec, not `1.0`

DavyCats:
Looks good to me. I'm not aware of any kind of testing for the specs.

cjllanwarne:
Two comments. I'd be *very* curious for thoughts on item (1) in particular.

1. If we're making some breaking changes for WDL 2.0 already, I'd really like to agree with @orodeh and aim to make the distinction between `File` and `String` more strict. To me this seems easiest by removing the "magic" `File`/`String` coercions - which are a nice idea but in reality seem to produce significant levels of confusion - and replace them with more exact engine functions like, eg:
    * `File x = file_output("foo.txt")`
    * `File? y = optional_file_output("foo_y.txt")`
    * There's a lot of detail in the weeds of the engine execution as to why this is my preference, but in brief, it means significantly less logic on the part of the engine in more complex cases in predicting what is a `File` output, what is a `File?` output, what is just a `String`, etc (and thus less room for surprises, deviations between engines, non-standard edge cases...). It also hopefully means that authors can be more exact and readers can better understand what the workflow is doing.
2. I don't think Cromwell allows optional outputs to work on cloud backends today because we need to be able to predict which files to delocalize before running the task. I don't think that's insurmountable but just FYI it wouldn't work quite yet. Making the file outputs into engine functions would probably help out a lot.

orodeh:
@cjllanwarne, of course I agree :-) The only thing is that we may want to move this part of the discussion to a new thread.

jtratner:
> If we're making some breaking changes for WDL 2.0 already, I'd really like to agree with @orodeh and aim to make the distinction between File and String more strict. To me this seems easiest by removing the "magic" File/String coercions - which are a nice idea but in reality seem to produce significant levels of confusion - and replace them with more exact engine functions like, eg:

I understand it may make the engine more complex, but I really dislike adding more boilerplate to WDL.  There's enough boilerplate already - if anything I'd rather that explciitly written WDL code has to have a string suffix, like `s` or `~` or `r`:

```
task mytask {
   input {
   File whatever = "my-default-file.txt"
   String mytext = ~"data"
   }
  command <<<
     $stuff
  >>>
   output {
    File out1 = "some/path.txt"
    String whatevs = read_string("anotherpath.txt")
   }
```

Adding a `file_output` or `optional_file_output` just spews way more text into WDL workflows and makes them look uglier.

I really think that elegance is a worthwhile target.

I looked through an example complex WDL pipeline that we have defined, I looked for all instances of the quote character that were in inputs or outputs, here's how it breaks out:

```
  17 Array[File]
   4 Array[File]+
  35 File
   4 File?
  18 String
```

Of the string outputs, nearly all of them are `read_string()` kind of things.

DavyCats:
To me it would make more sense if string literals were simply quoted text and files had some kind of additional marker.

Perhaps it might be good to add a file literal and a function to convert existing strings to files:
```wdl
File file_literal = file://some_path.txt
#or
File maybe_easier_to_parse_file_literal = f"some_path.txt"
#and
String path_in_string = "some_path.txt"
File file_from_string = file(path_in_string)
```

jtratner:
@geoffjentry 

> @jtratner note that this needs to be against the development spec, not 1.0

Okay I can repoint to development spec.  I thought this was just recognizing the behavior that already exists in cromwell/dxWDL today, hence why I put it into 1.0.

I *think* someone commented that this didn't work for them, but I tested with miniwdl cromwell and example provided worked as expected:

```
# example.wdl
version 1.0
workflow wf {
    call optional_output
}
task optional_output {
    command <<<
        touch example_exists.txt
        touch arr2.exists.txt
    >>>
    output {
        File example_exists = "example_exists.txt"
        File? example_optional = "example_optional.txt"
        Array[File?] array_optional = ["arr1.dne.txt", "arr2.exists.txt"]
    }
}
```

when run, produces
```
$ miniwdl cromwell example.wdl
Cromwell input: {}
+ java -DLOG_LEVEL=warn -DLOG_MODE=pretty -jar /var/folders/fm/sjgpzb856kld04jvq82bxrcw0000gr/T/cromwell-42.jar run parent/example.wdl -o parent/20190917_122232_wf/cromwell/options.json -i parent/20190917_122232_wf/inputs.json
[2019-09-17 12:22:43,43] [warn] 'docker.hash-lookup.gcr-api-queries-per-100-seconds' is being deprecated, use 'docker.hash-lookup.gcr.throttle' instead (see reference.conf)
{
  "outputs": {
    "wf.optional_output.example_optional": null,
    "wf.optional_output.array_optional": [null, "parent/20190917_122232_wf/cromwell/cromwell-executions/wf/9b3b9091-b19f-4e32-801e-4a3d37f099d0/call-optional_output/execution/arr2.exists.txt"],
    "wf.optional_output.example_exists": "parent/20190917_122232_wf/cromwell/cromwell-executions/wf/9b3b9091-b19f-4e32-801e-4a3d37f099d0/call-optional_output/execution/example_exists.txt"
  },
  "id": "9b3b9091-b19f-4e32-801e-4a3d37f099d0"
}
{
  "outputs": {
    "wf.optional_output.example_optional": null,
    "wf.optional_output.array_optional": [
      null,
      "parent/20190917_122232_wf/cromwell/cromwell-executions/wf/9b3b9091-b19f-4e32-801e-4a3d37f099d0/call-optional_output/execution/arr2.exists.txt"
    ],
    "wf.optional_output.example_exists": "parent/20190917_122232_wf/cromwell/cromwell-executions/wf/9b3b9091-b19f-4e32-801e-4a3d37f099d0/call-optional_output/execution/example_exists.txt"
  },
  "id": "9b3b9091-b19f-4e32-801e-4a3d37f099d0",
  "dir": "parent/20190917_122232_wf"
}
```

patmagee:
@jtratner Thanks for all of the clarification. I can confirm that althoguh this is the current behaviour in some scenarios (ie local) it does work for cloud backends. Many cloud batch execution backends require output files to be declared at runtime. A simple test with `cromwell` against GCP confirms this failure.

While this indeed is a clarification, it also is changing an interpreted section of the specification and will need to be put to a vote before merging. If you are satisified with the PR as I can open it for voting

patmagee:
@jtratner This got lost in the backlog, but if you are happy with everything, I can start a vote?

jtratner:
Yep I'm happy with it - thanks! Would be awesome if you could put up for
voting. I think I got confused about which version of spec it should be
against too :-/

DavyCats:
:+1: 

These changes are in the 1.0 spec, though, but given the comments above I'm not sure if that's considered an issue in this case.

yfarjoun:
This is great! thanks! 👍

rhpvorderman:
:+1:

aednichols:
👍

illusional:
👍

patmagee:
Vote passed 5 - 0

jdidion:

PR #313, opened 2019-05-30
Specify that runtime attributes can be declared as inputs

geoffjentry:
This PR replaces #301. The primary difference is that this PR only refers to things which are already part of the spec, i.e. references to input json format remain while references to input json **files** have been elided

### DIFF
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -91,6 +91,7 @@ Table of Contents
   * [Computing Task Inputs](#computing-task-inputs)
   * [Computing Workflow Inputs](#computing-workflow-inputs)
   * [Specifying Workflow Inputs in JSON](#specifying-workflow-inputs-in-json)
+  * [Specifying / Overriding runtime attributes in JSON](#specifying--overriding-runtime-attributes-in-json)
   * [Optional Inputs](#optional-inputs)
   * [Declared Inputs: Defaults and Overrides](#declared-inputs-defaults-and-overrides)
     * [Optional Inputs with Defaults](#optional-inputs-with-defaults)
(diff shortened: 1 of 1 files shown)


illusional:
+1 for this PR :)

versions/development/SPEC.md
```diff
@@ -2647,6 +2648,23 @@ In JSON, the inputs to the workflow in the previous section might be:
 
 It's important to note that the type in JSON must be coercible to the WDL type.  For example `wf.int_val` expects an integer, but if we specified it in JSON as `"wf.int_val": "three"`, this coercion from string to integer is not valid and would result in a coercion error.  See the section on [Type Coercion](#type-coercion) for more details.
 
+## Specifying / Overriding Runtime Attributes in JSON
```
cjllanwarne: Sub `JSON` for `Inputs`?

cjllanwarne:
One thing worth making explicit one way or the other, do we want to consider these to be providable as call inputs too? ie something like...
```wdl
call foo { input:
  docker: "ubuntu: latest"
```

pshapiro4broad:
> One thing worth making explicit one way or the other, do we want to consider these to be providable as call inputs too? ie something like...
> 
> ```wdl
> call foo { input:
>   runtime.docker: "ubuntu: latest"
> }
> ```
> 
> My gut feeling is yes but I could see an argument for no

I think it should be supported, in the interest of keeping the language consistent.

patmagee:
In light of #315, I wonder if we should revist this?

pshapiro4broad:
> In light of #315, I wonder if we should revist this?

Unless I'm misunderstanding your comment, I don't think it's required. There will be cases when PR overlap, but until one is accepted I don't think it's necessary to block or update one based on the other, since neither are yet part of the spec. In this case I think the PRs are orthogonal anyway; e.g. regardless of what changes are made to the contents of the `runtime` block, the topic of a how a user can provide key/value pairs for that block can be discussed.

geoffjentry:
Sorry this is what being out of town gets you. On the flip side, you should see the view I have while typing this :)

@patmagee I was going to suggest folding in the spirit of my PR into this, if that's not already done.

patmagee:
Closing this as per @geoffjentry suggestion. Changes have been folded into the #315

PR #314, opened 2019-05-30
Clarification around input specification

geoffjentry:

PR #315, opened 2019-05-31
Clarified Runtime and Added Hints

patmagee:
## The Problem

One of the key motivators for developing WDL has always been the dream of interopable workflows. That is Workflows that can run in any execution engine, on any platform, in the exact same way and resaonably expect the same outputs.

Recently several community members have tested this theory to see just how far we are from that target. Unfortunately, WDL did not get a passing grade as it currently is. One of the primary reasons for this is the `runtime` section within a task. We have always had the desire to rework this section, moving it away from something that is a grab bag of parameters and more into minimum execution requirements. However, this was never done, and because of that, a lot of properties have crept in there which are not part of the `WDL` specification but have largely become used within the community to execute a workflow on a specific platform or infrastructure. This does not help interopability or sharing.

## My Solution

To me the best solution moving forward is to start to be a bit opinionated about what goes into the `runtime` section and ensure that there will ALWAYS be enough information present there to run a task. To this end, I have codified the keys supported within the runtime section, and have made certain properties required for every task (even if the engine supports runtime defaults). While some people may groan at the idea of this, since it means more lines to write, I think its important to remember what a `task` is meant to do. It is essentially a definition of a wrapped `command` and the execution environment that command is supposed to run in. Previously, we only required the `command` part, and allowed the engine to specify defaults if it sees fit for the `environment`. This unfortunately meant that for some tasks, there was no concrete definition of what was required to run it, immediately making the task bound to the engine and infrastructure the author intended it for. Requiring a set of minimal attributes in the runtime gets around this, without being TOO obtrusive IMO. 

I have additionally removed the ability to specify arbitrary KV pairs within the `runtime` block, and have introduced a new section called `hints`. There is a very important distinction between the two sections: The runtime defines what is REQUIRED to run the task (and engine that can provide those resources should be able to run the task), and hints are optimizations that an engine can CHOOSE to apply if they understand the meaning of the hint.

I have opted for making hints a semi-structured section similar to how the runtime used to operate in V1.0 and earlier. There are a set of reserved keys which should carry similar meaning between engines (whether the engine chooses to adopt them is completely optional), and then additionally I have added a strong emphasis on convention and reuse of arbitrary properties.


versions/development/SPEC.md
```diff
@@ -1218,81 +1270,314 @@ $runtime = 'runtime' $ws* '{' ($ws* $runtime_kv $ws*)* '}'
...
+  ...
+  hints {
+		
+  }
+}
+```
+
+#### Reserved Keys
+
+Although `hints` are arbitrary, there are a number of keys which are reserved by the language in or oder to try and encourage interoperability of tasks and workflows between different execution engines. The list of Reserved keys is likely to grow, and in order to prevent name space collisions, engines should follow the [best practices](#conventions-and-best-practices) for specifying hints.
```
cjllanwarne: Typo `or oder`

cjllanwarne:
A few other attributes exist which change behavior (and though "runtime" is a weird description for them, they're certainly not optional):

* `continueOnReturnCode`
  * If a boolean, whether to allow the task to "succeed" even if the return code is non-zero
  * If an integer, a single return code value which represents success
  * If an integer array, a list of return code values which represent success
* `failOnStderr`
  * A boolean used to indicate that the task has failed if any content exists in the stderr output

versions/development/SPEC.md
```diff
@@ -1218,81 +1270,314 @@ $runtime = 'runtime' $ws* '{' ($ws* $runtime_kv $ws*)* '}'
...
+
+* **maxCpu**: Specify the maximum CPU to be provisioned for a task. It is up to the engine, whether or not it enforces the `maxCpu` hint. The type of this hint should be the same as `runtime.cpu`
+* **maxMemory**: Specify the maximum memory provisioned for a task. It is up to the engine, whether or not it enforces the `maxMemory` hint. the type of this hint should be the same as `runtime.memory`
+* **shortTask**: Tell the execution engine that this task is not expected to take long to execute, and therefore the engine can attempt to optimize the execution in whatever way it defines. This flag can be used to tell an engine to use the cost optimized instance types that many clouds provide (but are available only for a limited time). An example of this would be `preemptible` instances on `gcp` and `spot` instances on `aws`.
+* **localizationOptional**: Tell the execution engine that whenever possible it does not need to localize the defined `File` type inputs for this task. Important to note, is this directive should not have any impact on the success or failure of a task (ie it should run with or without localization). The type of this hint is a boolean value
```
cjllanwarne: Note to anyone else who almost makes the same mistake as me:

Athough this global `localizationOptional` is like a less useful version of the Cromwell `parameter_meta` hint, keep reading because there's a separate `inputs` hints section which covers the specific-per-input version of this hint.

patmagee:
@cjllanwarne to the runtime section I added a `returnCodes` entry, which is essentially the same as the `continueOnReturnCodes`, however makes a few tweeks from what cromwell does.  

Additionally, I personally find the "failOnStdErr" a bit of an odd feature even in cromwell, which is why I did not include it in the current runtime section. I can add it here of course, however I just wanted to make the point that if a task ALWAYS returns with a 200 status, there is no reason why the user could not capture the stdout and exit if the file is nonempty.

versions/development/SPEC.md
patmagee: Fixed a few other typos.. You are eagle eyed

versions/development/SPEC.md
```diff
@@ -1218,81 +1270,316 @@ $runtime = 'runtime' $ws* '{' ($ws* $runtime_kv $ws*)* '}'
...
   }
 }
 ```
 
-The value for the `docker` runtime attribute in this case is an array of values.  The parser should accept this.  Some engines might interpret it as an "either this image or that image" or could reject it outright.
 
-Since values are expressions, they can also reference variables in the task:
+#### container (**Required**)
 
+The container key must accept one ore more locations which inform the engine where to retrieve a container image to execute the task. It is expected (but not enforced) that each container image provided are identical, and will provide the same final results when the task is run. It is the responsibility of the individual execution engine to define the specific image sources which it supports, and to determine which image is the "best" one to use at runtime. Defining multiple images enables greater portability across a broad range of execution environments.
```
cjllanwarne: I think it *should* 100% be considered invalid if different images lead to different results

versions/development/SPEC.md
```diff
@@ -1218,81 +1270,316 @@ $runtime = 'runtime' $ws* '{' ($ws* $runtime_kv $ws*)* '}'
...
 
-Since values are expressions, they can also reference variables in the task:
+#### container (**Required**)
 
+The container key must accept one ore more locations which inform the engine where to retrieve a container image to execute the task. It is expected (but not enforced) that each container image provided are identical, and will provide the same final results when the task is run. It is the responsibility of the individual execution engine to define the specific image sources which it supports, and to determine which image is the "best" one to use at runtime. Defining multiple images enables greater portability across a broad range of execution environments. 
+
+Container source locations should use the syntax defined by the individual container repository. For example an image defined as `ubuntu:latest` would conventionally refer a docker image living on `DockerHub`, while an image defined as `quay.io/bitnami/python` would refer to a `quay.io` repository. 
+
+The supported attributes are:
```
cjllanwarne: > supported attribute \*types\* are ...

versions/development/SPEC.md
patmagee: I agree, the only way to guarantee that though is 1, execute the WF under different images, or 2, assert checksums of the different images (in which case each engine might not know the actual access method). I will add stronger language here though

versions/development/SPEC.md
```diff
@@ -1218,81 +1270,316 @@ $runtime = 'runtime' $ws* '{' ($ws* $runtime_kv $ws*)* '}'
...
 }
 ```
 
-Most key/value pairs are arbitrary.  However, the following keys have recommended conventions:
 
-#### docker
+#### cpu (**Required**)
 
-Location of a Docker image for which this task ought to be run.  This can have a format like `ubuntu:latest` or `broadinstitute/scala-baseimage` in which case it should be interpreted as an image on DockerHub (i.e. it is valid to use in a `docker pull` command).
+The `cpu` key defines the _minimum_ CPU required for this task, which must be available prior to the engine starting execution. The engine does not need to provide the exact amount of CPU requested (depending on the underlying infrastructure restrictions), however it may ONLY provision more CPU than requested and not less. For example if the wdl requested `cpu: 0.5`, but only discrete values were supported, then the engine might choose to provision `1.0` cpu instead.  Values are expected to be a `Float` (where supported) or an `Int`
```
cjllanwarne: I would remove the comment about "`(where supported)`" - you already explained what happened if the platform doesn't know what a partial CPU is. I'd just say
> Values must be either `Float` or `Int`.

versions/development/SPEC.md
```diff
@@ -1218,81 +1270,316 @@ $runtime = 'runtime' $ws* '{' ($ws* $runtime_kv $ws*)* '}'
...
-    docker: "ubuntu:" + ubuntu_version
+    container: ["ubuntu:latest", "gcr.io/standard-images/ubuntu:latest"]
   }
 }
 ```
 
-Most key/value pairs are arbitrary.  However, the following keys have recommended conventions:
 
-#### docker
+#### cpu (**Required**)
```
cjllanwarne: I'm nervous about increasing the amount of boilerplate this means adding into every. single. WDL. task. definition.

Can we instead say "if a `cpu` value is not specified then the task must work with exactly '1' CPU"? And something similar for `memory`?

patmagee: Hmm, that is a good suggestion, hardcoding the default in the spec. my worry though, is that people will just ignore that entirely and continue on using things like the `default-runtime` attributes, and relying on things that are very cromwell or other engine specific. This is what lead to alot of the motivation for redefining the runtime block. I suppose we could state:

```
If a `cpu` key is not defined, then then it must be interpreted that the task will run with exactly 1 CPU.  engines should not provide a mechanism to set their own default values which circumvents this
```

versions/development/SPEC.md
```diff
@@ -1218,81 +1270,316 @@ $runtime = 'runtime' $ws* '{' ($ws* $runtime_kv $ws*)* '}'
...
+    memory: "2 GB"
   }
+}
+```
 
-  command {
-    python process.py ${arg}
+#### gpu (**Optional**)
+
+The `gpu` key provides a way to accommodate modern workflows which are increasingly becoming reliant on GPU computations. This key is a `true` or `false` value and simply indicates to the engine that a task requires a GPU to run to completion. A task with this flag set to `true` is guaranteed to only run if a GPU is a available within the runtime environment. It is the responsibility of the engine to check prior to execution whether a GPU is provisionable, and if not, preemptively fail the task.
```
cjllanwarne: Maybe someone with more experience using GPUs can chime in:

* Is "I just need a GPU, I don't care what type" enough information? (ie is it possible to be agnostic about the number, type and driver version required?)
* Is having a GPU a genuine requirement for running such a task
  * ... or is it more a case of "if there isn't a GPU the task will still succeed but will be inefficient?"

patmagee: That is what I am a little naive about, so I will rely on someone else to have more information. I was trying to capture the requirement for a GPU without throwing in all of the provider specific terminology in this part of the block.

versions/development/SPEC.md
```diff
@@ -1218,81 +1270,316 @@ $runtime = 'runtime' $ws* '{' ($ws* $runtime_kv $ws*)* '}'
...
+```
+
+
+#### Conventions and Best Practices
+
+In order to encourage interopable workflows, wdl authors and engine creators should view the `hints` section strictly as an optimization that can be made for a specific task at runtime, and they should not view it as a set of requirements for that task. By following this principle, we can guarantee that a workflow is runnable on all platforms assuming the `runtime` block has the required parameters, regardless of what is contained within the hints section.
+
+The following is a basic set of guidelines that engines and authors should follow when writing hints:
+
+1. A hint should never be required
```
cjllanwarne: `should` => `must`?

versions/development/SPEC.md
```diff
@@ -2127,6 +2465,12 @@ task myTask {
       String name = a.name + "Potter"
       Int age = a.age * 2
     }
+    
+    runtime {
+		container: "my_image:latest"
+		cpu: 1
+		memory: "1 GB"
+	}
```
cjllanwarne: The tabulation looks a little bit off here

versions/development/SPEC.md
patmagee: @cjllanwarne the other idea I had, was to allow for a runtime section in the 'workflow' definition. This would allow for the definition of runtime defaults which are explicitly stated and it Also provides a mechanism for documenting them.

illusional:
Really excited that docker has been renamed to container (while still defaulting to Dockerhub), I think this is a super valuable step towards agnosticism of containers.

versions/development/SPEC.md
```diff
@@ -1218,81 +1270,316 @@ $runtime = 'runtime' $ws* '{' ($ws* $runtime_kv $ws*)* '}'
...
+#### disks (**Optional - but strongly encouraged**)
+
+The `disks` key provides a way to request one or more persistent volumes of at least a specific size and mounted at a specific location. When the disks key is provided the engine must guarantee the resources requested are available, or immediately fail the task prior to execution. This property does not specify exactly what type of persistent volume is being requested (ie SSD, HDD), but leaves this up to the engine to decide, based on what hardware is available or a `hints` value.
+
+The general format of a `disks` value is: `</absolute/mount/point> <Size in GB>`. A WDL may leave out the mount point and only provide the size in GB that is being requested for ONE disks entry. This should be interpreted by the engine as a persistent volume mounted at the root of the execution directory within a task. If more then one disks entries for a task leaves out the mount point, the engine should fail the tasks.
+
+There are several ways to specify the `disks` attribute:
+
+* `Int` - GB of disk space to request, eg: `100`, `200`.
+* `String` (`"<size>" || "<mount-point> <size>"`)
```
DavyCats: Should some system similar to the one for memory be allowed here? ie. specifying whether the size is in KBs or MBs or TBs.

versions/development/SPEC.md
```diff
@@ -1218,81 +1270,316 @@ $runtime = 'runtime' $ws* '{' ($ws* $runtime_kv $ws*)* '}'
...
 ```
 
-#### memory
+#### memory (**Required**)
 
-Memory requirements for this task.  Two kinds of values are supported for this attributes:
+The `memory` key defines the _minimum_ memory required for this task which must be available prior to the engine starting execution. The engine does not need to provide the exact amount of memory requested, however it may ONLY provision more memory then requested and not less. For example, if the wdl requested `1 GB` but only blocks of `4 GB` were available, the engine might choose to provision `4.0 GB` instead.  Two kinds of values are supported for this attribute:
 
 * `Int` - Interpreted as bytes
 * `String` - This should be a decimal value with suffixes like `B`, `KB`, `MB` or binary suffixes `KiB`, `MiB`.  For example: `6.2 GB`, `5MB`, `2GiB`.
```
DavyCats: Should there or should there not be a space between the number and the suffix? Or does it not matter?

rhpvorderman:
Wow! Great PR! 
We have exactly this problem. We are running on a SGE cluster where java jobs require insane memory allocations not to get killed. So the [biowdl/tasks](https://github.com/biowdl/tasks) sharing is hindered by that. Having a way to specify this in a more general way is a blessing. Thanks a lot @patmagee !

versions/development/SPEC.md
patmagee: @cjllanwarne on second thought, I Think I prefer enforcing an implicit definition of a default within the WDL as opposed to allowing defaults at the `workflow` level. Not every task will have a workflow, therefore people will likely abuse this approach for tasks and just assunme a user will define the default value

versions/development/SPEC.md
patmagee: Semantically that makes sense. A disk is a a size, and sizes can be specified in blocks. In practice however I feel like we will only ever see `GB` or `TB` definitions (since specifying a disk in MB seems archaic). And from that we will NORMALLY only see `GB`. So I figured the simplification makes sense. Wheras with memory, people can and regularily do specify both GB and MB for the size

orodeh: To clarify, are disk specifications like `"100GB`, or `"3TB"` valid? 

In my personal opinion, they should be legal.

versions/development/SPEC.md
```diff
@@ -1218,81 +1270,316 @@ $runtime = 'runtime' $ws* '{' ($ws* $runtime_kv $ws*)* '}'
...
+```wdl
+task gpu_test {
+  #.....
+  runtime {
+    gpu: true
   }
+}
+```
+
+
```
orodeh: I suggest having a default flag of `false` for this field. Otherwise, we would be a making tasks more verbose than necessary.

orodeh:
I think this is good. My only comment is that I like Cromwell's default runtime attributes, and they are supported by dxWDL. I would prefer to avoid adding required boilerplate to WDL tasks, if it isn't strictly necessary.

patmagee:
@orodeh As a language, this is one thing that we ultimately need to resolve. the `runtime attributes` is one of the culprits IMO behind making a workflow completely unrunnable in a different environment. It provides a way for users to move task configuration outside of the actual `task`/`wdl` and completely detracts from the portability of that workflow.   The balance I tried to strike can be seen in this comment: https://github.com/openwdl/wdl/pull/315#discussion_r289534699, where users can avoid setting a value but the task MUST be run with the defaults if there is nothing set. This guarantess portability

orodeh:
@patmagee. Very well, can we then make defaults for cpu, memory, and disk, so that the only thing you absolutely must have in the runtime section is the container image?

versions/development/SPEC.md
orodeh: Just to provide some color here.  

The GPU instance types DNAnexus supports today, on AWS, are the g2 and p3 families. The p3 is the current generation. To allow users to actually make use of these, we install an NVIDIA driver. An alternative is nvidia-docker.

patmagee:
@orodeh yes, I am open to suggestions on what those values should be. for CPU I would think that 1 is fine, disk I would so 10GB but am a bit hazy on memory

versions/development/SPEC.md
patmagee: @orodeh does the user need to know specifics about this? or is it enough for them to know that a GPU was added, and then allow additional config in the hints section?

patmagee: my motivatin  here was to abstract as much of the specific vendor terminology into the hints section, then have an engine guarantee a gpu was added. the flavor and size of the gpu would need to be either configured in the hints or the engine would need to make a resonable default value here

orodeh: I think it is a good start to say have a GPU flag in the runtime section, and leave the rest for hints.

orodeh:
How about 2GiB as a default number for memory.

patmagee:
that sounds reasonable to me

rhpvorderman:
> disk I would so 10GB 

This very much depends on the size of the input file. If I am running on a whole-genome sample and my input fastqs are 100 GB, then I need 100GB of disk space as well for the output when using cutadapt.

But If I run a test workflow with 200 kb files I don't want cromwell to crash by default because there is no 10GB available. All my test workflows run in /tmp by default ([pytest-workflow](https://github.com/lumc/pytest-workflow/)'s default).  I do have some spare gigabytes on `/` but not 10. 10 is a bit overly cautious IMHO.

Why not set it at 1GiB ?  We have a `size(FILE)` operator in WDL so it is extremely easy to set the requirements according to the input files. I think it would be good to promote that as best practice.

patmagee:
@orodeh I have added the default values with restrictions on execution engines providing their own defaults. WDYT

versions/development/SPEC.md
```diff
@@ -1273,7 +1259,7 @@ $runtime_kv = $identifier $ws* '=' $ws* $expression
...
 
-During execution of a task, resource requirements within the runtime section must be enforced by the engine. If the engine is not able to provision the requested resources, then the task should immediately fail prior to the start of execution. In order to ensure adequate information is provided to execute a task, there are a number or required keys, which are defined below.
+During execution of a task, resource requirements within the runtime section must be enforced by the engine. If the engine is not able to provision the requested resources, then the task should immediately fail prior to the start of execution. In order to ensure adequate information is provided to execute a task, there are is one required keys, and strict default values for optional keys, which are defined below. Default values for all optional keys are directly defined by the WDL specification in order to encourage portability of workflows and tasks. Execution engines should NOT provide additional mechanisms to set _default_ values for when no runtime attributes are defined.
```
orodeh: there are is one required keys ---> there is one required key

orodeh:
@patmagee thank you. I think that one compulsory key is the best we are going to be able to get.

versions/development/SPEC.md
```diff
@@ -46,8 +46,15 @@ Table of Contents
         * [Task portability and non-standard BaSH](#task-portability-and-non-standard-bash)
     * [String Interpolation](#string-interpolation)
     * [Runtime Section](#runtime-section)
-      * [docker](#docker)
-      * [memory](#memory)
+      * [container](#container-required)
+      * [cpu](#cpu-required)
+      * [memory](#gpu-optional)
```
cjllanwarne: This looks like a copy/paste typo

versions/development/SPEC.md
```diff
@@ -1218,81 +1256,343 @@ $runtime = 'runtime' $ws* '{' ($ws* $runtime_kv $ws*)* '}'
...
   }
 }
+
 ```
 
-Most key/value pairs are arbitrary.  However, the following keys have recommended conventions:
+#### memory (**Optional**)
 
-#### docker
+The `memory` key defines the _minimum_ memory required for this task which must be available prior to the engine starting execution. The engine does not need to provide the exact amount of memory requested, however it may ONLY provision more memory then requested and not less. For example, if the wdl requested `1 GB` but only blocks of `4 GB` were available, the engine might choose to provision `4.0 GB` instead.  Two kinds of values are supported for this attribute:
```
cjllanwarne: One side effect of a one-size-fits-all `memory` attribute which I've spotted is people end up having to write a "safe" amount of memory here, rather than the "should work 80% of the time" amount (and allow the engine to try again with more if it needs to).

My gut feeling is that the definition of memory here should actually be something more like "feel free to try with less memory at first if you really want to, but if you give the tool this much memory and it still fails, fail the task". My thinking is:
* The simplest engine implementation is to provide `memory` and fail if the task doesn't succeed
* More complicated engine implementations still get a ceiling that they know that it's not worth going above

Is that fair?

versions/development/SPEC.md
```diff
@@ -46,8 +46,15 @@ Table of Contents
         * [Task portability and non-standard BaSH](#task-portability-and-non-standard-bash)
     * [String Interpolation](#string-interpolation)
     * [Runtime Section](#runtime-section)
-      * [docker](#docker)
-      * [memory](#memory)
+      * [container](#container-required)
+      * [cpu](#cpu-required)
```
cjllanwarne: This is no longer required

versions/development/SPEC.md
cjllanwarne: One other thought: by allowing engines to be *more* generous, do we risk running into the exact same portability problems we have now:

* I write a WDL without any `memory` specification.
* My WDL works on my engine, because my engine gives me `10GiB` of ram even though the "default" is only `2GiB`
* I publish my WDL and someone else tries to run it, but fails, because their engine respects the default `memory` amount of `2GiB` but my task really needs more than that?

patmagee: One size fits all memory is an issue that I have come across to, anmd the only reliable solution I have found is generally trial and error. I think this is a difficult point. How do we define "less memory" . Trying with less and resulting in a failure could be a costly mistake that the user now has to pay for if they are in a cloud environment. 

To me resource requirements are minimum guarantees. As a task creator I 100% known that my Task will minimally be given the requested memory, and therefore I can design the task with that in mind. The idea of course is I have tested the command I am running in multiple ways (this might be alot to ask), and have Identified the appropriate amount of memory needed. The `maxMemory` is a flag which they can use in the `hints` section to specify the upper bound that it should be given. Allowing memory to be below this removes any sort of guarantees and now makes the execution of the task a bit unpredictable.

patmagee: @cjllanwarne I struggled with this as well but I am not sure if there is any way around it. The nature of the beast is that different infrastructure will assign memory in different blocks. Ie Cromwell uses the google machine types which are all discrete values.

There must be a balance that we can strike here, and maybe that balance comes from moving the `maxMemory` into the `runtime` section. If the maxMemory is not set then the implication is the task will be assigend at least what is defined by the memory `memory` key. If the Execution engine does not support that specific memory amount, then it must assign the closest value?

patmagee: The other thing would potentially be to make the `maxMemory` a default of 2* the `memory` or some other feature like that

patmagee: If it task runs out of memory, I am not sure itsd the responsibility of the execution engine to retry that task. It might be, but you could also argue that it is up to the user whether or not they want that retry to take place

cjllanwarne: Hmmm... it seems like the only way to **guarantee** portability is therefore to _require_ engines to retry failed tasks due to OOMs (potentially capped by a `maxMemory` hint if one is provided) - in other words make the scheme proposed in https://github.com/broadinstitute/cromwell/pull/5017 a requirement of the spec?

patmagee: I think I would favor an implicit maxMemory over required retries. Not every engine will know if a task failed because of an oom issue. Not to mention that approach violates the maxRetries policy,  which if set to 0 forbids retried of a task (for any reason).

cjllanwarne: That sounds ok, I guess I don't mind what we call the idea of a maximum as long as it exists and is enforced. I think we're actually proposing almost exactly the same thing, but with different names:

* I would have `memory` be a maximum/"guaranteed success" value, and have a separate `try_this_memory_first` hint to tell the engine there might be a better/lower average value to try first. 
* You would have an implicit `maxMemory` value which an engine cannot go above, and use `memory` as a "don't go below this value" floor. But for an engine to be sure that it can run a WDL written against some other engine, it must eventually be willing to go up as high as `maxMemory` after a few tries, even if it's not its first preference.

The tie-in with `maxRetries` is interesting... Maybe the implied behavior is to divide the difference between min and max memory into an appropriate number of chunks so that the final retry happens to also use the `maxMemory` value.

orodeh:
The dnanexus engine does not retry by default with increasing memory/disk sizes. I prefer @patmagee suggestion, which is putting minimal cpu/memory/disk requirements in the runtime section, and maximum values in the (new) hints section. 

I may be wrong, but I believe Cromwell could implement the retry-with-different-sizes behavior comfortably within this framework. It is a nice feature to have, for sure, I would rather not have it be a compulsory part of the standard.

cjllanwarne:
@orodeh my problem with just implementing it like that is the portability of the resulting workflows. If WDL authors at the Broad are assuming that `memory` is going to be doubled by Cromwell for them when they write their tasks, but dnanexus doesn't do that, then we're back to encouraging people to write WDLs that will only ever work on a single engine.

I'll try to make my objection clearer with an example which sounds extreme but IMO is equivalent to what's being suggested above:

Imagine Engine "X" introducing an `x_memory` hint which only it knows about, and which contains the real value which Engine X will _actually_ pass to the backend. Nothing goes wrong for people writing workflows against Engine X since they put the right amount of memory to use in the `x_memory` hint and leave the `memory` value at the default floor of `2GB`. Engine X is compliant because it always provides at least that default floor of 2GB in `memory`. But if I take a workflow with an `x_memory` hint to dnanexus or Cromwell the workflow fails because 2GB is not nearly enough to run the task for real.

orodeh:
@cjllanwarne, this is true, and the concrete example helps illustrate the situation. 

I believe this is just a limitation of the portability that we can reasonably expect to get. I don't think this is totally solvable.

patmagee:
@cjllanwarne I think what you being up is a good example of a bad task. Which people do write. They created a task which does not actually run with the given memory requirements that they specified. In on sense the only way to truly avoid this is by properly testing a task and benchmarking different values. 

I think though that specifying a minimum value for memory moves us forward from where we were before. Previously there was no expectation of what no memory or CPU in the runtime block would be interpreted as. Now at least we have a definition of what we can minimally expect , so it's a step closer.

cjllanwarne:
I don't agree that this is an intractable problem. The suggestion above of a "`maxMemory` _requirement_ in `runtime` (with a default of whatever `memory` is unless otherwise specified)" would alleviate my concerns.

With equivalents maximums for CPU, disk, etc, too, probably

versions/development/SPEC.md
```diff
@@ -1218,81 +1257,343 @@ $runtime = 'runtime' $ws* '{' ($ws* $runtime_kv $ws*)* '}'
...
+The runtime section defines a set of key/value pairs which represent the minimum requirements needed to run a task or, which define the conditions under which a task should be interpreted as a failure or success. All keys within the runtime section are well defined, and must be honored by the execution engine. This also means that arbitrary key/value pairs within the runtime section are not allowed, and should be rejected by an engine. Arbitrary Key/value pairs should be added to the `hints` section instead.
+
+During execution of a task, resource requirements within the runtime section must be enforced by the engine. If the engine is not able to provision the requested resources, then the task should immediately fail prior to the start of execution. In order to ensure adequate information is provided to execute a task, there is one required key, and strict default values for optional keys, which are defined below. Default values for all optional keys are directly defined by the WDL specification in order to encourage portability of workflows and tasks. Execution engines should NOT provide additional mechanisms to set _default_ values for when no runtime attributes are defined.
```
geoffjentry: The capitilaztion of `NOT` in `should NOT` seems like a weird emphasis when one considers RFC 2119 (ie `should not` already has a definition):

```
SHOULD NOT   This phrase, or the phrase "NOT RECOMMENDED" mean that
   there may exist valid reasons in particular circumstances when the
   particular behavior is acceptable or even useful, but the full
   implications should be understood and the case carefully weighed
   before implementing any behavior described with this label.
```

versions/development/SPEC.md
```diff
@@ -1218,81 +1257,343 @@ $runtime = 'runtime' $ws* '{' ($ws* $runtime_kv $ws*)* '}'
...
+
+Values can be any expression which evaluate to a valid entry for that runtime property. Since values are expressions, they can also reference variables in the task:
```
geoffjentry: s/can/may (also referencing RFC 2119)

versions/development/SPEC.md
```diff
@@ -1218,81 +1257,343 @@ $runtime = 'runtime' $ws* '{' ($ws* $runtime_kv $ws*)* '}'
...
   }
 }
 ```
 
-The value for the `docker` runtime attribute in this case is an array of values.  The parser should accept this.  Some engines might interpret it as an "either this image or that image" or could reject it outright.
 
-Since values are expressions, they can also reference variables in the task:
+#### container (**Required**)
+
+The container key must accept one ore more locations which inform the engine where to retrieve a container image to execute the task. It is expected that all container images provided are identical. They MUST produce the same final results when the task is run. It is the responsibility of the individual execution engine to define the specific image sources which it supports, and to determine which image is the "best" one to use at runtime. Defining multiple images enables greater portability across a broad range of execution environments.
```
geoffjentry: Similarly to above, I'm unclear on the need for capitalization (or more precisely, inconsistent capitalization)

versions/development/SPEC.md
```diff
@@ -1218,81 +1257,343 @@ $runtime = 'runtime' $ws* '{' ($ws* $runtime_kv $ws*)* '}'
...
+  runtime {
+    container: ["ubuntu:latest", "gcr.io/standard-images/ubuntu:latest"]
   }
+}
+```
+
+
+#### cpu (**Optional**)
+
+The `cpu` key defines the _minimum_ CPU required for this task, which must be available prior to the engine starting execution. The engine does not need to provide the exact amount of CPU requested (depending on the underlying infrastructure restrictions), however it may ONLY provision more CPU than requested and not less. For example if the wdl requested `cpu: 0.5`, but only discrete values were supported, then the engine might choose to provision `1.0` cpu instead.  Values are expected to be a `Float` or an `Int`
```
geoffjentry: `are expected to be`: is there a reason this wouldn't be a `must be`?

versions/development/SPEC.md
geoffjentry: @cjllanwarne With user defined overrides that same behavior is captured IMO

geoffjentry:
@patmagee On the assumption that #314 passes, would you be able to fold in the concepts from #313? And then we could just close #313

geoffjentry:
@patmagee Confirming that #314 did indeed pass & merge in

illusional:
This might be more the behaviour of Cromwell, but I think the terminology of [Optional literals](https://github.com/openwdl/wdl/blob/master/versions/development/SPEC.md#optional-literals) is a little confusing in the realm of WDL (and is similar to null vs undefined in JS). 

- An optional parameter that can take a null (eg: `String? param1`)
- A parameter that is non-optional but has a default (eg: `String param2 = "default value"`)
- The combination: an optional attribute that has a default (eg: `String? param3 = "default value"`) 

It's around the difference of an optional parameter and a parameter with a default that is "optional" to provide. This might not be related so much to this PR, but it comes through my usage of the `runtime.cpu` and `runtime.memory` attributes.

The runtime CPU and memory attributes are considered `**Optional**`, but this semantic in WDL means that you could pass a `None` value to it, where it seems actually declared as `Int cpu = 1` which if you pass a None to this value, it fails.

patmagee:
@geoffjentry Sorry about that, I did not see your comment. I have folded in the changes you were hinting at! 

@illusional I see what you mean, but its important to note that the values defined for runtime are not strictly optional. since if not defined a defualt is ALWAYS set

orodeh:
👍

aednichols:
👍

illusional:
👍

DavyCats:
:+1:

versions/development/SPEC.md
dinvlad: Wouldn’t it be sufficient to just compare image SHA in the engine?

patmagee: @dinvlad the image SHA is not required to be passed in by the user, and I think would actually be to cumbersome a restriction. The best approach IMO is just making it a non enforced requirment

dinvlad: @patmagee sorry, I guess what I meant is more on the implementation side (so perhaps it doesn't belong to this discussion). IIRC, engine implementations like Cromwell use image SHA in cache comparisons anyway. So this would be a low-hanging fruit to just compare SHAs that are already collected for these images anyway (I'm not talking about having to specify SHA inside the image URL, but rather the native caching mechanism already retrieves it from the Docker repo, if I understand that correctly).

geoffjentry: @dinvlad Keep in mind that the period for comments. At this point the PR is set in stone and we're voting on it

rhpvorderman:
:+1: Thanks for putting all the work in @patmagee! And the community also did a great job reviewing this. Thanks all!

geoffjentry:
:+1:

patmagee:
Voting is now closed. This pr passed with a vote of 6 - 0

versions/development/SPEC.md
```diff
@@ -1218,81 +1259,343 @@ $runtime = 'runtime' $ws* '{' ($ws* $runtime_kv $ws*)* '}'
...
+
+
+#### cpu (**Optional**)
+
+The `cpu` key defines the _minimum_ CPU required for this task, which must be available prior to the engine starting execution. The engine does not need to provide the exact amount of CPU requested (depending on the underlying infrastructure restrictions), however it may ONLY provision more CPU than requested and not less. For example if the wdl requested `cpu: 0.5`, but only discrete values were supported, then the engine might choose to provision `1.0` cpu instead.  Values must be be a `Float` or an `Int`
+
+
+##### Default Value
+
+If the `cpu` key is not defined, then the engine must set a default value of `1` CPU. The task, therefore will be guaranteed to run with 1 CPU. The engine must not provide a mechanism to set their own default value for disks which circumvents this, ie `runtime-defaults`.
```
jdidion: You have a copy-paste error here and elsewhere - s/disks/cpu/

versions/development/SPEC.md
```diff
@@ -1218,81 +1259,343 @@ $runtime = 'runtime' $ws* '{' ($ws* $runtime_kv $ws*)* '}'
...
+task maxRetries_test {
+  #.....
   runtime {
-    memory: "2GB"
+    returnCodes: "*"
   }
 }
 ```
 
+### Hints Section
```
jdidion: I know it's a bit late, but I'd like to suggest "reusable" as a reserved hint. Many execution engines have a concept of reusable results, i.e. for the exact same inputs on the exact same task, previously available outputs can be reused rather than re-running the task. The default value should be "true" but this will not be appropriate for a subset of tasks, and it may be desirable to (temporarily) turn this off for the entire workflow, e.g. when debugging.

patmagee: @jdidion unfortuinately no changes can be made to the current PR. Any changes we make need to be thropugh a separate PR process. If you would like to create a PR suggesting reusable as a runtime hint we would greatly welcome it!

versions/development/SPEC.md
```diff
@@ -1218,81 +1259,343 @@ $runtime = 'runtime' $ws* '{' ($ws* $runtime_kv $ws*)* '}'
...
+
+* **maxCpu**: Specify the maximum CPU to be provisioned for a task. It is up to the engine, whether or not it enforces the `maxCpu` hint. The type of this hint should be the same as `runtime.cpu`
+* **maxMemory**: Specify the maximum memory provisioned for a task. It is up to the engine, whether or not it enforces the `maxMemory` hint. the type of this hint should be the same as `runtime.memory`
+* **shortTask**: Tell the execution engine that this task is not expected to take long to execute, and therefore the engine can attempt to optimize the execution in whatever way it defines. This flag can be used to tell an engine to use the cost optimized instance types that many clouds provide (but are available only for a limited time). An example of this would be `preemptible` instances on `gcp` and `spot` instances on `aws`. "Short" is a bit relative, but should generally be interpreted as << 24h.
```
jdidion: Is this preferable to something more granular? For example, this could instead be "duration" which is the expected duration of the task to some degree of specificity. The specificity could be categorical ("short", "medium", "long") or a continuous value ("48h").

DavyCats: I actually like @jdidion's idea here. Some grids/clusters require you to set a maximum runtime of a job or choose a specific  queue based on how long a job takes. An engine could use a `duration` runtime attribute to make such choices in a more flexable manner then with a single flag specifying a task is "short".

patmagee: @DavyCats part of me wanted to avoid that on the first pass of this, mainly becuase thats a level of granularity that, although useful, might not be available when programatically running jobs with diverse  inputs. Additionally, it is not really applicable except in the case of Some grid/clusters. The runtime block is the "Required" section and I really dont want to have a required `minTime` and `maxTime` for EVERY execution on every platform since its only required in maybe 5% of situaations. Its harder to set sensible defaults for those as well. I would rather `minTime` and `maxTime` be part of the `hints` section then then the `runtime` section

jtratner:
To be clear, the flag and the semantics suggested for it (job < 24h
runtime) only make sense on GCP.

On AWS, spot instances can (potentially) run indefinitely, given a high
enough bid.

versions/development/SPEC.md
illusional: Chiming in later, but I agree that a `duration` / `maxTime` field would be less relative, and allow for better scheduling of jobs. For instance, today we were considering that if a job is less than 30 seconds, we might run it on the same instance as Cromwell to avoid spinning up or submitting a new job. For GCP, if the job is less than 24 hours you might consider submitting it as a preemptible job. 

You could always write an expression here to derive the duration based on the size or number of inputs etc.

versions/development/SPEC.md
```diff
@@ -2647,6 +3026,39 @@ In JSON, the inputs to the workflow in the previous section might be:
 
 It's important to note that the type in JSON must be coercible to the WDL type.  For example `wf.int_val` expects an integer, but if we specified it in JSON as `"wf.int_val": "three"`, this coercion from string to integer is not valid and would result in a coercion error.  See the section on [Type Coercion](#type-coercion) for more details.
 
+## Specifying / Overriding Runtime Attributes in JSON
```
pieterlukasse, line 3029: 🎉  👍

jdidion:
This is now implemented in wdlTools

patmagee:
@jdidion incredible news!! I can coordinate merging this and resolving any conflicts

mlin:
Hi all, belated discussion point here. The PR is written with the `hints` section grammar modeled on the `meta` sections, meaning the right-hand side values are literals only, and can *not* be WDL expressions with calculations or references to other declarations. That's in contrast to the `runtime` section where expressions are useful & widely used. It would seem curious that you can write a calculation for `runtime.memory` but not `hints.maxMemory`. 

On the other hand, if we want to allow general expressions for the hints values, there's a corner-case grammar collision between map literals and the nested runtime section, e.g. if we wrote

```
String bar = "xyz"
hints {
  inputs: { bar: true }
}
```

it's grammatically ambiguous whether the right-hand side of inputs: is a nested hints section or a `Map[String,Boolean]`. Various solutions are possible but we'd have to settle on one.

mlin:
Also importing a discussion point from https://github.com/openwdl/wdl/discussions/406#discussioncomment-121420 with @patmagee @jdidion @illusional.

`runtime.container` is just an alias for `runtime.docker` right now, presumptively referring to a docker registry/tag. How would one specify that another kind of container/image is meant? That seems to me important if the goal is portable generalization to other container tech like Singularity, and thus shouldn't be left to engine configuration. More structure might also be needed for the future possibility of supplying a Dockerfile (or Singularity recipe) instead of a prebuilt image tag, which really streamlines workflow "devops" for simple cases (also provided by CWL).

The concern IMO is introducing `runtime.container` in a way that's just an alias for `runtime.docker` at first, and whether that's going to place constraints on the likely need for more structure in the future.

rhpvorderman:
One solution could be that the `container` field always needs to specify a protocol. I.e. `docker://biowdl/chunked-scatter:0.1.0` or `singularity://ubuntu:16.04`. Or the container field becomes a map:
```WDL
runtime {
    container: {protocol: "docker", image: "biowdl/chunked-scatter:0.1.0"}
} 
```
The last solution also allows for several fields  if the need arises, but since I can not think of any use cases right now, I prefer the simpler version with the protocol.

jdidion:
I like the simple solution. For better backward compatibility, if there is no protocol, assume Docker.

jdidion:
(I was worried about having to escape URIs with colons in them, but apparently colons are allowed without escaping https://stackoverflow.com/questions/1737575/are-colons-allowed-in-urls#:~:text=Yes%2C%20unless%20it's%20in%20the,org%2Fwiki%2FTemplate%3AWelcome)

patmagee:
I agree with @jdidion, in this case simple is better. If we allow a protocol to be defined, and by default use docker that probably captures the vast majority of use cases

mlin:
* implied `docker://` saounds good to me; I noticed the pull commands in singularity, podman, and rkt have similar schemes
* nit: if we're trying to avoid backwards incompatibility for "WDL 1.1" then we mustn't declare `runtime.container` "required" as it is now (nor can we forbid other arbitrary runtime fields yet)
* bumping https://github.com/openwdl/wdl/pull/315#issuecomment-722785580 the previous comment on hints grammar

jdidion:
@mlin I agree on your point 2 - this is explicitly stated in the RFC (container is an alias, not a replacement; arbitrary keys are deprecated but still allowed)

mlin:
Here are several ideas for how to allow expressions in hints. This is a tedious issue obviously, but probably much less that it would be ex post facto!

0. Keep as-is -- don't allow them. (A valid answer, but might lead to regret?)
1. Make hints just like runtime, without any meta-style nesting. Say custom hints should give their names appropriate prefixes (e.g. `aws_`, `azure_`) to avoid conflict.
2. Make all hints exist at exactly one level of nesting (a namespace), so every hint is shaped like `hints { namespace { name: <expr> } }`
3. Say expressions are allowed in hints "except for map literals" so the `key: value` syntax always represents meta-style nesting
4. Allow expressions in hints only within `~{}` interpolations
5. from @jdidion on slack:
> require the user to define a struct for any object-like value they want to put in hints, and then use the struct literal syntax
> or for engine-specific hints, the engine pre-defines the struct format and other engines just ignore any hints they don’t recognize (including not trying to type-check their values)

jdidion:
@mlin I actually think #3 is the best solution. In fact, I'd be in favor of also disallowing map literals in runtime and the meta sections as well - it just makes parsing more difficult and confusing to the end user for very little benefit.

My second choices would be 0 or 5, and I'm pretty opposed to 1 and 2.

versions/development/SPEC.md
```diff
@@ -1218,81 +1259,343 @@ $runtime = 'runtime' $ws* '{' ($ws* $runtime_kv $ws*)* '}'
...
-    docker: ["ubuntu:latest", "broadinstitute/scala-baseimage"]
+    container: ubuntu_version
   }
 }
 ```
 
-The value for the `docker` runtime attribute in this case is an array of values.  The parser should accept this.  Some engines might interpret it as an "either this image or that image" or could reject it outright.
 
-Since values are expressions, they can also reference variables in the task:
+#### container (**Required**)
```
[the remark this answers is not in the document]
jdidion: Oh yeah - I agree that it shouldn't be required in 1.1, and I think there's a good argument to not ever make it required, for example if I just want to run my workflow locally in an environment where I already have all the dependencies installed.

jdidion:
Merged as part of #435

PR #316, opened 2019-06-03
Cycles are not allowed in the call graph

orodeh:
Adding a clarification to the `Call Statement` section, that cycles in the call graph are not allowed.


cjllanwarne:
I think that the changes in https://github.com/openwdl/wdl/pull/257 should cover this? I haven't seen any conversation in that PR for a while so maybe it's worth opening that one up to voting instead?

orodeh:
I voted for #257, I can close this one, if there are no objections.

PR #318, opened 2019-06-14
[trivial] standardize several occurrences of Integer to Int

mlin:
There are several occurrences of `Integer` in the spec which seemingly should be `Int` since that's the spec-defined type name. (Not sure if `Integer` is vestigially supported in Cromwell?)


geoffjentry:
I'm going to merge this as-is as this is clearly an error. Technically it **does** alter the spec, but ...

PR #320, opened 2019-06-28
[trivial] fix brackets in transpose() example

mlin:


geoffjentry:
Merging this w/o a vote. While it does technically modify the spec it's obviously a typo as it's referencing `Array` which has the bracket syntax

PR #321, opened 2019-07-15
Link to relevant Emacs packages for syntax highlighting and indentation

jmonlong:
Adding Emacs to the list of highlighters that makes writing WDL easier.

### DIFF
--- a/highlighters/emacs/README.md
+++ b/highlighters/emacs/README.md
@@ -1,0 +1,6 @@
+## WDL in Emacs
+
+The following packages are available through [MELPA](https://www.emacswiki.org/emacs/MELPA):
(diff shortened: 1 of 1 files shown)


geoffjentry:
@jmonlong This is awesome, thanks! As it's not a spec change, merging as-is

dinvlad:
We've also developed Language Server Protocol implementation for WDL, so any LSP client for Emacs should be able to pick that up. This way you can get full syntax validation as well.

jmonlong:
Thanks, that looks very interesting, I didn't know about LSP and lsp-mode but I'll try reading up on it and give it a try.

PR #324, opened 2019-07-25
Fix a potential syntax issue with the WDL SPEC

rexwangcc:
I'm not sure if I should use `master` or `develop` as the merge-base.


geoffjentry:
`master` it is. Thanks @rexwangcc

PR #326, opened 2019-07-30
Updated Core team members

patmagee:
Just a bit of cleanup on some recent changes.
@orodeh @cjllanwarne can you please confirm that all of your information is correct?


geoffjentry:
I could have sworn I did this already but apparently not. Good catch

orodeh:
My information is correct, thanks.

PR #331, opened 2019-08-30
Global String Interpolation

patmagee:
One of the things which I have found a bit fuddling is the contextual usage of string interpolation within a task and not within a workflow. It does not seem like there is necessary anything that is being done which is ultimately task specific and I am of the opinion we should generalize this behaviour to be able to use string interpolation everywhere. I have seen alot of workflows attempt to use String interpolation in a workflow which I need to advise them is not possible.

This issue also came up because I am in the process of re-writting the grammar/lexer as an EBNF using `antlr4`.  From a grammar perspective handling the different contexts is awkward and actually requires a lot of excessive code duplication.

## These Changes

This PR simply changes the context which string interpolations can be called from. They are now globally callable and are not bound to a task.

### DIFF
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -9,6 +9,11 @@ Table of Contents
   * [Global Grammar Rules](#global-grammar-rules)
     * [Whitespace, Strings, Identifiers, Constants](#whitespace-strings-identifiers-constants)
     * [String Interpolation](#string-interpolation)
+      * [Interpolating and concatenating optional strings](#interpolating-and-concatenating-optional-strings)
+      * [String Interpolation Options](#string-interpolation-options)
+        * [Sep](#sep)
+        * [True and False](#true-and-false)
+        * [Default](#default)
     * [Comments](#comments)
     * [Types](#types)
       * [Custom  Types](#custom--types)
@@ -83,7 +88,6 @@ Table of Contents
 * [Namespaces](#namespaces)
 * [Scope](#scope)
 * [Optional Parameters &amp; Type Constraints](#optional-parameters--type-constraints)
-  * [Prepending a String to an Optional Parameter](#prepending-a-string-to-an-optional-parameter)
 * [Scatter / Gather](#scatter--gather)
 * [Variable Resolution](#variable-resolution)
   * [Task-Level Resolution](#task-level-resolution)
@@ -169,7 +173,7 @@ task hello {
   }
 
   command {
-    egrep '${pattern}' '${in}'
+    egrep '~{pattern}' '~{in}'
   }
 
   runtime {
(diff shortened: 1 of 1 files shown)


cjllanwarne:
I completely agree. This is also already how the hermes grammar file implements things IIRC.

One comment that's implied in the examples, and (I'm pretty sure) elsewhere in the spec - but IMO could be worth calling out explicitly in the `String Interpolation` text - within the interpolation you can put any valid WDL expression (eg the name of an input is just a special case using a "variable dereference" expression). Whatever the result of evaluating that expression is, that's the value that then gets interpolated into the string.

patmagee:
thats a good point @cjllanwarne. The string interpolation accepts an expression, not necessarily an input. I think this should be the defined behavuour now that I think about it.

versions/development/SPEC.md
```diff
@@ -259,6 +259,64 @@ $float = (([0-9]+)?\.([0-9]+)|[0-9]+\.|[0-9]+)([eE][-+]?[0-9]+)?
...
+		File someFile
+	}	
+	
+	String sum = "the sum of x and y is ${x + y}"
+	String totalSize = "the size of file ${someFile} is ${size(someFile)}
+}
+
+```
+
+Any `${expression}` or `~{expression}` inside of a string literal must be replaced with the value of the expression.  If prefix were specified as `"foobar"`, then `"${prefix}.out"` would be evaluated to `"foobar.out"`.
```
pshapiro4broad: The word `prefix` here refers to a variable named `prefix`, not prefix as a concept.
```suggestion
Any `${expression}` or `~{expression}` inside of a string literal must be replaced with the value of the expression.  If `prefix` were specified as `"foobar"`, then `"${prefix}.out"` would be evaluated to `"foobar.out"`.
```

pshapiro4broad:
This seems like a good idea. Some general comments:

- I think the examples should be converted to use `~{}` instead of `${}` as `~` is always valid whereas `$` is not valid in some contexts
- The spec also describes the string interpolation operators in the `Expression Placeholders` section. If these are the same concepts, they should be defined in one place and use the same terminology. If they aren't the same, the spec should be explicit about what the differences are.
- It seems like the section `Interpolating and concatenating optional strings` should be merged into this one, or some other consolidated section, instead of being under the `Optional Parameters & Type Constraints` section.

patmagee:
@pshapiro4broad I really like all of those comments and I may or may not have taken it one step farther. I have applied your suggestions and am now treaing the placeholder within a command a just a String interpolation where instead of happening quotations, it occurs within `<<< >>>` or `command {...}`. I have also made the assumption that expression options, (Like `sep` and `default`) should actually be made available in normal interpolated strings

cjllanwarne:
> I have also made the assumption that expression options, (Like sep and default) should actually be made available in normal interpolated strings

I agree with the concept (that all interpolations are equal) but I believe there's another PR to do away with those expression options altogether so maybe we should hold off on making that change for now? (otoh I might be mis-remembering)

patmagee:
So turns out this bejaviour was already encoded in the #229 (including global string interpolation and expressions)  This makes things mich easier!

Issue #329, opened 2019-08-10
workflow/subworkflow namespacing in input JSON


mlin:
When we write workflow input JSON for Cromwell we precede the value name with the workflow name as the namespace, like `"wf.x": "foo"`. I always wondered a little why we have to put the workflow name there, but it's not a big deal.

There's a knock-on effect when a workflow calls another workflow, and the sub-workflow has an optional input we wish to supply. Then Cromwell expects *three* namespaces in the JSON key: workflow name, call name, and sub-workflow name `"wf.callname.subwf.x": "foo"`. In contrast, if the "callee" is a task rather than a workflow, we only have to write `"wf.callname.x": "foo"`. 

This syntax difference seems surprising given that no such distinction exists when we write `call callee { input: ... }` or refer to the call outputs in WDL. I can't think of any ambiguity solved by requiring the sub-workflow name there. The relevant [spec section](https://github.com/openwdl/wdl/blob/master/versions/development/SPEC.md#cromwell-style-inputs) doesn't seem to speak to this case.

Wondering if Cromwell team could comment on this, is there a specific rationale we should clarify in the spec or is it kind of just a historical path? @geoffjentry @cjllanwarne 

Based on @DavyCats https://github.com/chanzuckerberg/miniwdl/issues/193 which has a full example

geoffjentry:
Not really the comment you were looking for @mlin - I need to actually form an opinion first :) However, this seems related to a few things from the past:

- #127 
- #192 
- #262

DavyCats:
I would personally expect there to be no difference between the inputs for a sub-workflow or a task in the inputs JSON. As @mlin mentions, they are called in the same way in a WDL file, so I don't see any reason why inputs would have to be supplied differently. I'm sure there are implementation details I'm not aware of though.

patmagee:
Closing this in favor of the discussion in the PR addressing this issue #340

PR #340, opened 2019-11-08
clarify namespacing

DavyCats:
This PR aims to clarify the namespacing of calls to workflows. Currently miniwdl and cromwell handle this differently (at least in the input JSON; see also #329), so some clarification seems to be needed in the spec. 

Pinging @mlin and @cjllanwarne as engine implementers.


mlin:
Hi @DavyCats, thanks for pushing this forward! To be clear, is this saying the JSON key for a subworkflow input should *not* include the subworkflow name following the call name (as Cromwell currently expects)? If so, I like that on first principles but we probably need to talk about how breaking this is for Cromwell and existing input JSON files.

patmagee:
@DavyCats I like this as well. The namespacing has always felt a little weird. There are a considerable number of breaking changes currently in the propsosed spec, I wonder if we could do this in a non-breaking way: ie Propose that `<wf>.<task>.<subworkflow>.<input>` is actually an alias for `<wf>.<task>.<input>` where the optional `<subworkflow>` will be deprecated in a future version of the WDL specification. @mlin  @cjllanwarne as implementors would this feel better?

mlin:
Generally sounds reasonable to me. There might be some dumb corner cases to worry about, for example if the subworkflow has a call given the same name as the subworkflow itself:

* lib.wdl:
```
version 1.0

workflow subsubwf {
  input {
    Int bar
  }
}
```
* subwf.wdl:
```
version 1.0

import "lib.wdl"

workflow subwf {
  input {
    Int bar
  }

  call lib.subsubwf as subwf
}
```

Then with the aliasing, `wf.callname.subwf.bar` would be an ambiguous reference. I'm not saying this dumb case should shoot down the overall proposal, just that we'd need a rule for resolving it.

DavyCats:
@mlin I feel like that should be a name collision:
> "[...]The names of the contained namespaces, tasks, and workflow need to be unique within that namespace. For example, one cannot import two workflows while they have the same namespace identifier. Additionally, a workflow and a namespace both named `foo` cannot exist inside a common namespace. Similarly there cannot be a task `foo` in a workflow also named `foo`. [...]"
>
> last part of https://github.com/openwdl/wdl/blob/master/versions/development/SPEC.md#namespaces

patmagee:
Hey @mlin, in the example you gave, wouldnt `wf.subwf.subsubwf.bar` always evauluate to the `bar` input of `subsubwf` even when you dealias and turn it back into`wf.subwf.bar`?

DavyCats:
@patmagee I made some adjustments based on your comment. Is this what you had in mind?

patmagee:
@DavyCats that makes alot sense to me. @mlin can you verify that this works?

mlin:
My crappy example https://github.com/openwdl/wdl/pull/340#issuecomment-553092014 currently validates with both womtool and miniwdl. I'm open to simply declaring them both in spec violation :sunglasses: The section @DavyCats cited might be clarified

> Similarly there cannot be a ~~task~~ **call** foo in a workflow also named foo

Def interested in Cromwell team input here

rhpvorderman:
Good work @DavyCats .

> The names of the contained namespaces, tasks, and workflow need to be unique within that namespace.

Should we also specify inputs here? For example

```wdl
version 1.0

import foo_workflows.wdl as foo_wf
workflow DirtyNamespace {
    input {
        String foo
    }
    call foo_wf.foo as foo {
          input:
               foo=foo
     }
}
```
`DirtyNamespace.foo` as both a call and an input here. So how does that work according to the spec?

DavyCats:
After some consideration of @rhpvorderman's comment and reading the namespaces section a few times, I found that it was a bit hard to follow (it doesn't clearly define what a namespace is) and thus decided to rewrite it entirely. It also contained some notes about imports, which seemed redundant and out of place, which I  have now removed.

My intend is for the specifications to be the same as before, just spelled out a bit more clearly. Please let me know if this should be changed (or reverted).

mlin:
https://github.com/chanzuckerberg/miniwdl/pull/309

There's a branch on miniwdl which (i) prevents calls from having the same name as their containing workflow and (ii) attempts to delete the subworkflow name from the input JSON keys. Seems OK so far

cjllanwarne:
Sorry that I didn't spot this PR earlier - what you found in Cromwell looks like a bug rather than intended design since Cromwell (and WDL!) never intentionally allowed passing inputs through to subworkflow calls from input json.

Since #347 is addressing that - and in theory we'll be "adding" it for the first time, I'd be inclined to remove the "backwards-compatible-with-a-bug" option and define this as the only logical (at least to me...) syntax of `wf.call.subcall.subcall.inputname`.

DavyCats:
Thanks @cjllanwarne, I guess it was a common misconception that inputs could be passed to sub workflows through the inputs json. I have removed the "backwards-compatible-with-a-bug" option.

rhpvorderman:
Can #347 be closed if this is accepted? I think so. @DavyCats what do you think?

Also. I think we should get this into voting. I believe Cromwell 48 is waiting on a decision to support namespaced inputs in https://github.com/broadinstitute/cromwell/pull/5317. I believe @cjllanwarne coded the input JSON parsing to easily allow support for namespaced inputs on purpose. But currently it is not turned on because the current development spec does not allow it.

DavyCats:
I'm not sure, the changes in #347 should still be made I think but we could fold it into this PR. This PR doesn't (in its current state) alter the fact that the spec explicitly disallows the provision of inputs to subworkflows by the end-user.

rhpvorderman:
Well in that case we can also take the time to carefully evaluate this PR and make sure #347 is merged quickly.

patmagee:
@rhpvorderman is this ready for voting?

DavyCats:
@patmagee Yes, I believe so

aednichols:
👍

patmagee:
@mlin @cjllanwarne @jdidion  before I pass this and merge, can you please give your final thoughts on it

versions/development/SPEC.md
```diff
@@ -185,7 +185,7 @@ workflow wf {
 
 This describes a task, called 'hello', which has two parameters (`String pattern` and `File in`).  A `task` definition is a way of **encapsulating a UNIX command and environment and presenting them as functions**.  Tasks have both inputs and outputs.  Inputs are declared as declarations at the top of the `task` definition, while outputs are defined in the `output` section.
 
-The user must provide a value for these two parameters in order for this task to be runnable.  Implementations of WDL should accept their [inputs as JSON format](#specifying-workflow-inputs-in-json).  For example, the above task needs values for two parameters: `String pattern` and `File in`:
+The user must provide a value for these two parameters in order for this task to be runnable. Implementations of WDL should accept their [inputs as JSON format](#specifying-workflow-inputs-in-json). The inputs described in such a JSON file should be fully qualified according to the namespacing rules described in the [Fully Qualified Names & Namespaced Identifiers](#fully-qualified-names--namespaced-identifiers) section. For example, the above task needs values for two parameters: `String pattern` and `File in`:
```
jdidion, line 188: s/should/must/

versions/development/SPEC.md
```diff
@@ -2260,47 +2283,19 @@ aliased name.
...
-```wdl
-import "tasks.wdl"
-
-workflow wf {
-  call tasks.x
-  call tasks.y
-}
-```
+The following namespaces exist:
+* A WDL file: When imported the name equals that of the basename of the file by default, but may be aliased using the `as identifier` syntax.
```
jdidion, line 2287: add link to relevant place in the spec

versions/development/SPEC.md
```diff
@@ -2260,47 +2283,19 @@ aliased name.
...
+  * May contain namespaces, tasks, structs (please see the notes at [Importing Structs](#importing-structs)) and at most one workflow.
+* A call (of a task or workflow): The name equals that of the called task or workflow by default, but may be aliased using the `as identifier` syntax.
+  * May contain inputs, outputs, runtime_attributes (if the call is to a task), variables (accessibility limited by [scope](#scope)) and calls (if the call is to a workflow).
```
jdidion, line 2290: add link to relevant place in the spec

jdidion (approved):

jdidion:
👍 with minor comments

mlin:
LGTM, my only comment re @jdidion on the JSON is that the first namespace -- the name of the top-level workflow being run -- doesn't constrain or disambiguate anything, so I think it's nice to leave a little wiggle room to omit it from the JSON keys.

DavyCats:
I addressed your comments @jdidion.

jdidion:

Issue #338, opened 2019-10-22
Update RFC to Include Test cases


patmagee:
Before making a PR to the RFC process, I wanted to start a conversation here to see whate you guys think.

Whenever an new change is submitted, we currently have now way of testing whether the feature has been implemented or no, OR whether the implemented feature meets the requirements of the WDL specification. This approach makes it a bit difficult for engine implementors to know whether or not they are adding the requested features as intended.

It would be good for each PR to include a set of positive and negative test cases which could help guide engine implementors when developing new features. These test cases should be simple WDL's that are made to specifically test a single feature against the current release(? or maybe development?) specification, assuming that the new feature has been added to the spec. 

The question we should focus on are:

1. Is this necessary?
2. How will this benefit the Engine implementors?
3. Where do we put test cases? 
4. How many test cases are sufficient to accept a single PR?
5. How do we create test cases that will actually work when the development specification is rapidly changing?

illusional:
Yeah +1 to this, I've been mostly working out of the development spec and there have definitely been a few times where I've been unsure whether the feature has been implemented or not quite working too. I guess it's not super related to this PR, but I have a few thoughts:

- there should be an element of "Engine implements Feature X". I do like CWL's approach to this with the Requirements (even though sometimes they feel clunky), it's very clear whether an engine supports the "feature" you're using. Also gives you a basic metric like: "this engine has a 100% conformance to the WDL test cases".

- The contributor providing _simple_ test cases should be a part of the issue, or one of the first things that occurs. More test cases should mean less accidental regression of functionality, or if it's a legitimate issue then it can become part of the test suite. To address (5), it will make it explicit how the functionality is changing.

I'm not an engine implementor, but as a pipeline author (and tester of different engines), it's hard to know whether an engine properly supports a spec without running your workflows through them, so some things are important to me.

rhpvorderman:
>     1. Is this necessary?
> 
>     2. How will this benefit the Engine implementors?
> 
>     3. Where do we put test cases?
> 
>     4. How many test cases are sufficient to accept a single PR?
> 
>     5. How do we create test cases that will actually work when the development specification is rapidly changing?

1. Yes. I think this is a great idea. This will be beneficial in multiple ways:
    + By writing actual WDL it will become more clear if a feature is needed, it will allow the change requester to think deeper about the request.
[...]

PR #304, opened 2019-04-03
Add min, max to list of standard library methods

pshapiro4broad:
A spec addition for issue https://github.com/openwdl/wdl/issues/276

Also fix markdown typos in select_all, select_first

PR #250, opened 2018-09-12
Revert "Clarify type conversions and meanings"

geoffjentry:
Reverts openwdl/wdl#235

Placeholder PR as I accidentally prematurely merged 235. If that work can't be implemented, we can merge this PR

PR #342, opened 2019-11-20
Grammar Remake

patmagee:
# Motivation
For a long time I have been bothered by the fact that the current Grammar and Parser language is more or less a dead project. While it has served us well, the knowledge of how to use it has largely since left the openWDL community. This generally means that modification of the Grammars is not an active part of the PR or development process, nor is there an easy way to test changes added the specification which is user accessible.

While writing grammars is not something that absolutely everyone knows how to do, I wanted to choose a langauge that was accessible for even a beginner which fit our needs, and had a large amount of supporting documenation. To that end I chose to write the grammars as `EBNF's` using [ANTLR4](antlr.org).

By changing the grammar I hope to encourage more community members to edit it as part of the PR process (potentially opening the way of requiring changes as part of the PR), as well as help facilitate testing of new features as being discussed #338 

# Why ANTLR

There are a large number of languages for writing grammars out there with varying different syntaxes. I originally started by trying to implement the spec strictly as a `BNF`. This quickly ran into issues as 1) there are not alot of tools for  parser generation fo strict BNF's 2) WDL is already to complext to properly implement without special rules for command and string interpolation.

Antlr4 provided a syntax which was familiar (EBNF), while still providing the ability to create parsers in a 8 different languages while allowing the grammar to be (mostly) agnostic of the target language.

# The Changes

1)  I have ONLY implmenented the CURRENT development specificaiton, including all PR's whose voting has previoulsy passed. This means: #315 #304 #297 #250 #243 #229 #222
2) My intention is to fully deprecate hermes in a future version of wdl (not 2.0), but for now I have moved the parsers and code under a directory named `hermes`
3) I have implemented a parser and two different lexers in antlr. The duplication simply has to do with the way that antlr calls actions within the lexer. If you diff both lexers you will find the only difference is `this` and `self` references
4) I have added rudimentary implemenations of the Python3, Python2, Java, and Javascript parsers, with some preliminary tests. The parsers do not implement any custom `Visitors` or `Listeners` and I do not try to modify any of the tokens/tree that is returned. There are interfaces and base implementations  for eachlanguage of `visitors` and `listeners` that anyone can extend to fit their needs
5) I have added documentation on building and setup for each language. The actual code generated by antlr is NOT included in this repo. Instead the base lexer needed for each language is added. Simply follow the build instructions to create the code

A big shout out to everyone whoe formally edited the hermes files, and for @mlin with their miniwdl implementation in Lark that helped with some decisions


orodeh:
This is great! It would be some much better than the hermes grammar that we have now.

versions/development/parsers/antlr4/Python3/README.md
```diff
@@ -0,0 +1,28 @@
+# WDL Python2 Parser
```
rhpvorderman: This is the python3 parser. Something is wrong in the generation script.

versions/development/parsers/antlr4/Python3/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+This directory provides a parser implemented in Python3. The parser is built on top of the Base [WDLBaseLexer.py](WdlParser/WDLBaseLexer.py)
+
+# Requirements
+- python3
+
+# Installing the runtime
+
+```bash
+pip install antlr4-python2-runtime
```
rhpvorderman: Again python2 instead of python3

As a more general comment: Python2 is going to be deprecated within two months.
It might be better to remove the python2 parser all together.

versions/development/parsers/antlr4/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+It is important that the lexers only differ in terms of their actions, and the actual rules stay the same to guarantee that all parsers are equivalent. 
+
+## Parsers
+
+The grammars themselves cannot be used to parse a WDL as is, they first must be converted into a parser in one of the supported target languages. Currently
+Java, Python3, Python2, and JavaScript are supported. For Python and JavaScript you can use the `antlr4` command line tool to generate the required code that is used in the Parsers. For the java target, the easiest way to generate the code is to use the `mvn` plugin defined directly within the `pom.xml`.
```
rhpvorderman: Let's not merge a PR that combines "python2" and "supported" in one sentence. Python2 will be deprecated in a few days, but it still has a user base. It's much easier to not ship python2 support now than to get rid of it later.

patmagee:
Wow @rhpvorderman I completely forgot that  the deprecation was happening. I will remove python2 support. I have no desire of maintaining deprecated code

patmagee:
@rhpvorderman I have removed the python 2 code

geoffjentry:
Not an official review but this is amazing. I suspect it'll take a long time (perhaps forever) before Cromwell is off of Hermes but this is a huge benefit for WDL in general

cjllanwarne:
I agree, very awesome indeed!

@patmagee a few minor comments -

1. I don't think "Cromwell adoption" should _necessarily_ block adoption of this grammar as the openWDL standard. In the same way that miniWDL rolls its own grammar outside of the openWDL repo, so could Cromwell until it catches up.
    - That said, having at least one working engine in this new grammar sooner rather than later would be a big win IMO.
    - The way WDL->WDLOM->WOM works in Cromwell, I don't think it would be an enormous project to use this as an alternative parser (assuming I can still traverse over Java-object parse trees that look reasonably similar to what Hermes generated) - for WDL 1.0 at least. it's not a completely trivial project for us though so we'd probably need to be able to justify it (say because the grammar has moved on from what the Hermes versions can do, for example)
2. Do you imagine that submitter-proposed changes in this grammar become a new bar to allowing PRs to be one or more of submitted/reviewed/merged?
3. Is anything in this PR generated code? It *seemed* like it from a quick scan, and in an ideal world I'd rather leave generated code out of the repo **if possible** (but, for example, I know that was _not_ possible for hermes!)
4. I have a whole bunch of test cases for the more (let's say) annoying parts of the grammar to represent in the hermes parser - would you be open to my submission of some test cases for the new grammar prior to final adoption?
5. How deep do you allow "interpolation-within-string-within-interpolation-within-string-within....." cycles to go?
    - eg `String s = "~{"~{"~{"~{"hello!"}"}"}"}"`
6. Do you support WDL 1.0 or `development`? Or both?
    - I'm assuming that draft-2 is not supported, but let me know if I'm wrong!

patmagee:
@cjllanwarne Thanks for the comments!

1. I think its a good idea to keep both hermes and this grammar for at least a few version before making any decisions to deprecate. After testing we may indeed find out that this is actually easier to use! I know from the past that using the parsers with a visitor is actually pretty easy for complex languages
2. It would be my hope that in teh future the RFC process would eventually require grammar changes before a PR is merged. But, I think keeping with test cases for now at least would be a good stop over
3. Nope, there is 0 generated code checked in!! For java you use the maven wrapper to build it, for python, make, and for javascript npm!
4. Bring on the tests!!!!
5. Theoreticaly it could be infinite. But at the moment I only allow one layer deep using the same quotations. Upon entering a quoted section a flag is set that will prevent entering anotehr quote of the same type
6. This is just for development right now, but it would not be much work to retrofit to 1.0. Not sure if we should though

patmagee:
In think this should be backported to v1.0 so we can adequately test this against existing workflows before we can think of merging. I will get working on making a v1.0

patmagee:
@orodeh I have added support for 1.0

patmagee:
@cjllanwarne I have made a number of fixes and Improvements to the grammar based on the feedback from @orodeh as he works through creating a scala library from it. Additionally I have been working on making java bindings from the grammar as well.

At the moment I think the grammar is in a good state and seems to be able to handle even a complex wdl (from testing locally). It would be great of you could review this again so we could potentially get it merged soon!

orodeh:
@patmagee I updated to the latest grammar, and it works nicely with my scala code. This is great!

I ran into an issue with `//` at the beginning of the line. In languages like C/C++/scala/Java it denotes the beginning of a comment line. In this grammar, it looks like the rest of the document gets dropped. Is this intentional? 

For example, the following task is parsed correctly:
```
version 1.0

# These are various expressions
task district {
  Int i = 3
}
```

However, if you make a mistake and use `//` instead of `#`, like this:
```
version 1.0

// These are various expressions
task district {
  Int i = 3
}
```

The `district` task is just dropped from the resulting document.

orodeh:
In case anyone is interested, the code is at https://github.com/dnanexus-rnd/wdlTools/tree/develop

patmagee:
Thanks for reviewing this @orodeh. That issue with the comment is strange. I'll test it locally to see if I can replicate it! I would not expect it to happen

orodeh:
One more thing, name access in an object results in the wrong derivation (I think). 

```
version 1.0

# An expression that access a field in an object
task district {
  Int k = x.a
}
```

`x.a` is parsed as an identifier, instead of the `#at` rule.

patmagee:
@orodeh thanks for the sleuthing!

I pushed two fixes for the issue.
1) The identifier was actually defined as `CompleteIdentifier (DOT CompleteIdentifier)*` which was causing that string to be matched as a Token instead of being cauaght by the `dot` accessor rule
2) I did not explicity define EOF for the `document` rule. Apparantly ANTLR's default behaviour is to just throw out all tokens that come after an illegal character within a rule, unless `EOF` is explicitly added

orodeh:
I pulled the updates, and both problems are fixed now. 

Thanks!

orodeh:
I have worked with the grammar, and works very well. 
👍

geoffjentry:
Couldn't be more supportive of this. Thanks so much 👍

rhpvorderman:
Thanks @patmagee for tackling this substantial amount of work, and thanks @orodeh for putting it into practice immediately, so it could be thoroughly evaluated.

:+1:

aednichols:
This is amazing 👍

PR #343, opened 2019-11-21
permit one-letter identifiers

mlin:
:sunglasses: 

I assume this is some vestigial typo as the spec has many such examples

### DIFF
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -234,7 +234,7 @@ These are common among many of the following sections
 
 ```txt
 $ws = (0x20 | 0x09 | 0x0D | 0x0A)+
-$identifier = [a-zA-Z][a-zA-Z0-9_]+
+$identifier = [a-zA-Z][a-zA-Z0-9_]*
 $boolean = 'true' | 'false'
 $integer = [1-9][0-9]*|0[xX][0-9a-fA-F]+|0[0-7]*
 $float = (([0-9]+)?\.([0-9]+)|[0-9]+\.|[0-9]+)([eE][-+]?[0-9]+)?


rhpvorderman:
I assume this was intentional.

One-letter identifiers were fine when I was programming BASIC on a Casio calculator that only had room for 26 variables anyway. But now that we actually have grown-up languages and grown-up computers we can ~~encourage~~ enforce meaningful variable names. If people can not use one-letter identifiers, but have to use meaningful variable names it will make sharing pipelines much easier.

Since this is WDL and not a programming language, I can not think of any example where usage of a one-letter identifier would be advantageous.

```WDL
task t {
    input {
        String i
    }
    command {
        t -a ~{i} > o
    }
    output {
         File o = "o"
    }
}

workflow w {
    input {
         String s
         File f
         Int i
    }
    call t {
       input:
           i = s
    }
    output {
         File c = t.o
    }
}

**Comment by patmagee** [comment]
@rhpvorderman @mlin technically according to the grammar this is alreadyy allowed, and therefore this is not a spec change, but in fact a typo fix

```
r'{%_identifier_start%}({%_identifier_follow%})*' -> _identifier
```

**Comment by geoffjentry** [comment]
FWIW I can't recall if this was intentional or accidental. I suspect the latter.

**Comment by patmagee** [comment]
@geoffjentry is ist your take that the spec currently allows 1 character identifiers?

**Comment by geoffjentry** [comment]
@patmagee technically it **is** in the spec. I've never been a fan of the Hermes grammar blocks in the spec itself, but the `identifier` description is in the spec. So regardless of typo or not, it's defined that way in the spec document

**Comment by cjllanwarne** [comment]
My gut reaction is that the language should be permissive in this regard, even if one letter names are often (but not always!) a bad idea in practice.

Why shouldn't I be allowed to say `scatter (i in indices) { ... }`, for example?

Since you were already changing this... I'd also vote for removing the "Hermes-in-markdown" altogether and replacing it with words and/or WDL examples to the same effect.

**Comment by rhpvorderman** [comment]
> Why shouldn't I be allowed to say scatter (i in indices) { ... }, for example?

Because `i in indices` will move off screen and then there are all these task calls that use `i` as input and you have to scroll all the way up to figure out what it means. This is fine for one instance, but if you have to do it for multiple instances this makes the code very hard to read. See the example I posted above. It is bad code style. `scatter (index in indices) { ... }` is much better. It will cost you 4 characters (equals roughly 2 seconds) now, but will save you much more than that when you look at your pipeline later.

This is of course just my opinion. There is a case to be made for not enforcing good code style and allowing more developer freedom. But why not enforce this? Is two seconds that bad for the huge readability gains we get? I mean even `idx` is much, much better. Then it is just one second! 

In WDL I can absolutely not see any case where having one letter identifiers is a good idea. We might as well forbid it then, and make pipelines easier to read, and therefore easier to share.

But there is of course also the implementation side. I don't want the cromwell and miniwdl developers to worry about code style issues if it is a huge hassle.

**Comment by DavyCats** [comment]
One could argue that certain parameter are commonly referred to by one character and it would, therefore, be readable and maybe even expected that a given input or variable has a one character name.

eg.:
```wdl
task kMeans {
    input {
        Int k
[...]

# or

task Pythagoras {
    input {
        Float A
        Float B
[...]
```

rhpvorderman:
@DavyCats . I forgot about that use case. Thanks!

Well in that case I am all in favor of formalizing this in the spec.

illusional:
Might not be best practice, but as a programmer when I see `i`, I often presume it’s a loop / temp variable.

mlin:
Hi all, just to reiterate, the spec document has *numerous* examples with one-letter identifiers, so I really assumed the formal regexp was erroneous in forbidding them. I don't disagree with the code style considerations, but forbidding one-letter identifiers while allowing two-letter ones hardly seems to me the way to go about that...

geoffjentry:
Just to be clear, I think that one letter IDs are fine. I was just stating that from my reading of things it'd be a spec change and thus need voting, as much as I hate the reason as to why that is the case.

patmagee:
It seems like we have a general consensus here that "even though it shouldnt be done, there are some natural use cases for single char variables". @mlin if you are okay with It, i can move this PR to voting

mlin:
:+1:

DavyCats:
:+1:

rhpvorderman:
:+1:

patmagee:
:+1:

cjllanwarne:
👍 on the change itself - **but**... I think we'd be better off removing those inline grammar sections altogether when we spot errors, rather than appearing to endorse them in general by trying to fix them ad hoc / one piece at a time.

geoffjentry:
Agree with @cjllanwarne in that I'd never object to someone opening a PR removing all of that crap and/or removing it every time they modify something related. But that's not what we're voting on here :)

+1

orodeh:
👍

aednichols:
👍

cjllanwarne:
It looks like this PR has passed easily, and since it's how at least one engine implements things, I think this can merge immediately

PR #349, opened 2019-12-16
remove file:// protocol

rhpvorderman:
I was working on [wdl-packager](https://github.com/biowdl/wdl-packager/) and stumbled across the implementation of the `file://` protocol. Apparently [miniwdl](https://github.com/chanzuckerberg/miniwdl) does not support relative imports with this protocol.

I made a [PR](https://github.com/chanzuckerberg/miniwdl/pull/310). @mlin did some research and find out that relative imports should not be allowed at all with this protocol.

This makes it useless for use in portable WDLs. Also [the number of slashes  is confusing](https://en.wikipedia.org/wiki/File_URI_scheme#How_many_slashes?). I have never used it in a workflow. I have never seen it used in a workflow. Since we are heading towards 2.0 now is a good moment to remove this protocol.

### DIFF
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -809,7 +809,6 @@ workflow wf {
 Engines should at the very least support the following protocols for import URIs:
 
 * `http://` and `https://`
-* `file://`
 * No protocol (see below)
 
 In the event that there is no protocol the import is resolved **relative** to the location of the current document. If a protocol-less import starts with `/` it will be interpreted as starting from the root of the host in the resolved URL. It is up to the implementation to provide a mechanism which allows these imports to be resolved correctly.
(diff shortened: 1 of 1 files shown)


mlin:
Yes, I read the [file:// URI RFC](https://tools.ietf.org/html/rfc8089) saying they're only to be used for absolute paths.

>The path component represents the absolute path to the file in the
>   file system.

> In a POSIX file system, the root of the file system is represented as
>   a directory with a zero-length name, usually written as "/"; the
>   presence of this root in a file URI can be taken as given by the
>  initial slash in the "path-absolute" rule.

Leading me to believe that the relative file:// imports shown in the WDL spec examples are ill-conceived (and unnecessary -- just don't put the file:// prefix). (I'm happy to be corrected, of course, if I've misunderstood something here.)

geoffjentry:
I know we (for some `we`) have had this debate before. I don't remember if it was physically in the hallowed halls of the Broad, electronically when all of this was back in the Cromwell repo, or post-OpenWDL ... but it has definitely come up before. Probably multiple times and thus all of the above.

IIRC the current state is understood to be not correct but viewed as a necessary evil to enable certain patterns which were going to happen anyways (e.g. constructs like `.` and `..` are terrible, but one will never convince ppl to lift & shift from traditional formats if they're not included)

rhpvorderman:
@geoffjentry The terribleness of the protocol-less version of imports (that allows `.` and `..`) is a different issue that we can discuss in a different place. I agree whole-heartedly that `.` and `..` should be exterminated from wdls but let's not side-track this discussion.

This is about the `file://` protocol. Which is unsuitable for WDL by design as it only allows absolute paths. This means you can never write a portable workflow with `file://`. So I don't see why we should burden the implementers from Cromwell and miniwdl with this. They have plenty of stuff to work on already.

patmagee:
@rhpvorderman is this ready for voting?

rhpvorderman:
Yes. Thanks.

aednichols:
👍

cjllanwarne:
Seems fine to me. This is removing paths starting with `file://` rather than the ability to import from files on the local filesystem

patmagee:
Vote passed 2-0

Issue #344, opened 2019-11-22
Publish Changelogs with each new Version


patmagee:
In the process of trying to categorizer what PRs need to implemented as part of the "current' version, I realized that we have no easy way of seeing the changes that were introduced into the specification.. We should publish a changelog that minimally shows the PRs that were included.

To this end, I have started creating milestones (first for 2.0) that should help us track these things more easily.

Do people have any preferences on how the changelog should look, where it should live etc?

rhpvorderman:
We usually include a `CHANGELOG.md` or `CHANGELOG.rst` in the root of the project. This is the [pytest-workflow one](https://github.com/LUMC/pytest-workflow/blob/develop/HISTORY.rst) ([raw version here](https://raw.githubusercontent.com/LUMC/pytest-workflow/develop/HISTORY.rst)).

To enforce this, including a PR template usually works. [Like this one](https://github.com/LUMC/pytest-workflow/blob/develop/.github/PULL_REQUEST_TEMPLATE.md). This will nag for every PR that there should be something in the changelog. This is surprisingly effective. We also do this for our BioWDL pipelines (such as [this one](https://github.com/biowdl/germline-DNA)) and [BioWDL tasks](https://github.com/biowdl/tasks).

We are happy the way this works. It is easy for developers to maintain and we hope it has some added value for the users. 

I am curious what other people prefer and use.

DavyCats:
Since openWDL isn't using branches and tags to make releases, I wonder if it's the best option to put the changelog in the root. Since each version is given its own folder, perhaps these folders should include a changelog for each version?

I agree with @rhpvorderman that adding some PR template could be helpful for enforcement of the RFC process.

patmagee:
@rhpvorderman @DavyCats these are both great suggestions. I like the Idea of the PR template alot, and am just wondering where makes sense for the CHANGELOG.. I initially had the same feeling as you @DavyCats, however It does not capture changes between versions...

DavyCats:
That's a fair point. An alternative could also be to put the changelog inside of the spec itself. When a new version is made, the changelog is simply copied over (along with the rest of the spec). Each version will then list the differences between itself and the previous versions (but not future ones). One downside to this is that it will bloat an already large document.

I'm also fine with just putting one master changelog in the root of the repo or inside the /versions directory.

rhpvorderman:
[...]

PR #350, opened 2019-12-16
Add a changelog.

rhpvorderman:
This implements the discussion in #344:

+ Changelog in the root.
+ Add pull request template.

I added most of the v2.0 milestone merged changes. In some cases I have excluded grammar changes because they merely implemented a thing that was already listed as a change.

The unmerged changes are not added, and should be added upon merging. Any future PR's will get a reminder to add the changes to the log.

I have added some style guides in the comment header of the changelog as well. And also a guide on what should be mentioned per change. This is of course up for debate.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,3 @@
+
+### Checklist
+- [ ] Pull request details were added to CHANGELOG.md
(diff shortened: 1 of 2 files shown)


patmagee:
HUGE +1 Sorry I did not see this before the break!

rhpvorderman:
Yes, it was some work to go through all the merged and non-merged PR's to write the changelog. Shall we merge this so we can prevent someone from having to do this in the future :slightly_smiling_face: ?

patmagee:
@rhpvorderman I think this looks good to merge

PR #357, opened 2020-02-19
Added a required input and output format for workflow engines.

rhpvorderman:
This is a result from the OpenWDL hackathon discussion on WDL compliance testing.

To make it easier to test compliance all workflow engines should be able to process the same input format and supply the same output format. 

It was decided to use the JSON format for both in- and output formats for the following reasons:

- Is already the de facto standard for inputs and outputs across engines.
- Human-readable
- Easy to parse (compared to XML and YAML)

We decided to rename the JSON input format to something other than "Cromwell-style inputs" because we value our engine diversity :rainbow: .

### Checklist


rhpvorderman:
ping @patmagee, @jdidion

versions/development/SPEC.md
```diff
@@ -2896,9 +2896,15 @@ Note that because some call inputs are left unsatisfied, this workflow could not
...
 
-### Cromwell-style Inputs
+### Common WDL JSON Input Format
 
-The "Cromwell-style" input format is widely supported by WDL implementations and recommended for portability purposes. In the Cromwell-style format, workflow inputs are specified as key/value pairs in JSON or YAML. The mapping to WDL values is codified in the [serialization of task inputs](#serialization-of-task-inputs) section.
+The Common WDL JSON input format is an input format that is required to be 
+accepted by any workflow engine. It is the input format that is used for 
+performing language compliance tests on the engine.
+
+In this common format, workflow inputs are
```
jdidion: This should be usable for both workflow and task inputs/outputs.

rhpvorderman: This is in the workflow inputs chapter. Can you clarify what you mean by task inputs in this context?

rhpvorderman:
@mlin, @cjllanwarne , @orodeh as engine implementers, what are your opinions on this? Are these workable input and output formats?

versions/development/SPEC.md
jdidion: Some workflow engines (Cromwell, dxWDL) support running individual tasks in addition to workflows. This input specification should be the same for both workflows and inputs, and it should be required for any workflow engine that allows running tasks.

orodeh: I think this is good. My only comment is to consider shortening the concept name from `Common WDL JSON Input Format` to `JSON Input Format`. It makes it more readable, and reduces potential confusion with the "Common Workflow Language". After all, this document **is** the authoritative document on the WDL language, so the words `common` and `WDL` should be superfluous. Any changes an engine makes are not standard/common by definition.

rhpvorderman: @orodeh Very good suggestion. It was also a bit tedious to type. 

@jdidion, running individual tasks is a feature that is not listed in the spec. So workflow engines can do as they please with regards to running tasks. We can of course make running tasks part of the spec, but that will be another PR entirely.

rhpvorderman:
I removed the Common WDL stuff. Is this okay for everyone to get into voting?

orodeh:
I think so.

orodeh:
👍

patmagee:
:+1:

pieterlukasse:
👍

aednichols:
👍

Issue #351, opened 2020-01-04
conditionals in meta block


ying-w:
In the if statement example, there is a [conditional in runtime block](https://github.com/openwdl/wdl/blob/master/versions/1.0/SPEC.md#if-then-else)

I was wondering if conditionals should be allowed in the meta block, for example DNAnexus allows you to choose between parallel downloads (default) vs streaming using meta. I would like the capability of doing something like this:

```wdl
  task mytask {
    input {
        Array[File] bams
        Array[File] bais
    }

    parameter_meta {
        if(len(bams) == 1) {
            bams : "stream"
            bais : "stream"
        }
    }
}
```

I think there was some conversation about this earlier in #73

DavyCats:
Shouldn't this be handled by the runtime block? As far as I understand, `parameter_meta` and `meta` are static key-value pairs intended for documentation/descriptions of the inputs/outputs.

patmagee:
@ying-w Agreeing here with @DavyCats the two meta sections are really meant for documenatation purposes only, althought at the moment they can support nested JSON. In the next version of WDL (2.0), there is a new section called `hints` wich supports expressions which is meant to provide the execution engine with a means of defining additional behaviour to help optimize the run. DNAnexus doing it this way in the parameter_block, is slightly out of spec and really meant to fit their internal use cases.  

I for one would prefer to keep this type of expression out of the parameter meta, since it reinforces behaviour that is not supposed to work nor is it in the spec

ying-w:
Thanks for feedback, I'll pass it along to DNAnexus

PR #359, opened 2020-03-06
Write a specification for unsatisfied task inputs.

rhpvorderman:
This PR is a result of the discussion in #347 . This alternate PR was made because the scope shifted during the discussion.

These were some of the points made in the discussion:
1. Users should be able to know what inputs they should provide by looking at the input section of the top-level workflow. This should not be hindered by implicit optional inputs bubbling up.
2. For clinically validated workflows only a select number of inputs may ever be changed.
3. Allowing engines to pursue different models for computing inputs will lead to compatibility problems.
4. Requiring all task optional inputs to be filled in the top-level workflow input section is infeasible and an enormous hassle.

This PR attempts to address all these points by:
* Requiring all **required** task/workflow inputs to be filled by the calling workflow. This way the top-level workflow inputs section will always contain all the inputs necessary for running the workflow. (Solves point 1) 
* Requiring workflows to only accept inputs for the top-level workflow by default (solves 2 and 3)
* Give engines the flexibility to allow for bubbled-up optional inputs if a configuration option is provided. This gives users the flexibility needed to solve issue 4.

Supporting bubbled-up inputs optionally is feasible in Cromwell thanks to how @cjllanwarne implemented the input calculation. I don't know about the other engines, but since it is optional, it should not be an extra burden.

Pinging @cjllanwarne , @mlin , @orodeh and @DavyCats as they brought up these points in the discussion.

EDIT: I also updated the example to reflect changes in the WDL spec so far.

### Checklist


versions/development/SPEC.md
```diff
@@ -2600,12 +2600,35 @@ In this example, `i`, and `f` are inputs to this task even though `i` is not dir
...
+    * Example: my_outer_workflow.my_workflow.my_task.my_task_input.
+* There is currently no way to supply a bubbled-up input in an outer workflow's 
+  call block.
+    * Example: one cannot say call my_workflow as subworkflow 
+      `{ inputs: my_task.my_task_input=... }`
+* By default an engine only allows inputs that are specified in the input
+  section of the top-level workflow.
+* An engine may optionally support supplying bubbled-up optional inputs, but
+  this has to be explicitly enabled on the engine (via configuration, command
+  line flags or otherwise).
```
DavyCats: I still feel it might be better to add some form of syntax in the WDL file itself. That way one would be able to see that there may be interesting nested inputs in the WDL files and, thus, that it would be good to use an engine which supports them.

In addition, a workflow might be written without the intention for the bubble-up behavior to be used. If someone then uses such a flag or configuration to allow for the bubble-up, they might be able to set optional inputs which will mess up the workflow.

An alternative to the last two points in the list:
* By default only inputs defined in the top-level workflow are made available to the user.
* The keyword `etc` must be added to a workflow's input section to enable the bubble-up behavior described above. Only if this `etc` keyword is present in the workflow's input section may an engine allow for this bubble-up behavior and an engine may choose not to support to feature at all.

orodeh: The `etc` keyword makes sense to me. However, why not just allow it in the workflow `meta` section? This way, we don't need new syntax.

rhpvorderman: @DavyCats good points! @orodeh good suggestion! Setting it in the workflow seems indeed better than setting it in the engine, and using the meta section for this seems the most logical thing to do.

Something like:
```wdl
meta {
    bubbled-up-inputs: true
}
```
Except that I don't like the bubbled-up word. It is in the PR, but more on "a lack of a better term" basis.

How about nested inputs?
```wdl
meta {
     allow-nested-inputs: true
}
```
Or some other suggestions:
```wdl
meta {
    allow-hidden-inputs: true
    allow-subworkflow-inputs: true
    allow-extended-inputs: true
```

DavyCats: I'm not sure that the meta section is the appropriate place for this. The meta section is just a set of key-value pairs that hold no relevance to the execution of a workflow. It is intended to be a place where you might store contact information or a link to the documentation. This was also discussed (briefly) in #351.

This feels more like something that would fit in the runtime section, except that there is no runtime section for workflows right now.

DavyCats: `allow-nested-inputs` seems fine to me, though I wonder if using `-`s might cause issues with the grammars.

rhpvorderman: Pinging @patmagee for the grammar question.

Well we could consider adding a runtime section for workflows, but that requires another spec/language change. (Which is why I wanted to set it engine-side) . Adding a runtime section could be useful, but there need to be other relevant things stored there. Having an entire runtime section for `allow-nested-inputs` seems a bit ludicrous to me.

I think the meta section is quite appropriate: It states that this workflow has nice nested inputs. A user does not have to use them. I think it is more documentation than runtime information, since an engine does not have to support nested-inputs per se.

patmagee: These are actually really great examples of something that can be added to the `hints` section, and we leave it up to the engine to decide what to do. We tried to really trim down the runtime  lock and make it the bare minimum requirements needed to run a workflow.

We already have the idea of "semi reserved" kw used in the hints section. These are hints that have similar meanings across all engines but do not need to be implemented.

I'd suggest doing something like:

```wdl
hints {
  allowNestedInputs: true
}
```

rhpvorderman: @patmagee, yes, that seems to be the best section but in the PR for the hints section I only see it at the `task` level. Will there also be a hints section at the the workflow level?

orodeh:
I have joined @patmagee in trying to build a grammar for WDL with ANTRL4. My part is writing scala code that builds an AST from a WDL program (https://github.com/dnanexus-rnd/wdlTools). I now have a much better appreciation for the amount of work it takes to modify and maintain the grammar. Hence my reluctance to add additional syntax, unless we really need it.

rhpvorderman:
Given the feedback I coin the following proposal:

If there is no additional use for a hints or runtime section for workflows, then we should settle on the meta block for harboring a key called `allowNestedInputs`. 

If in due time the need arises for a runtime or hints section before WDL2.0 is released, we can always change this. If the need arises after WDL2.0 is released, we can deprecate (but still support) the use of allowNestedInputs in the meta section to provide a smooth transition. 

What do you think?

orodeh:
Sounds good to me, thanks.

versions/development/SPEC.md
cjllanwarne: This doesn't feel like it quite matches the semantics of a "hint" to me to be honest.

A hint to me should mean "if an engine completely ignores this, it might miss out on some runtime optimization but the logical evaluation of the workflow should still be possible".

What we're talking about here is:
1. a warning to users that their workflow might not work on all engines. 
2. a *requirement* to supporting engines that they should allow bubbled up inputs.

I'd almost like to add a way to flag `plugin`s or `non-standard-features` or `engine-requirements`. I know Ohad has a number of other ideas for features which would be really cool but tricky for Cromwell. A way to flag these high up in the context of a WDL file (perhaps even above the workflow scope) would be really nice.

cjllanwarne: eg something like:

```
version 2.0

engine_requirements {
  bubbled-up-inputs,
  streamable-input-files,
  ???
}
```

it's not 100% ideal because it allows engines to diverge, but at least it'd be obvious right at the top of the file whether your workflow will work with any particular engine (which would presumably be able to publish or otherwise indicate which features they support). It would also allow engines to "get out quickly" if they see required features which they didn't have.

cjllanwarne:
@rhpvorderman I should have read further down before replying in thread!

Is the idea that an engine could choose either to allow bubbling up only if this flag was set, or else to fail the workflow when it sees that flag?

Your answer seems fine to me for now. The more general "how do we allow for diverging engine features" can be part of a future discussion.

rhpvorderman:
> Is the idea that an engine could choose either to allow bubbling up only if this flag was set, or else to fail the workflow when it sees that flag?

I think the engine should allow bubbling up. If it does not support bubbling up I think it should fail with a message, but only if bubbled-up namespaced inputs are actually used. Since all required inputs should be filled by the top-level workflow, this way all workflows can be run by all engines, regardless whether the flag is present or not.

versions/development/SPEC.md
```diff
@@ -2600,12 +2600,35 @@ In this example, `i`, and `f` are inputs to this task even though `i` is not dir
...
+  In this case, the inputs bubble up to become an input to the workflow instead.
+    * Example: an unsupplied input might have the fully-qualified name 
+      `my_workflow.my_task.my_task_input.`
+* If that workflow is used as a subworkflow, the input is allowed to bubble up 
+  again with a further-qualified name.
+    * Example: my_outer_workflow.my_workflow.my_task.my_task_input.
+* There is currently no way to supply a bubbled-up input in an outer workflow's 
+  call block.
+    * Example: one cannot say call my_workflow as subworkflow 
+      `{ inputs: my_task.my_task_input=... }`
```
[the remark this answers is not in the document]
rhpvorderman: Yes maybe. But we should do that in another PR I think.

rhpvorderman:
I added the flag to the PR. I did not add any stuff about engines optionally supporting this feature yet. Do we want to specify this diverging behavior now, or can this wait for later since WDL2.0 is not released yet?
As it is in the PR currently, all engines should support it, as no exceptions are made.

DavyCats:
Regarding diverging behavior, I see two viable stances that might be taken:
* The spec never declares a feature as optional, but an engine may simply choose not to implement a certain feature. The engine wouldn't be fully spec-compliant, but if that's properly documented it is probably okay.
* In cases where a feature might interfere with the possibility for a workflow to be run on certain infrastructure, such a feature may be declared in the spec as an optional feature that an engine may choose not to implement. Only features who's absence would not interfere with the runnability of a workflow may be optional, ie. even if the feature is not available in an engine the engine should still be able to run a workflow which would otherwise have made use of it, simply with less functionality/convenience/configurability.

rhpvorderman:
I propose we follow @cjllanwarne's suggestion and discuss diverging engine behavior in a future discussion.

I think the PR is ready. Are there any points that need to be addressed on this PR before it gets into voting?

DavyCats:
Should voting be opened on this? There hasn't been any further discussion in a while.

rhpvorderman:
@patmagee Could you initiate the vote on this? It might be accepted or rejected, but it's getting a bit stale at the moment. I would do it myself if I had the label powers :innocent:.

patmagee:
@rhpvorderman can you please update the changelog.md  :)? Once you do that, I can opn this up for voting

rhpvorderman:
@patmagee Done!

orodeh:
👍

DavyCats:
:+1:

versions/development/SPEC.md
```diff
@@ -2806,12 +2806,42 @@ In this example, `i`, and `f` are inputs to this task even though `i` is not dir
...
+* A workflow is allowed not to specify optional inputs in a call's input block. 
+  In this case, the inputs bubble up to become a nested input to the workflow 
+  instead.
+    * Example: an unsupplied input might have the fully-qualified name 
+      `my_workflow.my_task.my_task_input.`
+* If that workflow is used as a subworkflow, the input is allowed to bubble up 
+  again with a further-qualified name.
+    * Example: my_outer_workflow.my_workflow.my_task.my_task_input.
+* There is currently no way to supply a bubbled-up input in an outer workflow's 
+  call block.
```
pieterlukasse, line 2826: I'm confused by this point. Does it contradict the previous one? I.e. the previous one seems to introduce an option, but this point seems to state it is not possible. Can you please clarify?

pieterlukasse:
thanks @rhpvorderman . Looks good to me 👍 
I had only 1 question regarding part of the documentation. Maybe you can take a look at that?

versions/development/SPEC.md
rhpvorderman: The first point is about the input json (or whatever other input format)

The outer workflow's call block is in the WDL code itself. When you call a workflow from inside a workflow. In that area there is currently no way to supply nested inputs.

rhpvorderman:
Cromwell implementation pending in https://github.com/broadinstitute/cromwell/pull/5523.

I know this spec change does not generate a lot of buzz. However, it is important that it gets merged as a lot of the behavior it described was previously undefined. Cromwell, for example, allowed for calling tasks without supplying any inputs to the call block. These inputs could then be supplied by the inputs json of the workflow. This carries with it a lot of implicitness that makes workflows very hard to read. 
In order to get the implementation to be similar across engines a clear spec is needed.

patmagee:
This has passed with a 3-0 vote. @rhpvorderman would you be able to rebase the changelog so that this can be merged

rhpvorderman:
@patmagee done. Thanks

versions/development/SPEC.md
```diff
@@ -2783,13 +2831,16 @@ workflow wf {
...
 The inputs to `wf` would be:
 
-* `wf.t1.s` as a `String`
-* `wf.t2.s` as a `String`
+* `wf.t1s` as a `String`
+* `wf.t2s` as a `String`
 * `wf.int_val` as an `Int`
 * `wf.my_ints` as an `Array[Int]`
 * `wf.ref_file` as a `File`
+* `wf.t2.t` as a a `Int?`
```
mcovarr, line 2839:
```suggestion
* `wf.t2.t` as an `Int?`
```

DavyCats:
@rhpvorderman This was implemented in cromwell 52, wasn't it?

rhpvorderman:
@DavyCats partially.
The optional inputs are allowed now. The very strict checking on whether all required options are satisfied by the top-level workflow was not.

versions/development/SPEC.md
```diff
@@ -2705,12 +2705,42 @@ In this example, `i`, and `f` are inputs to this task even though `i` is not dir
...
+
+* A call has its inputs supplied when called by a workflow. 
+    * Example: `call my_task { input: my_task_input=... }`
+* All required call inputs which do not have defaults should be filled by the 
+  calling workflow.
+* A workflow is allowed not to specify optional inputs in a call's input block. 
+  In this case, the inputs bubble up to become a nested input to the workflow 
+  instead.
+    * Example: an unsupplied input might have the fully-qualified name 
+      `my_workflow.my_task.my_task_input.`
```
jdidion, line 2720: Period should go outside the backtick, I think

jdidion (approved):

versions/development/SPEC.md
rhpvorderman: Thanks for spotting this!

versions/development/SPEC.md
```diff
@@ -2705,12 +2705,42 @@ In this example, `i`, and `f` are inputs to this task even though `i` is not dir
...
+  again with a further-qualified name.
+    * Example: my_outer_workflow.my_workflow.my_task.my_task_input.
+* There is currently no way to supply a bubbled-up input in an outer workflow's 
+  call block.
+    * Example: the following inputs would not be valid for a call to a subworkflow
+      `{ inputs: my_task.my_task_input=... }`
+* By default an engine only allows inputs that are specified in the input
+  section of the top-level workflow.
+* If a workflow is suitable for use with nested inputs it should be explicitly
+  stated. This is done by setting `allowNestedInputs` to `true` in the meta
```
jdidion, line 2731: Can `allowNestedInputs` be in `hints` rather than `meta`?

jdidion (changes requested):

versions/development/SPEC.md
jdidion, line 2731: My problem with this is that things like `allowNestedInputs` don't really belong in `meta` - there's just no where else to put them. I am going to open a proposal to add a `hints` section to workflow with the same semantics as it has in task. I think things like this more appropriately belong in hints than in meta.

patmagee: I think  when designing the `hints` section, I ultimately planned on having one in the `workflow` as well. however it felt like a more natural transition to start with it in the task. In the lont term this is probably a good move. If we add the hints in 1.1, I feel like 2.0 can propogate the hints to the workflow

rhpvorderman: I agree! Meta was always a bit of a "let's put it where it doesn't hurt" manoeuver. 

Not having this feature proved to be a major hindrance in getting actual work done. But at least it is implemented in cromwell now, so it is at least possible to get work done until this feature lands.

If the hints proposal is landed I will rewrite this PR.

jdidion:

PR #360, opened 2020-04-02
Add document with differences between WDL versions

jdidion:
Beginning a document to enumerate the differences between WDL versions. This will be useful to users to help them know what they need to update to move their WDLs from one version to another, and it will be useful to those implementing parsers or workflow engines to know what new features they need to support and old features they need to deprecate.

I started by documenting the difference between Draft-2 and 1.0 as it seems like these are the two versions most often found "in the wild." Eventually we should enumerate all the differences between each successive pair of versions.

I would greatly appreciate help in answering the "Open Questions" as well as additions/changes anywhere I've missed something/gotten something wrong.

### DIFF
--- a/versions/Differences.md
+++ b/versions/Differences.md
@@ -1,0 +1,47 @@
+# Differences Between WDL Versions 
+
+⚙️denotes a change that affects the parser
+
+## Draft-2 to 1.0
+
+*   ⚙️Version statement required (as of draft-3)
+*   ⚙️Structs
+*   ⚙️For both tasks and workflows, unbound inputs must be within the `input {}` block
+    *   Declarations outside input block cannot be overridden
+    *   As a consequence of this, the way workflow inputs are calculated differs between the two specs - refer to the “Computing Inputs” sections for details
+*   The workflow `output` section is optional in both specs, but the 1.0 spec clarifies that the output section is required in a workflow called as a sub-workflow.
+*   Task command section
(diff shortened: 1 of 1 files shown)


patmagee:
hey @jdidion thanks alot for this! I was wondering if these changes are different then waht can be groked from the [CHANGELOG.md](https://github.com/openwdl/wdl/blob/master/CHANGELOG.md)

jdidion:
Good point - I could just merge this document into CHANGELOG

illusional:
Maybe it's not worth it because it's not finalised yet, but I'm very interested in 1.0 and development. 

Some initial I got from the diff:

- New Directory type
- Standard library:
  - Removal of:
    - `*_object*` methods (eg: read_object
  - New methods:
    - `as_pairs`, `as_map`, `keys`, 

I guess a Removal of `Object` maybe? Or certainly a very heady reduce in it's usage.

Lots of clarification to specs.

patmagee:
@illusional There are pretty significant changes in 2.0 vs 1.0. I
d suggest checking out the changelog and the open PR's in this milestone: https://github.com/openwdl/wdl/milestone/1 for more information. But Amongs the biggest changest are the removal of the `object` type, the rework of the `runtime`, and the addition of `hints`

jdidion:
Thanks @patmagee. FWIW we also support automatic draft-2 to 1.0 migration with wdlTools (using the 'upgrade' command).

orodeh:
Thanks @jdidion

PR #361, opened 2020-04-02
Add ANTLR4 grammar for draft-2

jdidion:
I also normalized the spacing in the file (1 tab for indent, max 1 blank line between statements)


patmagee:
@jdidion I made a fix in #363, would you be able to backport the fix to this PR?

jdidion:
will do...I also have a couple other changes I'll add

versions/1.0/parsers/antlr4/WdlLexer.g4
```diff
@@ -1,6 +1,9 @@
-lexer grammar WdlLexer;
+lexer grammar V10WdlLexer;
```
patmagee, line 1: I would prefer the syntax of `Wdl<Version>Lexer` and `Wdl<Version>Parser`

versions/1.0/parsers/antlr4/WdlLexer.g4
```diff
@@ -53,6 +56,7 @@ BoolLiteral
 	: 'true'
 	| 'false'
 	;
+Null: 'null';
```
patmagee, line 59: `Null` did not make its way into the grammar until the upcoming development version. IN v1.0 and draft-2 it is not valid

versions/1.0/parsers/antlr4/WdlParser.g4
```diff
@@ -47,34 +47,35 @@ number
...
+	: DQUOTE string_part string_expr_with_string_part* DQUOTE
+	| SQUOTE string_part string_expr_with_string_part* SQUOTE
+	;
 
 primitive_literal
 	: BoolLiteral
 	| number
 	| string
 	| Identifier
+	| Null
```
patmagee, line 78: Null is not a valid symboll in V1

versions/1.0/parsers/antlr4/WdlParser.g4
```diff
@@ -157,21 +158,25 @@ struct
 	: STRUCT Identifier LBRACE (unbound_decls)* RBRACE
 	;
 
-meta_kv
-	: Identifier COLON expr
+meta_expr
+	: expr
+	| Null
```
patmagee, line 163: Null is not a valid symboll in V1. Additionally, in a future version of the parser its `None` and its placed within the expression resolution, howver you would not explicitly have to add it there, since it it added to the `primitive_literal` rule

versions/draft-2/parsers/antlr4/WdlLexer.g4
```diff
@@ -0,0 +1,194 @@
+lexer grammar WdlLexer;
```
patmagee, line 1: For consistency with the above changes you should change this to `WdlDraft2Lexer`

versions/draft-2/parsers/antlr4/WdlParser.g4
```diff
@@ -0,0 +1,272 @@
+parser grammar WdlParser;
```
patmagee, line 1: For Consistency this should be changed to `WdlDraft2Parser`

versions/draft-2/parsers/antlr4/WdlParser.g4
```diff
@@ -0,0 +1,272 @@
...
+	| NOT expr #negate
+	| (PLUS | MINUS) expr #unirarysigned
+	| expr_core LBRACK expr RBRACK #at
+	| IF expr THEN expr ELSE expr #ifthenelse
+	| Identifier LPAREN (expr (COMMA expr)*)? RPAREN #apply
+	| Identifier #left_name
+	| expr_core DOT Identifier #get_name
+	;
+
+import_alias
```
patmagee, line 137: Import alias was added in v1.0 and is not present in draft-2

versions/draft-2/parsers/antlr4/WdlParser.g4
```diff
@@ -0,0 +1,272 @@
...
+
+workflow_element
+	: workflow_output #output
+	| inner_workflow_element #inner_element
+	| parameter_meta #parameter_meta_element
+	| meta #meta_element
+	;
+
+workflow
+	: WORKFLOW Identifier LBRACE workflow_input workflow_element* RBRACE
```
patmagee, line 262: Similar situation to the task above. Workflow inputs are not bound to be defined at the start of the workflow, but can be interspersed throughout as either a declaration or a bound declartion.  moving the `workflow_input` to the workflow_element and getting rid of the `bound_decls` from the `workflow_element` rule will fix this

versions/1.0/parsers/antlr4/WdlLexer.g4
jdidion: It appears in the 1.0 spec https://github.com/openwdl/wdl/blob/master/versions/1.0/SPEC.md#parameter-metadata-section

versions/draft-2/parsers/antlr4/WdlParser.g4
jdidion: "Tasks define all their inputs as declarations at the top of the task definition." https://github.com/openwdl/wdl/blob/master/versions/draft-2/SPEC.md#task-inputs

IMO this should be interpreted literally - any declaration before the first non-declaration is considered an input, and others must be bound and are added to bound_decls.

versions/1.0/parsers/antlr4/WdlParser.g4
jdidion: I added it to primitive_literal by mistake. But according to the 1.0 spec it is a valid value within the meta sections.

versions/1.0/parsers/antlr4/WdlParser.g4
jdidion: This was added here by mistake.

versions/draft-2/parsers/antlr4/WdlParser.g4
patmagee: hmm good point. I guess this is one of the ambiguities that later versions hoped to resolve with the `input {...}` block.

versions/1.0/parsers/antlr4/WdlLexer.g4
patmagee: ugh, this is a sneaky ugly addition. Adding a `null` keyword that can only be used in a given parameter meta block. 

`null` is not a valid special  value anywhere else except here. I would be in favor of not treating it special at the level of the parser/lexer, but allow the implementations to worry about this. The reason being is that the following is valid under v1.0 since `null` is not technically a keyword, except in the case of the meta
```wdl
workflow a {
 input {
   String null
 }
}
```
Inn this case `null` SHOULD evaluate to whatever the user set it as. However, if we add a lexer rule for this, `null` will not be considererd an `identifer` and this workflow would fail to parse.

I say we treat `null` as a special case wihtin the downstream libraries. IN the general case its considered an identifier, however within the `parameter-meta` section, if the `null` value is not set, then the library should evaluate this to a literal `null`

jdidion: Gotcha, that makes sense. I'll revert.

versions/1.0/parsers/antlr4/WdlLexer.g4
jdidion: Do you like WdlV10Lexer or Wdl10Lexer? I would prefer WdlV1_0Lexer or Wdl1_0Lexer but it doesn't seem like Antlr likes underscores in these names.

versions/1.0/parsers/antlr4/WdlLexer.g4
jdidion: But then shouldn't meta_kv be `Identifier COLON expr?`?

jdidion: Oh nevermind, I see what you're saying - treat null as an identifier in the parser, but the engine should special-case the null identifier - if it doesn't match any declaration then treat it as null literal.

versions/1.0/parsers/antlr4/WdlLexer.g4
patmagee: Could we do WdlV1Lexer for this case?

versions/draft-2/parsers/antlr4/WdlDraft2Lexer.g4
```diff
@@ -0,0 +1,194 @@
...
+
+mode HereDocCommand;
+
+HereDocUnicodeEscape: '\\u' (HexDigit (HexDigit (HexDigit HexDigit?)?)?)?;
+HereDocEscapedChar: '\\' . -> type(CommandStringPart);
+HereDocCurlyString: '{' -> type(CommandStringPart);
+HereDocEscapedEnd: '\\>>>' -> type(CommandStringPart);
+EndHereDocCommand: '>>>' -> popMode, type(EndCommand);
+HereDocEscape: ( '>' | '>>' | '>>>>' '>'*) -> type(CommandStringPart);
+HereDocStringPart: ~[{>]+ -> type(CommandStringPart);
```
patmagee, line 127: I believe you are missing a token for starting a command interpolation, as well as the `$` token. It should probably have the following as well
 ```antlr
HereDoctDollarString: '$' -> type(CommandStringPart);
HereDocStringCommandStart:  ('${') -> pushMode(DEFAULT_MODE), type(StringCommandStart);
```

Comment (patmagee) [reply versions/draft-2/parsers/antlr4/WdlDraft2Lexer.g4:127]: Also, the `HereDocStringPart` is missing the `$` as well

File: versions/draft-2/parsers/antlr4/WdlDraft2Parser.g4 [hunk 10]
```diff
@@ -0,0 +1,268 @@
+parser grammar WdlDraft2Parser;
+
+options { tokenVocab=WdlLexer; }
```
patmagee, line 3: This should be `WdlDraft2Lexer`

versions/draft-2/parsers/antlr4/WdlDraft2Parser.g4
```diff
@@ -0,0 +1,268 @@
...
+	| expr_infix4 DIVIDE expr_infix5 #divide
+	| expr_infix4 MOD expr_infix5 #mod
+	| expr_infix5 #infix5
+	;
+
+expr_infix5
+	: expr_core
+	;
+
+expr_core
```
patmagee, line 121: See https://github.com/openwdl/wdl/pull/364 for an ordering issue that @orodeh reported

versions/draft-2/parsers/antlr4/WdlDraft2Parser.g4
```diff
@@ -0,0 +1,268 @@
...
+
+task_runtime
+	: RUNTIME LBRACE (task_runtime_kv)* RBRACE
+	;
+
+task_input
+	: (any_decls)*
+	;
+
+task_output
```
patmagee, line 169: in the draft-2 specification (apparantly) there is a secondary syntax supported for outputs of tasks and workflows. This syntax cannot be mixed with the declarative syntax.

```wdl
workflow foo {
  call a
  call b
  output {
    a.* # return ALL of the outputs from a
    b.value # return a specific output from b
  }
}
```

versions/draft-2/parsers/antlr4/WdlDraft2Parser.g4
```diff
@@ -0,0 +1,268 @@
...
+conditional
+	: IF LPAREN expr RPAREN LBRACE inner_workflow_element* RBRACE
+	;
+
+workflow_input
+	: (any_decls)*
+	;
+
+workflow_output
+	: OUTPUT LBRACE (bound_decls)* RBRACE
```
patmagee, line 247: See comment above in regards to task outputs

versions/draft-2/parsers/antlr4/WdlDraft2Parser.g4
jdidion: In wdlTools we are disallowing that. I am not sure the right grammar for it so feel free to add it if you want it in the official grammar.

jdidion: We disallow this in wdlTools. If you think it should be in the official spec feel free to add it.

PR #362, opened 2020-04-03
Added Suffix and similar Quoting Engine Functions

patmagee:
Introduce a new engine function for adding a suffix to elements in an array. This is the natural complement to the `prefix` function.

The desire for this function came from a discussion in the [Slack Support Channel](https://openwdl.slack.com/archives/CTE06LC85/p1585891984096500), where a user wanted to add  a `"` around each element in an array of strings before passing it to the `sep` placeholder in a command

Additonally, added two convenience functions: `quote(Array[X])` and `squote(Array[X])` which work functionally similar to `suffix("'",prefix("'",x))`

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,8 @@ Keep the changelog pleasant to read in the text editor:
 
 version 2.0.0
 ---------------------------
-+ Added an engine function for adding a suffix to an array of primitives
++ Added an engine function for adding a suffix to an array of primitives as well
+  as well as `quote` and `squote` engine functions.
   [PR 362](https://github.com/openwdl/wdl/pull/362) @patmagee
 
 + Added a required input and output format for workflow engines.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ Keep the changelog pleasant to read in the text editor:
 
 version 2.0.0
 ---------------------------
++ Added an engine function for adding a suffix to an array of primitives
+
 + Added a required input and output format for workflow engines.
   [PR 357](https://github.com/openwdl/wdl/pull/357)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ Keep the changelog pleasant to read in the text editor:
 version 2.0.0
 ---------------------------
 + Added an engine function for adding a suffix to an array of primitives
+  [PR 362](https://github.com/openwdl/wdl/pull/362) @patmagee
 
 + Added a required input and output format for workflow engines.
   [PR 357](https://github.com/openwdl/wdl/pull/357)
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -127,6 +127,8 @@
   * [Array\[X\] flatten(Array\[Array\[X\]\])](#arrayx-flattenarrayarrayx)
   * [Array\[String\] prefix(String, Array\[X\])](#arraystring-prefixstring-arrayx)
   * [Array\[String\] suffix(String, Array\[X\])](#arraystring-suffixstring-arrayx)
+  * [Array\[String\] quote(Array\[X\])](#arraystring-quotearrayx)
+  * [Array\[String\] squote(Array\[X\])](#arraystring-squotearrayx)
   * [X select_first(Array\[X?\])](#x-select_firstarrayx)
   * [Array\[X\] select_all(Array\[X?\])](#arrayx-select_allarrayx)
   * [Boolean defined(X?)](#boolean-definedx)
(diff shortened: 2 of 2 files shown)


illusional:
I support this change so much. Most of the issues I’ve been recently having are related to quotation if parameters.

This is a massive win for me, great PR!

illusional:
On second review @patmagee, could we also make it take a single string as well as an array of Strings. I would hope that the engine would escape any single or dual characters in the original string.

rhpvorderman:
@illusional You could always take a single element and put it in an array using `quote([element])`. This is also commonly done with `select_first([optional_element])`.

@patmagee Great PR. I say we start the vote as soon as possible :tada:.

patmagee:
@illusional I was toying with the idea of accepting and returning a single string. I feel like that is a good direction to move in, however I feel like I want these to perform the exact same as `suffix` and `prefix` do at the moment for consistency. Maybe in a future PR we can make all of those work for `String` or `Array[String]` elements.

WDYT?

illusional:
Would the quote function take a separator argument? If so sounds good to me. You could nest a prefix function to apply the prefix to each element. But yeah sounds good to me! 

I’ve been thinking about a join function as it would allow me to better pass nested arrays to command lines.

patmagee:
@illusional I dont think it would, but you could use the new `sep` function thats part of 2.0 as well to do something like:

```wdl
command <<< 
    echo ~{sep(quote([x]))}
>>>
````

**Comment by illusional** [comment]
@patmagee Ignore me, it was late at night when I was commenting and I misunderstood. I think it would be good to accept and return a single string, though it's not hard to work around it with using an array literal (and potentially `select_first`).

**Comment by illusional** [comment]
Realised this change encompasses #230 which has been closed.

**Comment by rhpvorderman** [comment]
@illusional `sep` is already defined in WDL 1.0 and can already be used: https://github.com/openwdl/wdl/blob/master/versions/1.0/SPEC.md#sep .

**Comment by illusional** [comment]
@rhpvorderman It's defined as a "Expression Placeholder Options", but can't be used as it's own function. Is that something we could add to this PR?

I tested this task across:

- Cromwell 49
    ```
    Failed to read task definition at line 3 column 6 (reason 1 of 1): Failed to read declaration 
    (reason 1 of 1): Failed to parse expression (reason 1 of 1): Unknown engine function: 'sep'
    ```

- MiniWDL v0.7.3
    ```
    (test_sep_operator.wdl Ln 8 Col 19) No such function: sep
          String joined = sep(inp, ", ")
    ```

**task**
```wdl
version development

task test_sep {
  input {
    Array[String] inp = ["val1", "val2", "val3"]
  }

  String joined = sep(inp, ", ")

  command <<<
    echo '~{joined}'
  >>>
  output {
    String out = read_string(stdout())
  }
}
```

illusional:
👍

orodeh:
👍

pieterlukasse:
Should WDL really be the place to do string manipulations? I'm still new to WDL, and perhaps I'm overlooking some things...But I'm also trying to align this with the description in https://github.com/openwdl/wdl#workflow-description-language-wdl, which sounds more like WDL's purpose is to be a very high level description language. I worry that there is a risk that more features like the one in this PR will increase the amount of "programming" found in .wdl files. At the same time, I do understand that it can be useful for basic plumbing activities of linking one step to the next one. I'm not ready to vote in favor of this feature, but perhaps someone can convince me by addressing some of the points above ;)

patmagee:
@pieterlukasse I agree with you that we should always be diligent to make sure that WDL does not tend towards a full blown programming language. However, there are a number of cases, where NOT having these features actually makes this more difficult and require more knowledge in programming. For reference, this issue has been encountered by a number of WDL users with no easy way to solve

the main case where this was meant to handle is the "Pass an array of things to a command". If we do not have functions like quote, sep, squote, etc, then interacting with this array becomes unintuitive and actually requires alot of careful handling.

For example, take the following case where we want to pass a list of "things" to a command parameters
```
Array[String] things = ["thing-1", "thing-2 with-space"]
my-command --parameter ~{ sep(" ", things)} #--parameter thing-1 --parameter thing-2 with-space
```

this will fail if the strings within "things" have spaces within them and the user failed to properly wrap each string with an escaped quote. beacuse of the way that Bash works. In order to make this work you would need to do some... complex bash that is both ugly and difficult to understand.

whereas with the proposed changes would make this parsed properly

```
Array[String] things = ["thing-1", "thing-2 with-space"]
my-command --parameter ~{ sep(" ", quote(things))} #--parameter "thing-1" --parameter "thing-2 with-space"
```
I think a good rule of thumb when considering new features like this, is to consider what problem it is trying to solve and how ugly the solution is without it. This is one of those cases where the solution requires MORE programming logic that is VERY difficult to reliably do without WDL functions

rhpvorderman:
:+1:

pieterlukasse:
@patmagee thanks for the detailed explanation. I will be pondering on it for a while... Meanwhile, one quick question I had regarding your specific example of `sep` + `quote` is: wouldn't it maybe be better to let the implementation of `sep` solve this problem directly? I.e. when "things" is `Array[String]`, it could automatically quote them...this could be the _default_ behavior. IMO this could help towards the "principles of least surprise" in this particular example, and it would decrease the amount of coding in .wdl itself for the most common use case.

patmagee:
@pieterlukasse sep shouldnt change the prefix or the suffix of a string. its behaviour should simply be to take an array of primitives and combine them into a single string with the specified separator. Additional behaviour beyond this is unintuitive and in fact would probly be undesirable in many situations.

aednichols:
👍

cjllanwarne:
👍

patmagee:
This Change has passed with a vote of 5-1, and is now awaiting merging of https://github.com/chanzuckerberg/miniwdl/pull/379

mlin:
Now that this is done, I belatedly wonder if we might have wanted a function similar to [Python's `shlex.quote`](https://docs.python.org/3/library/shlex.html#shlex.quote) to handle the corner cases where the strings to be quoted themselves contain quote marks. (However, I don't know how to write a detailed specification for such a function)

PR #367, opened 2020-06-03
Stop parsing signed numbers as literals

jdidion:
Including SignedDigits in IntLiteral (and SignedFloatFragment in FloatLiteral) causes the lexer to always parse a `+` or `-` followed directly by a number as a signed literal, even in the context of expressions where the +/- should be treated as an operator. For example: `i + 5` parses fine but `i +5` does not. Just as in the Hermes grammar, `+5` should be treated as either a unary expression or the RHS of an addition, and it is up to the expression evaluator to convert a stand-alone unary operation to a signed number.


patmagee:
@jdidion lgtm! Feel freee to merge

PR #369, opened 2020-06-04
Add null literal

jdidion:
This goes along with the previous PR - null can now be recognized as a meta value without screwing up expression parsing.


versions/1.0/parsers/antlr4/WdlLexer.g4
```diff
@@ -35,7 +35,7 @@ MAP: 'Map';
 PAIR: 'Pair';
 OBJECT: 'Object';
 OBJECT_LITERAL: 'object';
-
+NULL_LITERAL: 'null'
```
patmagee, line 38: Since `null` is not actually part of the WDL 1.0 spec (but strangely allowed in the meta section) I would probably ammend your previous PR, and add the `NULL_LITERAL` to the `meta` mode

PR #375, opened 2020-06-22
Add Draft2 ANTLR grammar

jdidion:
This PR replaces the previous one. The grammar here is the one used in the first release version of wdlTools. All issues from the previous PR are addressed with the exception of supporting the alternative workflow outputs syntax. If we want to support that syntax I request that @patmagee add it.

### Checklist


jdidion:
@patmagee in draft-2 the only meta values that are allowed are strings>

PR #414, opened 2020-11-04
Multi-line strings

jdidion:
The ability to have multi-line strings has been requested several times. It would also make easier several of the other proposals under consideration, such as in-line Dockerfile and evaluation of arbitrary expressions. I went with Python/Scala-style rather than HEREDOC style as that seems most idiomatic for WDL.

This is implemented in the v2 ANTLR4 grammar in a wdlTools branch: https://github.com/dnanexus-rnd/wdlTools/tree/multiline-strings/src/main/antlr4/v2

### Checklist

### DIFF
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -364,27 +364,29 @@ Strings can also contain the following types of escape sequences:
 Strings that begin and end with three consecutive single- or double-quotes may span multiple lines. The opening quotes of a multi-line string may be followed by whitespace (optional) and must have a newline before the first non-whitespace character; these leading whitespace/newline characters are removed. All subsequent non-empty lines are then used to determine the multi-line string's *common leading whitespace* - the minimum number of whitespace characters occuring before the first non-whitespace character in a line or the end of the line, whichever comes first. This common leading whitespace is stripped from the beginning of all the lines in the multi-line string.
 
 ```wdl
-# These three strings are equivalent. The middle line of each is empty and so does not count
-# towards the common leading whitespace determination.
+# These strings are equivalent. The middle lines of strings B, C, and D ire empty and so do 
+# not count towards the common leading whitespace determination.
 
-String multi_line_A = '''
+String multi_line_A = """
+    this is a
+  a multi-line string"""
+
+String multi_line_B = '''
         this is a
 
           multi-line string'''
 
-String multi_line_B = """
+String multi_line_C = """
   this is a
 
     multi-line string"""
 
-String multi_line_C = """
+String multi_line_D = """
 this is a
 
   multi-line string"""
 ```
-String multi_line_D = """
-    this is a
-  a multi-line string"""
+
 Newline characters are not stripped out unless they are escaped, i.e. when the last character of a line is `\`.
 
 ```wdl
@@ -1619,7 +1621,7 @@ Keep in mind that the command section is still subject to the rules of [string i
 
 #### Stripping Leading Whitespace
 
-When a command template is evaluate, the execution engine first strips out all *common leading whitespace* (just like [multi-line strings](#multi-line-strings)).
+When a command template is evaluated, the execution engine first strips out all *common leading whitespace* (just like [multi-line strings](#multi-line-strings)).
 
 For example, consider a task that calls the `python` interpreter with an in-line Python script:
 
@@ -1652,7 +1654,7 @@ python <<CODE
 CODE
 ```
 
-Each whitespace character is counted regardless of whether it is a space or tab, so care should be taken when mixing whitespace characters. For example, if a command block has two lines, and the first line begins with `<space><space><space><space>`, and the second line begins with `<tab>` then only one whitespace character is removed from each line.
+If the user mixes tabs and spaces, the behavior is undefined. The execution engine should, at a minimum, issue a warning and leave the whitespace unmodified, though it may choose to raise an exception or to substitute e.g. 4 spaces per tab.
 
 ### Task Outputs
 
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -367,6 +367,7 @@ String s = <<<
   This is a
   multi-line string!
 >>>
+```
 
 In multi-line strings, leading *whitespace* is removed according to the following rules. In the context of multi-line strings, whitespace refers to space (`\x20`) and tab characters only and is treated differently from newline characters.
 
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -378,7 +378,7 @@ In multi-line strings, leading *whitespace* is removed according to the followin
 4. Use all remaining non-*blank* lines to determine the *common leading whitespace*.
   * A blank line contains zero or more whitespace characters followed by a newline.
   * Common leading whitespace is the minimum number of whitespace characters occuring before the first non-whitespace character in a non-blank line.
-  * Each whitespace character is counted once regardless of whether it is a space or tab, so care should be taken when mixing whitespace characters.
+  * Each whitespace character is counted once regardless of whether it is a space or tab (so care should be taken when mixing whitespace characters).
 5. Remove common leading whitespace from each line.
 
 ```wdl
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -1582,8 +1582,7 @@ workflow wf {
 
 ### Command Section
 
-The `command` section is the only required task section. It defines the command template that is evaluated and executed when the task is called. The command template is a `bash` script that may
-contain placeholder expressions. There may be any number of commands within a command section.
+The `command` section is the only required task section. It defines the command template that is evaluated and executed when the task is called. The command template is a `bash` script that may contain placeholder expressions. There may be any number of commands within a command section.
 
 ```wdl
 command <<<
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -382,7 +382,9 @@ this is a
 
   multi-line string"""
 ```
-
+String multi_line_D = """
+    this is a
+  a multi-line string"""
 Newline characters are not stripped out unless they are escaped, i.e. when the last character of a line is `\`.
 
 ```wdl
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -881,7 +881,7 @@ String s = read_string("/path/to/file")
 
 WDL provides the standard unary and binary mathematical and logical operators. The following table lists the valid operand and result type combinations for each operator. Using an operator with unsupported types results in an error.
 
-In operations on mismatched numeric types (e.g. `Int` + `Float`), the `Int` type is first cast to a `Float`; the result type is always `Float`. This may result in loss of precision, for example if the `Int` is too large to be represented exactly by the `Float`. Note that a `Float` can be converted to an `Int` with the [`ceil`, `round`, or `floor`](#int-floorfloat-int-ceilfloat-and-int-roundfloat) functions.
+In operations on mismatched numeric types (e.g. `Int` + `Float`), the `Int` type is first cast to a `Float`; the result type is always `Float`. This may result in loss of precision, for example if the `Int` is too large to be represented exactly by the `Float`. A `Float` can be converted to an `Int` with the [`ceil`, `round`, or `floor`](#int-floorfloat-int-ceilfloat-and-int-roundfloat) functions.
 
 ##### Unary Operators
 
@@ -988,7 +988,7 @@ Boolean is_false1 = [1, 2, 3] == [2, 1, 3]
 Boolean is_false2 = {"a": 1, "b": 2} == {"b": 2, "a": 1}
 ```
 
-Note that [type coercion](#type-coercion) can be employed to compare values of different but compatible types. For example:
+[Type coercion](#type-coercion) can be employed to compare values of different but compatible types. For example:
 
 ```wdl
 Array[Int] i = [1,2,3]
@@ -1091,7 +1091,7 @@ WDL provides a [standard library](#standard-library) of functions. These functio
 
 #### Expression Placeholders and String Interpolation
 
-Any WDL string expression may contain one or more "placeholders" of the form `~{*expression*}`, each of which contains a single expression. Note that placeholders of the form `${*expression*}` may also be used interchangably, but their use is discouraged for reasons discussed in the [command section](#expression-placeholders) and may be deprecated in a future version of the specification.
+Any WDL string expression may contain one or more "placeholders" of the form `~{*expression*}`, each of which contains a single expression. Placeholders of the form `${*expression*}` may also be used interchangably, but their use is discouraged for reasons discussed in the [command section](#expression-placeholders) and may be deprecated in a future version of the specification.
 
 When a string expression is evaluated, its placeholders are evaluated first, and their values are then substituted for the placeholders in the containing string.
 
@@ -1122,6 +1122,14 @@ Int i = 3
 Boolean b = true
 # s evaluates to "4", but would be "0" if b were false
 String s = "~{if b then '${1 + i}' else 0}"
+```
+
+Placeholders are evaluated in multi-line strings exactly the same as in regular strings:
+
+```wdl
+String multi_line = """
+  Hello ~{name}
+  Welcome to ~{company}"""
 ```
 
 ##### Expression Placeholder Coercion
@@ -1555,7 +1563,7 @@ There are two different syntaxes that can be used to define command expression p
 | `command <<< >>>`        | `~{}` only                 |
 | `command { ... }`        | `~{}` (preferred) or `${}` |
 
-Note that the `~{}` and `${}` styles may be used interchangably in other string expressions.
+The `~{}` and `${}` styles may be used interchangably in other string expressions.
 
 Any valid WDL expression may be used within a placeholder. For example, a command might reference an input to the task, like this:
 
@@ -2653,7 +2661,7 @@ workflow wf {
 }
 ```
 
-Note that there is no mechanism for a workflow to set a value for a nested input when calling a subworkflow. For example, the following workflow is invalid:
+There is no mechanism for a workflow to set a value for a nested input when calling a subworkflow. For example, the following workflow is invalid:
 
 `sub.wdl`
 ```wdl
@@ -2702,7 +2710,7 @@ workflow abbrev {
 
 Calls may be executed as soon as all their inputs are available. If `call x`'s inputs are based on `call y`'s outputs, this means that `call x` can be run as soon as `call y` has completed. 
 
-As soon as the execution of a called task completes, the call outputs are available to be used as inputs to other calls in the workflow or as workflow outputs. Note that the only task declarations that are accessible outside of the task are its output declarations, i.e. call inputs cannot be referenced. To expose a call input, add an output to the task that simply copies the input:
+As soon as the execution of a called task completes, the call outputs are available to be used as inputs to other calls in the workflow or as workflow outputs. The only task declarations that are accessible outside of the task are its output declarations, i.e. call inputs cannot be referenced. To expose a call input, add an output to the task that simply copies the input:
 
 ```wdl
 task copy_input {
@@ -2729,7 +2737,7 @@ workflow test {
 }
 ```
 
-To add a dependency from x to y that isn't based on outputs, you can use the `after` keyword, such as `call x after y after z`. But note that this is only required if `x` doesn't already depend on an output from `y`.
+To add a dependency from x to y that isn't based on outputs, you can use the `after` keyword, such as `call x after y after z`. However, this is only required if `x` doesn't already depend on an output from `y`.
 
 ```wdl
 task my_task {
@@ -3163,7 +3171,7 @@ workflow cond_test {
 
 The scoping rules for conditionals are similar to those for scatters. Any declarations or call outputs inside a conditional body are accessible within that conditional and any nested scatter or conditional blocks. After a conditional block has been evaluated, its declarations and call outputs are "exported" to the enclosing scope. However, because the statements within a conditional block may or may not be evaluated during any given execution of the workflow, the type of each exported declarations or call output is implicitly `X?`, where `X` is the type of the declaration or call output within the conditional body.
 
-Note that, even though a conditional body is only evaluated if its conditional expression evaluates to `true`, all of the potential declarations and call outputs in the conditional body are always exported, regardless of the value of the conditional expression. In the case that the conditional expression evaluates to `false`, all of the exported declarations and call outputs are undefined (i.e. have a value of `None`).
+Even though a conditional body is only evaluated if its conditional expression evaluates to `true`, all of the potential declarations and call outputs in the conditional body are always exported, regardless of the value of the conditional expression. In the case that the conditional expression evaluates to `false`, all of the exported declarations and call outputs are undefined (i.e. have a value of `None`).
 
 ```wdl
 workflow foo {
@@ -3185,7 +3193,7 @@ workflow foo {
 }
 ```
 
-Also note that it is impossible to have a multi-level optional type, e.g. `Int??`; thus, the outputs of a conditional block are only ever single-level optionals, even when there are nested conditionals.
+It is impossible to have a multi-level optional type, e.g. `Int??`; thus, the outputs of a conditional block are only ever single-level optionals, even when there are nested conditionals.
 
 ```wdl
 workflow foo {
@@ -3527,7 +3535,7 @@ workflow max_test {
 
 Given 3 String parameters `input`, `pattern`, `replace`, this function replaces all non-overlapping occurrences of `pattern` in `input` by `replace`. `pattern` is a [regular expression](https://en.wikipedia.org/wiki/Regular_expression) that will be evaluated as a [POSIX Extended Regular Expression (ERE)](https://en.wikipedia.org/wiki/Regular_expression#POSIX_basic_and_extended).
 
-Note that regular expressions are written using regular WDL strings, so backslash characters need to be double-escaped. For example:
+Regular expressions are written using regular WDL strings, so backslash characters need to be double-escaped. For example:
 
 ```wdl
 String s1 = "hello\tBob"
@@ -3867,7 +3875,7 @@ task do_stuff {
 
 Reads an entire file as a string, with any trailing end-of-line characters (`\r` and `\n`) stripped off. If the file is empty, an empty string is returned.
 
-Note that if the file contains any internal newline characters, they are left intact. For example:
+If the file contains any internal newline characters, they are left intact. For example:
 
 ```wdl
 # this file will contain "this\nfile\nhas\nfive\nlines\n"
@@ -5292,7 +5300,7 @@ task test {
 | --- | ----- |
 | foo | bar   |
 
-Note that using `write_json`/`read_json` to serialize to/from a `Map` can cause subtle issues due to the fact that `Map` is ordered whereas an object value is not. For example:
+Using `write_json`/`read_json` to serialize to/from a `Map` can lead to subtle issues due to the fact that `Map` is ordered whereas an object value is not. For example:
 
 ```wdl
 Map[String, Int] s2i = {"b": 2, "a": 1}
--- a/versions/development/SPEC.md
+++ b/versions/development/SPEC.md
@@ -1600,8 +1600,7 @@ The command template is evaluated *after* all of the inputs are staged and befor
 4. Remove common leading whitespace from each line.
 5. Evaluate placeholder expressions.
 
-Notice that there is one major difference between the evaluation of multi-line strings vs the command template: line continuations are removed in the former but left as-is in the latter. This also means that continued lines are considered when determining common leading whitespace, and that
-common leading whitespace is removed from continued lines as well.
+Notice that there is one major difference between the evaluation of multi-line strings vs the command template: line continuations are removed in the former but left as-is in the latter. This also means that continued lines are considered when determining common leading whitespace, and that common leading whitespace is removed from continued lines as well.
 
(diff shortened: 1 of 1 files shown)


patmagee:
@jdidion This is amazing. I believe this will be considered a highly welcome addition to the WDL specification!

cjllanwarne:
* IMO the "least surprising" behavior would be that all whitespace in the multi-line string are preserved by default. 
* I like the idea of starting a line with `|` to indicate "strip margin" but in practice, people might try to use this to generate bash commands, in which case erasing `|`s and `>` feels dangerous. Could that be a function (eg like `strip_margin(multiline_string)` rather than default behavior)?

cjllanwarne:
We could also have a `remove_newlines(multiline_string)` to replace the "we remove all newlines by default" behavior currently propsed

jdidion:
My feeling was that the most common usage for multi-line strings will be to break up long strings. If you want to write bash, you put that in the command block, no?

I don't think we need a special `remove_newlines` function - just use `sub(str, "\n", " ")`.

cjllanwarne:
Coming from scala might be distorting my sense of "what people expect" w.r.t. newlines in multi-line strings.

Usually the "what would people expect" tie-breaker in cases of different aesthetic preferences is "what would python do", which seems appropriate here too.

cjllanwarne:
As far as I can tell, python also preserves newlines in multi-line string literals (https://www.techbeamers.com/python-multiline-string/). Unless there's a really strong case against it, I think that should be what we do.

Yes, I think bash commands _usually_ go in command blocks, but I still think having a syntax that erases very common bash-style syntax by default in multi-line string literals feels dangerous to me

jdidion:
Personally, I find it very annoying to have to litter my Scala code with `.stripMargin.replaceAll("\n", " ")`. I agree that "least surprising" is a good approach, but I also feel it can be over-ridden by the desire to simplify the most common use case.

cjllanwarne:
I'd argue with your premise there though that the most common use case is "I want to split a single line string across many lines". I think the most common usage would be "I want a to write a multi-line string directly in my WDL". In which case the simplest way to provide that is to make it the default.

But we're arguing our own personal preferences here - which will never have a definitive answer. I think "choose the options that's most similar to python" is the most objective possible way to resolve these discussions.

jdidion:
Fair enough. Let's wait to get a bit more feedback from others. I'm happy to update the spec and grammar if the consensus is to leave in whitespace by default.

cjllanwarne:
Agreed! An n > 2 is always a good idea...

patmagee:
Personally, I find the least surprising thing to be represent the string "as-is" with new lines intact. for me that is the main point of a multiline string, is to write across multiple lines. The tricky part I see is around preserving indents or removing them. I believe in python (since its tab/space delimeted) this is easy, since you simply strip out leading tabs corresponding to the current indent level. In WDL this is a bit trickier. we could

1. Not touch anything 
2. Strip the least number of common tabs
3. Define a syntax (ie `|` ) for retaining those indents

jdidion:
Ok, how about:

* By default, leave newlines and strings alone
* If a line is prefixed by `|`, strip all the leading whitespace before it (as well as the `|` itself)
* If the user wants to remove newlines they have to `sub(str, "\n", " ")`

cjllanwarne:
I really worry about strings like
```
String command_prefix = """echo hi \
  | grep foo \
  | wc -l"""
```

jdidion:
Oh, good point.

In that case I guess we may need a strip_whitespace function. By default this would strip the least number of common tabs. We could have an optional argument that would do any/all of the following:

* If it is an integer, strip that many whitespace characters
* If it is a whitespace-only string, strip exactly that whitespace from the beginning of each line
* If it is a single character, strip all whitespace up to and including that character any time that character is the first non-whitespace character in the line

jdidion:
A simpler alternative would be a `dedent` function (akin to the one in python textwrap https://docs.python.org/3/library/textwrap.html#textwrap.dedent) that just removes leading common whitespace, and not try to have a syntax for margin.

cjllanwarne:
Those ideas all sound reasonable to me.

pshapiro4broad:
Have you looked at the spec for Java text blocks? The java spec supports both automatically stripping newlines and preserving relative indent and doesn't require a lot of extra syntax. https://openjdk.java.net/jeps/378

cjllanwarne:
I do like that java text block spec's interpretation of "intrinsic whitespace", which matches very closely the `command` block behavior.

patmagee:
Ive just started using Java's implementaiton of text blocks and they are very intuitive, so I give a big +1 to that!

patmagee:
is the discussion on how to handle whitespace still live or do we have general consensus?

jmesterh:
This would be really helpful, we have a large CWL pipeline we want to convert to WDL which uses YAML text blocks extensively for SQL queries. Not having an equivalent in WDL has been frustrating, as the SQL needs to be squashed into a single line. This feature would be **greatly** appreciated.

patmagee:
@jdidion any update on the status of this? It would be great to get this to voting

jmesterh:
FWIW I'd vote for the Java text block syntax. Since my last post we've resorted to a [yaml → wdl converter](https://github.com/susom/wdl-kit/blob/main/src/main/python/utils/yaml2wdl.py) which isn't fun.

jdidion:
I am good with Java text block syntax. I'll work on an update to the spec and grammar. In the meantime, let's see if we can come up with any counter-examples - i.e. cases where one would *not* want leading whitespace to be automatically removed.

jdidion:
PR updated https://github.com/openwdl/wdl/pull/414

jdidion:
For now I went with option 2. I also updated the section on stripping leading whitespace from commands to make it consistent.

It looks like my editor also reformatted all the tables in the spec. If that's a problem I can redo the commit.

geoffjentry:
I'm good w/ this

versions/development/SPEC.md
```diff
@@ -344,20 +345,77 @@ Tasks and workflow inputs may be passed in from an external source, or they may
...
+  this is a
+
+    multi-line string"""
+
+String multi_line_C = """
+this is a
+
+  multi-line string"""
+```
+
```
illusional, line 385: I think it would be good add an example where the first line is longer than other lines:

```suggestion
String multi_line_D = """
    this is a
  a multi-line string"""
```

versions/development/SPEC.md
```diff
@@ -344,20 +345,77 @@ Tasks and workflow inputs may be passed in from an external source, or they may
...
+```wdl
+# The following two strings are equivalent:
+
+String s1 = "    text indented by 4 spaces"
+
+# Even though the first line of this string is indented by six spaces, the common leading
+# whitespace in this string is 2, due to the two spaces preceeding the closing quotes.
+String s2 = """
+      text indented by 4 spaces
+  """
```
illusional, line 408: IMO, if the last line is whitespace only, it shouldn't count towards the _common leading whitespace_.

illusional (commented):
LGTM, one small suggestion.

versions/development/SPEC.md
jdidion: This is in line with how Java text blocks work.
e.g. see the example under section 3 here https://www.baeldung.com/java-text-blocks

mlin:
Should we just explicate that/whether expression placeholders work inside multi-line strings?

jdidion:
Thanks @mlin - I added a note on this in the section on string interpolation.

patmagee:
Looks good to me!

versions/development/SPEC.md
```diff
@@ -344,20 +345,79 @@ Tasks and workflow inputs may be passed in from an external source, or they may
...
+String multi_line_escaped = """
+  this is a \
+  double-quoted string \
+  that contains no newlines"""
+```
+
+Keep in mind that if the ending quotes are on a line by themselves and are preceeded by whitespace, that line is included when determining the common leading whitespace.
+
+```wdl
+# The following two strings are equivalent:
```
mlin, line 402: Is one of the "two strings" missing?

versions/development/SPEC.md
```diff
@@ -344,20 +345,79 @@ Tasks and workflow inputs may be passed in from an external source, or they may
...
 
 Strings can also contain the following types of escape sequences:
 
 * An octal escape code starts with `\`, followed by 3 digits of value 0 through 7 inclusive.
 * A hexadecimal escape code starts with `\x`, followed by 2 hexadecimal digits `0-9a-fA-F`. 
 * A unicode code point starts with `\u` followed by 4 hexadecimal characters or `\U` followed by 8 hexadecimal characters `0-9a-fA-F`.
 
+##### Multi-line Strings
+
+Strings that begin and end with three consecutive single- or double-quotes may span multiple lines. The opening quotes of a multi-line string may be followed by whitespace (optional) and must have a newline before the first non-whitespace character; these leading whitespace/newline characters are removed. All subsequent non-empty lines are then used to determine the multi-line string's *common leading whitespace* - the minimum number of whitespace characters occuring before the first non-whitespace character in a line or the end of the line, whichever comes first. This common leading whitespace is stripped from the beginning of all the lines in the multi-line string.
```
mlin, line 364: > must have a newline before the first non-whitespace character

This 'must' is a little surprising, where does it come from? NBD but sometimes it's useful to write one-line string literals that contain both single and double quotes, e.g. in python:

```
>>> print('''He continued in another tweet on Aug. 29: "When you see 'anonymous source,' stop reading the story, it is fiction!"''')
He continued in another tweet on Aug. 29: "When you see 'anonymous source,' stop reading the story, it is fiction!"
```

patmagee: @mlin this comes from the Java specification which is the inspiration for the multi line strings

rhpvorderman: That makes sense. In python the multi-line strings must be dedented using the `textwrap.dedent` function if there is leading whitespace. I do not want that sort of helper functions to be necessary in WDL. So java-style multiline strings that auto-strip the indentation make the most sense to me. Even if that means there are extra requirements to writing the multiline string.

versions/development/SPEC.md
rhpvorderman: But it is not in line with how the command block works in WDL. I predict this is going to lead to questions.
```WDL
String s2 = """
    Text that I do not want indented
    But I do want it indented in the source file
    Indentation in the source makes it easier to read
""" 

 # Looks correct, but leads to the wrong result
```
```WDL
String s2 = """
    Text that I do not want indented
    But I do want it indented in the source file
    Indentation in the source makes it easier to read"""

# Looks wrong, but leads to the correct result.
```

rhpvorderman:
I have not much to add to what has been already said. It is quite a hard problem to find a syntax that satisfies everyone. I do not have a very strong opinion about this, except that I do not want to need some extra `dedent` like functionality. 

Having said that, I wish the formatting changes of the tables etc. that are in the diff are tackled in another matter. IMO these can be merged without PR. Now the diff is larger than what is actually discussed.

versions/development/SPEC.md
jdidion: No, but maybe that's not clear. Please let me know if the change clears things up.

versions/development/SPEC.md
jdidion: We could provide the behavior @mlin is looking for at the expense of making the specification more complex. We'd have to decide what behavior we want in the case of a multi-line string without a newline before the first non-whitespace:

1. Don't consider the first line when determining common leading whitespace
2. Determine leading whitespace the same way for the first line as subsequent lines

Consider the following examples:

```wdl
String s1 = """This is
    a multi-line string"""
String s2 = """  This is
    a multi-line string"""
String s3 = """    This is
    a multi-line string"""
```

If we go with option 1, then the common leading whitespace is 4 for for all three strings. Also, in `s2` and `s3`, do we remove the leading whitespace from the first line or not?

If we go with option 2, then the common leading whitespace is 0 for `s1`, 2 for `s2`, and 4 for `s3`.

I don't think either of these options is particularly intuitive, but option 2 would (IMHO) be less surprising behavior than option 1.

jdidion:
Ok @rhpvorderman I'll reformat the tables in `main` and then rebase this PR.

versions/development/SPEC.md
jdidion: You're right. And actually I don't really like this aspect of java text blocks because the handling of the first and last lines feels inconsistent. If we ignore whitespace on the first line we should also ignore it on the last line. That consistency - as well as having consistency between multi-line strings and the command block, seems more important to me then the small benefit of being able to use the last line to "override" the indentation of the rest of the string. I am going to update the spec and we'll see what others think.

versions/development/SPEC.md
mlin: I'm not going to veto anything here, just noting I was surprised

versions/development/SPEC.md
```diff
@@ -358,6 +359,65 @@ Strings can also contain the following types of escape sequences:
...
+
+```wdl
+# These strings are equivalent. The middle lines strings B, C, and D are empty and so do not count
+# towards the common leading whitespace determination.
+
+String multi_line_A = """
+    this is a
+  a multi-line string"""
```
mlin, line 372: "a" shouldn't appear both at the end of one line and the beginning of the next here, right?

mlin: Doesn't multi_line_A start with two spaces after removal of common leading whitespace? (which isn't the same as B/C/D)

versions/development/SPEC.md
```diff
@@ -358,6 +359,65 @@ Strings can also contain the following types of escape sequences:
...
+```wdl
+# These two strings are equivalent:
+String single_line = "this is a double-quoted string that contains no newlines"
+String multi_line_escaped = """
+  this is a \
+  double-quoted string \
+  that contains no newlines"""
+```
+
+Keep in mind that if the ending quotes are on a line by themselves and are preceeded by whitespace, that line is included when determining the common leading whitespace.
```
mlin, line 401: Nit- the rest of the diff goes out of its way to delete phrases like "Keep in mind that"

versions/development/SPEC.md
```diff
@@ -358,6 +359,65 @@ Strings can also contain the following types of escape sequences:
...
+
+```wdl
+# The following two strings are equivalent. Even though the first line of `s2` is indented by six 
+# spaces, the common leading whitespace in this string is 2, due to the two spaces preceeding the 
+# closing quotes.
+String s1 = "    text indented by 4 spaces"
+String s2 = """
+      text indented by 4 spaces
+  """
```
mlin, line 410: Notwithstanding the above comments on this line, I'm also not sure why wouldn't s2 end with a newline? (i.e. shouldn't s1 end in a newline to equal s2?)

versions/development/SPEC.md
rhpvorderman: That is an excellent question. Which shows that the Java people already thought about this quite well. In case I want, or do not care about the final newline I can also write:
```
String s2 = """
    Text that I do not want indented
    But I do want it indented in the source file
    Indentation in the source makes it easier to read
    """ 

 # Looks nice and properly strips indentation.
```
You have convinced me that we should keep the java spec. It is simple and there is no ambiguity. I see now that suggestions like the above only serve to introduce corner cases which are ambiguous, and that is not desirable.

mlin:
Hi all, I have a miniwdl branch with a draft implementation of this feature:

https://github.com/chanzuckerberg/miniwdl/pull/612

It's mostly working although I'm sure there are still bugs to iron out. I appreciate if anyone (@illusional @rhpvorderman @patmagee ?) has time to try adding more tricky test cases to the framework that's there, either in python (`python3 -m unittest -f tests.test_0eval.TestEval.test_multiline_string`) or the test WDL (`python3 -m WDL run tests/multistring.wdl --dir /tmp --verbose`) -- the latter avoids dealing with two levels of escaping.

One more worthwhile clarification @jdidion would be whether common whitespace detection occurs before or after placeholder evaluation. It feels like the whitespace detection should go first, but it's worth clarifying because that way is actually more difficult to implement (since we have to detect the common whitespace in the context of an expression tree, not just a single block of text).

rhpvorderman:
@mlin, would it be easier to detect it in a two step process. I.e just get the common whitespace just as it is done with the command block, then after that evaluate the multiline string extraneous whitepace separately?

versions/development/SPEC.md
mlin: I def stumbled with this once or twice writing the test cases. If we keep the last-line whitespace significant then maybe we also add a linter warning when it's determining the common leading whitespace. (I'm not vetoing since I think "what java does" is a reasonable heuristic, but I also think @rhpvorderman example is persuasive and now backed up by practical experience)

mlin:
> @mlin, would it be easier to detect it in a two step process. I.e just get the common whitespace just as it is done with the command block, then after that evaluate the multiline string extraneous whitepace separately?

@rhpvorderman Thanks, your comment made me realize I didn't pay enough attention to unifying the `command <<< >>>` processing with the new multi-line string processing, which I'll need to go back and think about, although I'm not sure I understand the finer points of your comment.

A contrived example of the difference,

```
String a = "\t\ta"
String b = "\t\tb"
String ab = """
${a}
${b}
"""
```

is `ab` indented or not? Feels to me like it should not be, but alas then I'll need to go back and change miniwdl's `command <<< >>>` handling! :sweat_smile:

rhpvorderman:
@mlin, I also think `ab` should not be indented in this case. 

What I meant is:

```WDL
workflow some_workflow {
     input {
        Boolean switch
######## This is common whitespace. 
        String ml = """"
            Dit is tekst die
            over meerdere regels gespreid
    }


}

**Comment by mlin** [comment]
Since @rhpvorderman comments above I haven't been able to shake this thought...instead of introducing two new delimiters, why not simply allowing the *existing* `<<< >>>` to work anywhere? (I know this is *very* late, but at least I took the time to implement the current draft first :sweat_smile:)

File: versions/development/SPEC.md (hunk 5 above)
Comment (jdidion) [reply versions/development/SPEC.md:372]: good catch - fixed

File: versions/development/SPEC.md (hunk 7 above)
Comment (jdidion) [reply versions/development/SPEC.md:410]: good catch, fixed

**Comment by jdidion** [comment]
@mlin yes I've added a clarification that common whitespace removal happens before interpolation of the command block. I'm open to using `<<< >>>` as the multiline string delimiter. We just tend to use python as our model, and `""" """` are the multiline string delimiters there.

**Comment by patmagee** [comment]
@jdidion @mlin I am not opposed to the idea of using the command block style for multi line strings, but it's worth mentioning that Java also uses the `""" """` style. So there is the community knowledge around that style of strings.

Another way we could have feasibly gone would have been allowing a string to follow a command:

```
command "echo foo"

command """
 echo foo
 echo bar
 echo baz
"""
```

That being said, looking at a multi line string using `<<<>>>` it really does not look that bad. And it basically makes the above example possible 
```
String description = <<<
  Hello, this is a multi line string.
  You can type across multiple lines,
  And all white space is preserved
>>>

command <<<
  Normal multi line command
>>>

command "single line command"
```

**Comment by jdidion** [comment]
I think the biggest issue is the handling of the ending delimiter is different between multi-line strings and the command block.

In multi-line strings as currently specified, if `"""` is on a line by itself, the final newline is retained (unless it's escaped) and any whitespace preceding the `"""` is considered when determining common leading whitespace.

In the command block, it is not currently specified how the last line is handled, but I am guessing that most/all execution engines strip all trailing whitespace.

If we use `<<< >>>` for multi-line strings, then I would want trailing newline and whitespace to be handled the same for multi-line strings and the command block to avoid confusion.

**Comment by rhpvorderman** [comment]
A key difference is that any trailing whitespace does not really matter in the command block as extraneous whitespace is usually ignored by the shell. For multiline strings whitespace is presumably more important. On the other hand, that just means that clearly specifying the behavior of `<<< >>>` strings will not break backwards compatibility. 

In general I think reusing the existing delimiters `<<<` `>>>` will not add increased cognitive load as it is already part of WDL. Using both `"""` and `<<<>>>` will require us to write an explanation as to why one is used in one case and the other in the other. And that explanation will not make sense I am affraid. "Why not use one type?" will be the question at the back of the user's mind. @mlin thanks for pointing this out.

**Comment by jdidion** [comment]
I am good with this. Here is a summary of how I have updated the spec:

* Multi-line strings use `<<< >>>` as the start/end tokens.
* Multi-line strings and the command block are treated exactly the same, with one exception: as the final step to evaluating a multi-line string, line continuations (i.e. a line ending with `\` immediately preceding the newline) are stripped, whereas line continuations are left in-tact (i.e. escaped) when evaluating the command block.
* All whitespace following `<<<` up to and including a newline (if any) is stripped.
* All whitespace prior to `>>>` up to and including a newline (if any) is stripped.
* A blank line can be used if it is desired to have the string/command block start/end with a new line.

```wdl
# all of these strings evaluate to "hello world"
String A = <<<hello world>>>
String B = <<<   hello world   >>>
String C = <<<   
    hello world>>>
String D = <<<   
    hello world
    >>>
String E = <<<
  hello \
  world \
>>>

# this string evaluates to "\nhello world\n"
String F = <<<

hello world

>>>

# this command block evaluates to 'echo "hello \\\nworld"
command <<<
echo "hello \
world"
>>>
```

jdidion:
My only complaint is I don't really love the way `String s = <<< ... >>>` looks. How do people feel about:

```wdl
String s <<<
  a multi
  line string
>>>
```

rhpvorderman:
> How do people feel about dropping the assignment operator?

Exceptional cases are not exceptional enough to break the rules. Assignment should always have the assignment operator, in my opinion. Sure it may look ugly, but more important than that, it will look consistent. Consistency and clarity are most important in my opinion. See also: https://www.osnews.com/story/19266/wtfsm/ ;-).

versions/development/SPEC.md
```diff
@@ -358,6 +359,102 @@ Strings can also contain the following types of escape sequences:
...
+
+```wdl
+# The following strings are equivalent. String `s2` has an empty line, which ensures that it
+# ends in a newline. Even though the first line of `s3` is indented by 6 spaces, the common 
+# leading whitespace is 2 and the string ends in a newline due to the second (blank) line
+# containing 2 spaces.
+String s1 = "    text indented by 4 spaces\n"
+String s2 = <<<
+    text indented by 4 spaces
+
```
mlin, line 441: Since this line is empty, and it doesn't factor into common leading whitespace, wouldn't the 4 spaces on the line above get removed?

versions/development/SPEC.md
```diff
@@ -1064,6 +1161,22 @@ Boolean b = true
...
 ```
 
+Placeholders are evaluated in multi-line strings exactly the same as in regular strings. Common
+leading whitespace is stripped from a multi-line string *before* placeholder expressions are evaluated.
+
+```wdl
+String spaces = "  "
+String name = "Henry"
+String company = "Acme"
+# This string evaluates to: "  Hello Henry\n  Welcome to Acme!"
```
mlin, line 1171: missing comma after Henry

mlin:
@jdidion awesome, updated draft implementation in https://github.com/chanzuckerberg/miniwdl/pull/616/files

a couple other possible clarifications between command vs non-command usage:

- within `command <<< >>>` only `~{}` and not `~{}` interpolations work, should this be true for non-command multi-line strings too?
- and should non-command multi-line strings allow the (old/deprecated?) interpolation modifiers such as `~{sep=', ' my_array}` 

Overall though, I think these differences (and the continuations which you already noted) are minor and worthwhile overall. I like not having two more sets of delimiters, and the revised treatment of whitespace on the first and last lines. Great work!

patmagee:
@mlin my thought would be that the multi line string is a great place to start phasing out those string interpolation features. We have wanted to deprecate the `${}` syntax for a long time now, and people are already used to not being able to use it in the command block.

If anything, I would want to start to move towards full deprecation/removal of those features

hkeward:
> If anything, I would want to start to move towards full deprecation/removal of those features

Very much looking forward to the day it’s fully phased out and I can use `${}` for shell variables!

versions/development/SPEC.md
jdidion: fixed

versions/development/SPEC.md
jdidion: Oh yea, good point :) So (as the spec is currently written) there is no way to force text to be indented. Are we good with this? Or should we reconsider making empty lines contribute to common leading whitespace determination?

jdidion:
I agree - only `~{}` for interpolation in multi-line strings.

jdidion:
Actually, do we want to just go through with removal of `${}` from the development spec, since it's already been deprecated in 1.1?

versions/development/SPEC.md
mlin: @jdidion I don't claim this is great, but at least there's a way

```
        <<<
        \x20 Forced
        \x20 indentation
        >>>

=> "  Forced\n  indentation"
```

rhpvorderman:
> Actually, do we want to just go through with removal of ${} from the development spec, since it's already been deprecated in 1.1?

I am fine with that. The next release should be 2.0 right? (Also removing `object`)

patmagee:
@rhpvorderman i think we are striving for a 1.2 release atm, but the next major release would be 2.0

rhpvorderman:
In that case, maybe it is best if the development spec reflects our desired state of the language while the 1.x branches represent any backwards-compatibility efforts? (I do not have a very strong opinion on this.)

versions/development/SPEC.md
jdidion: The more I think about it, the more I dislike the distinction between empty and blank lines. I am going to remove this distinction and see what people think. Can always revert.

mlin: @jdidion as I understand the revision, a blank line could have some whitespace but less than the "common leading whitespace." Is the procedure supposed to treat the blank line as if it were empty in this case?

jdidion: That's right. If a blank like has `W` whitespace characters, then min(`C`, `W`) whitespace characters are removed, where `C` is common leading whitespace.

versions/development/SPEC.md
```diff
@@ -358,6 +358,96 @@ Strings can also contain the following types of escape sequences:
...
+  multi-line string!
+>>>
+
+In multi-line strings, leading *whitespace* is removed according to the following rules. In the context of multi-line strings, whitespace refers to space (`\x20`) and tab characters only and is treated differently from newline characters.
+
+1. Remove all whitespace following the opening `<<<`, up to and including a newline (if any).
+2. Remove all whitespace preceeding the closing `>>>`, up to and including a newline (if any).
+3 Remove all line continuations and subsequent white space.
+  * A line continuation (`\`) comes at the end of a line (i.e. immediately preceding a newline) and indicates that two consecutive lines are actually the same line (e.g. when breaking a long line for better readability).
+  * Removing a line continuation means removing the `\` character, the immediately following newline, and all the whitespace preceeding the next non-whitespace character or end of line (whichever comes first).
```
mlin, line 377: Can I suggest processing the line continuations before the opening/closing whitespace? This avoids having to clarify+implement a couple of tedious corner cases, where the first or last newline is an escaped one.

Missing period after 3

versions/development/SPEC.md
```diff
@@ -358,6 +358,96 @@ Strings can also contain the following types of escape sequences:
...
+4. Use all remaining non-*blank* lines to determine the *common leading whitespace*.
+  * A blank line contains zero or more whitespace characters followed by a newline.
+  * Common leading whitespace is the minimum number of whitespace characters occuring before the first non-whitespace character in a non-blank line.
+  * Each whitespace character is counted once regardless of whether it is a space or tab (so care should be taken when mixing whitespace characters).
+5. Remove common leading whitespace from each line.
+
+```wdl
+# all of these strings evaluate to "hello world"
```
mlin, line 385: comment should have two spaces between hello & world

versions/development/SPEC.md
mlin: Another nit is that the backslash at the end of a line could itself be escaped, 

```
<<<abc\\
def>>> 
=> "abc\\\ndef"
```

I think we can say that a continued line ends in an *odd number* of backslashes.

versions/development/SPEC.md
jdidion: thanks, fixed

versions/development/SPEC.md
jdidion: Thanks, I made both those changes.

mlin:
Included this feature in [miniwdl v1.8.0](https://github.com/chanzuckerberg/miniwdl/releases/tag/v1.8.0) `version development`; I believe it's up-to-date with the spec as of this comment, but def invite anyone interested to hammer on it.

jdidion:
@mlin great! I'm personally satisfied with the spec for this feature. I'll tag people for reviews.

rhpvorderman (approved):
It looks excellent. Great work @jdidion !

geoffjentry:
LGTM

patmagee:
This looks great! Amazing effort from everyone on getting to a design that feels perfect!

patmagee (approved):

versions/development/SPEC.md
```diff
@@ -358,6 +358,104 @@ Strings can also contain the following types of escape sequences:
...
+##### Multi-line Strings
+
+Strings that begin with `<<<` and end with `>>>` may span multiple lines.
+
+```wdl
+String s = <<<
+  This is a
+  multi-line string!
+>>>
+
```
illusional, line 370: Missing closing \`\`\` backticks here.

illusional (approved):
Minor formatting issue, but otherwise 👍

versions/development/SPEC.md
jdidion: fixed, thanks!

jdidion:
Last call for comments. Plan to merge later this week.

mlin (approved):
:ship:

jdidion:
🫡

PR #428, opened 2020-12-10
Version 1.1 Spec

jdidion:
This is a draft of v1.1. I encourage and implore as many people as possible to review this document in as much detail as possible. I have tried to make it polished but I'm sure some things fell through the cracks.

My process for creating this specification was as follows:

1. Start with the current development version of the spec.
2. Merge in all accepted but unmerged PRs.
3. Manually go through all PRs to the development spec from the very beginning to make sure no content was lost due to bad merges.
4. Remove all reference to `Directory` type and `hints` section.
5. Reorganize the document into three main sections and two appendecies, and reorganize within each section to put next to each other all the related information and to reduce redundancy.
6. Review all of the text and revise where necessary to ensure consistent and correct use of terms. (Please flag any areas where my view of what is correct is wrong)
7. Perform spell checking, check for broken links, etc.

With input from the group on Slack, I have tried to correctly resolve all the conflicts I found between different parts of the spec, and to add additional clarification where needed. I was also fairly aggressive about removing wrong, redundant, or unnecessarily verbose text. Anywhere I have been too aggressive, please let me know.

I also took the liberty of adding three new pieces of content that have not previously been PR'd:

1. A section on equality and inequality of compound types. It is my understanding that this has previously been implicit, and is supported in Cromwell, so I wanted to make it explicit.
2. Serialization to JSON is undefined for Pairs and for Maps with non-String keys. This make sense, because there is no unambiguous way to represent such types in JSON. Therefore, I added suggested conversions to serializable types (and back to WDL types after deserialization). 
3. To convert a Map with a non-String key, the best conversion is to a struct with two array members (keys and values). However, there is no way to write this conversion without an unzip function. Thus, I have added an unzip function to the spec.

```wdl
struct IntStringMap {
  Array[Int] keys
  Array[String] values
}
Map[Int, String] m = {0: "a", 1: "b"}
Pair[Array[Int], Array[String]] u = unzip(m)
IntStringMap i = IntStringMap {
  keys: u.left,
  values: u.right
}
# after deserialization, we can convert back to Map
Map[Int, String] m2 = as_map(zip(i.keys, i.values))
```

I hope that items 1 and 2 are uncontroversial. For item 3, I realize we may not want to add another function to the standard library without going through the standard PR process. If that's the case, I'll remove `unzip` from this draft and I will update the serialization guide to state that serialization to/from JSON for Pairs and for Maps with non-String keys is not possible. I will then add a PR that we can review and decide whether to include in 1.1 or push it to development.


CHANGELOG.md
```diff
@@ -20,10 +20,14 @@ Keep the changelog pleasant to read in the text editor:
 
 version 2.0.0
```
geoffjentry, line 21: Should there be a `1.1.0` section in the changelog?

jdidion: Yes - I'll add it

jdidion: The question is what to put there? Do I move all the changelog items from 2.0 to 1.1? Copy them? Just say "the following items were backported from 2.0"?

geoffjentry: Since 2.0 doesn't yet technically exist - what about moving the backported items to just 1.1 and then 2.0 becomes the net-new beyond 1.1?

jdidion: Done.

rhpvorderman:
Looks really good @jdidion ! Thanks for your continued efforts!

Might I suggest renaming the changelog chapter `2.0.0` to `develop` which is the official name for using the latest spec? (Is it official yet? I believe there is a conversation somewhere where we decided on `version develop`). Using a stable version number such as `2.0.0` in the changelog might mislead people into thinking that this 2.0.0 spec is already released while in fact, it is still in development.

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5771 @@
...
+The body of the command section (i.e. the command "template") can be though of as a single string expression, which (like all string expressions) may contain placeholders.
+
+There are two different syntaxes that can be used to define command expression placeholders, depending on which style of command section definition is used:
+
+|Command Definition Style|Placeholder Style|
+|---|---|
+|`command { ... }`|`~{}` (preferred) or `${}`|
+|`command <<< >>>`|`~{}` only|
+
+Note that the `~{}` style of placeholder may only be used within the `command` section - for all other string expressions, the `${}` style must be used.
```
rhpvorderman, line 1656: Is that so? Cromwell does seem to accept `~{}` in strings outside of the command section. Or am I misremembering?
Regardless, this is an inconsistency that should be removed in 2.0 -> everything should be `~{}`.

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5771 @@
...
+    # the "write_lines" function writes each string in an array
+    # as a line in a temporary file, and returns the path to that
+    # file, which can then be referenced by other commands such as 
+    # the unix "cat" command
+    cat write_lines(str_array)
+  >>>
+}
+```
+
+In most cases, the `~{}` style of placeholder is preferred, to avoid ambiguity between WDL placeholders and Bash variables, which are of the form `$name` or `${name}`. If the `command { ... }` style is used, then `${name}` is always interpreted as a WDL placeholder, so care must be taken to only use `$name` style Bash variables.
```
rhpvorderman, line 1691: At the end of the paragraph "Or use `command <<< ... >>>` style". It is good to make that explicit.

CHANGELOG.md
```diff
@@ -20,13 +20,53 @@ Keep the changelog pleasant to read in the text editor:
...
 ---------------------------
+
++ Runtime section no longer accepts arbitrary keys. Added new hints section for arbitrary runtime hints.
```
rhpvorderman, line 24: Nitpick: The changelog style guide mentions 80 characters per line max. But I am biased since I wrote that style guide :wink:. This is up for debate of course.

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5771 @@
...
+task single_image_test {
+  #....
+  runtime {
+    container: "ubuntu:latest"
+  }
+```
+
+Container source locations should use the syntax defined by the individual container repository. For example an image defined as `ubuntu:latest` would conventionally refer a docker image living on `DockerHub`, while an image defined as `quay.io/bitnami/python` would refer to a `quay.io` repository.
+
+The `container` key also accepts an array of URI strings. All of the locations must point to images that are equivalent, i.e. they must always produce the same final results when the task is run with the same inputs. It is the responsibility of the execution engine to define the specific image sources it supports, and to determine which image is the "best" one to use at runtime. The ordering of the array does not imply any implicit preference or ordering of the containers. All images are expected to be the same, and therefore any choice would be equally valid. Defining multiple images enables greater portability across a broad range of execution environments.
```
rhpvorderman, line 1922: I'd prefer the execution engine here to simply try them in order and use the first one that works:

+ Simpler logic in the execution engine
+ Predictable behaviour.
+ Simpler spec

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5771 @@
...
+            -k ${min_seed_length} \
+            -I ${sep=',' min_std_max_min} \
+            ${reference} \
+            ${sep=' ' reads+} > output.sam
+  }
+  output {
+    File sam = "output.sam"
+  }
+  runtime {
+    container: "my_image:latest"
```
rhpvorderman, line 2341: `myimage:latest` -> `quay.io/biocontainers/bwa:0.7.17--hed695b0_7`
This will make the example fully working.

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5771 @@
...
+```wdl
+floor(1.0) == 1
+floor(1.9) == 1
+ceil(2.0) == 2
+ceil(2.1) == 3
+round(1.49) == 1
+round(1.50) == 2
+```
+
+## ✨ Int min(Int, Int), Float min(Int|Float, Int|Float)
```
rhpvorderman, line 3381: `Int min(Int, Int), Float min(Float, Int), Float min(Int, Float), Float min(Float, Float)` is more correct. The current notation leaves room for `Float min(Int, Int)` which is incorrect. In this case I think spelling it out adds to clarity.

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5771 @@
...
+  }
+  output {
+    # these two expressions are equivalent
+    Float min1 = if value1 < value2 then value1 else value2
+    Float min2 = min(value1, value2)
+  }
+}
+``` 
+
+## ✨ Int max(Int, Int), and Float max(Int|Float, Int|Float)
```
rhpvorderman, line 3408: Same comment as for min. This can be spelled out as `Int max(Int, Int), Float max(Float, int) , Float max(Int, Float) and Float max(Float, Float)`

rhpvorderman (commented):
I found a few nitpicks when reading trough the entire spec.

A non-content related comment: I noticed that many lines span more than 80-100 characters. While we can debate about the exact character length, I think there is the more pressing issue of lines that contain multiple sentences, even entire paragraphs. This transforms small typo edits in huge differences when viewed with `git diff`. So I recommend splitting some lengthy paragraphs across multiple lines.

versions/1.1/SPEC.md
rhpvorderman: Non-content comment: This is one of the examples of way to many sentences on one line. Maybe use one line per sentence here.

CHANGELOG.md
```diff
@@ -75,19 +111,13 @@ version 2.0.0
...
-+ Added a new directory type to make it easier when working with inputs that
-  consist of multiple files.
-  [PR 241](https://github.com/openwdl/wdl/pull/241) by @cjllanwarne.
-
 + Several bugs in the grammar have been fixed.
   [PR 238](https://github.com/openwdl/wdl/pull/238) and
   [PR 240](https://github.com/openwdl/wdl/pull/240) by @cjllanwarne.
 
-<!---
 This is not implemented yet.
```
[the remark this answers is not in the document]
jdidion: That was a mistake

jdidion:
I have seen `development` being used in customer WDLs. I think this is what people have been using because it's not stated specifically in the spec.

versions/1.1/ImplementationGuide.md
```diff
@@ -0,0 +1,45 @@
...
+6. Input localization: Each `File` typed value - including compound values with nested `File`s - that references a non-local resource is localized, which involves:
+    * Creating a local file (or file-like object, such as a `fifo`), according to the [input localization](#task-input-localization) rules.
+    * Making the contents of the remote resource readable from that file, either by downloading the remote resource or by streaming its contents.
+    * Replacing the value of the `File` typed declaration with the path to the file.
+7. Command instantiation:
+    * The command template is evaluated in the context of all the (localized) input and private declaration values, and all expression placeholders are replaced with their (stringified) values. 
+    * The instantiated command is written to local disk and made executable.
+8. Container resolution:
+    * If alternative container images are specified, the "best" one is selected
```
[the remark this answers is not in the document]
jdidion: The spec specifically states that all images should be equivalent, and it's up to the engine to pick which one is "best". I can put a link to that in the implementation guide.

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5771 @@
+# Workflow Description Language
+
+This is version 1.1 of the WDL specification. It introduces a number of ✨ new features and clarifications to the 1.0 version of the specification. It also 🗑 deprecates several aspects of the 1.0 specification that will be removed in version 2.0.
```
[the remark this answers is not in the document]
jdidion: Yes - that's the way I originally had it, but I don't like the way GitHub represents an emoji next to a parenthesis. Maybe I'll say "(signified by the ✨ symbol)".

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5771 @@
...
+$float = (([0-9]+)?\.([0-9]+)|[0-9]+\.|[0-9]+)([eE][-+]?[0-9]+)?
+$string = "([^\\\"\n]|\\[\\"\'nrbtfav\?]|\\[0-7]{1,3}|\\x[0-9a-fA-F]+|\\[uU]([0-9a-fA-F]{4})([0-9a-fA-F]{4})?)*"
+$string = '([^\\\'\n]|\\[\\"\'nrbtfav\?]|\\[0-7]{1,3}|\\x[0-9a-fA-F]+|\\[uU]([0-9a-fA-F]{4})([0-9a-fA-F]{4})?)*'
+```
+
+Tasks and workflow inputs may be passed in from an external source, or they may be specified in the WDL document itself using literal values. Input, output, and other declaration values may also be constructed at runtime using [expressions](#expressions) that consist of literals, identifiers (references to declarations or call outputs), built-in [operators](#operator-precedence-table), and [standard library functions](#standard-library).
+
+#### Strings
+
+A string literal may contain any characters between single or double-quotes, with the exception of a few special characters that must be escaped:
```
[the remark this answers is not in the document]
jdidion: done

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5771 @@
...
+
+|Escape Sequence|Meaning|\x Equivalent|
```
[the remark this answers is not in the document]
jdidion: Good point - I'll add a "Notes" column to the table with this information.

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5771 @@
...
+- It's impossible for a task to change an upstream value which has been provided as an input: even if it makes changes to its local copy the original value is unaffected.
+
+#### Primitive Types
+
+The following primitive types exist in WDL:
+
+* A `Boolean` represents a value of `true` or `false`.
+* An `Int` represents a signed integer in the range \[-2^63, 2^63).
+* A `Float` represents a finite 64-bit IEEE-754 floating point number.
+* A `String` represents a character string following the format described [above](#strings).
```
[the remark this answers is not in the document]
jdidion: done

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5771 @@
...
+* A `File` represents a file (or file-like object).
+  * A `File` declaration can have a string value indicating a relative or absolute path on the local file system.
+  * Execution engines may support other ways to specify `File` inputs (e.g. as URIs); the execution engine must [localize inputs](#task-input-localization) so that the runtime value of a `File` variable is a local path.
```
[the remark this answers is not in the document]
jdidion: Here it is specifically meant that engines may accept inputs in whatever format they want (as long as they also support the official format). Within a WDL I agree that `File`s should only be local string paths. I'll add a clarification.

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5771 @@
...
+}
+```
+
+## WDL Documents
+
+A WDL document is a file that contains valid WDL definitions.
+
+A WDL document must contain:
+
+* A [version statement](#versioning)
```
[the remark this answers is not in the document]
jdidion: done

versions/1.1/SPEC.md
jdidion: Interesting. I'll ask about this on Slack and update the spec if this is indeed the case.

jdidion: I think you're right. I'm not sure if `${}` is actually meant to be deprecated, but it seems clear that both formats should be supported.

versions/1.1/SPEC.md
jdidion: done

CHANGELOG.md
jdidion: I think this is the same wrap-vs-no-wrap argument. We should decide on one or the other and be consistent.

versions/1.1/SPEC.md
jdidion: Right, but that is not what is stated in @patmagee PR #315. The discussion should be had there.

versions/1.1/SPEC.md
jdidion: done

versions/1.1/SPEC.md
jdidion: You're right, I was just trying to shorten it, but I'll spell it out.

versions/1.1/SPEC.md
jdidion: done

versions/1.1/SPEC.md
jdidion: From discussions on Slack, it sounds like `${}` should be a soft deprecation - discourage its use, update examples to use `~{}`, and warn that it may be (hard) deprecated in a future version.

versions/1.1/SPEC.md
patmagee: My answer from slack
> my intention was not to give an equivalent of CSS font selection (ie fallback options if the main container is not available). Containers should all be the exact same (ie same hash) and the engine can choose whatever is most optimal

The reasoning for this was that certain container repos would be supported by an engine while others were not. So in this scenario the engine can choose whichever container it likes, and still is able to run the task in the exact same way. 

Take the following example:
ie: 
```wdl
runtime {
  container: ["foo/image:latest", "bar/image:latest", "bizz/image:lateast"]
}
```

Does not imply the precedence relationship of `"foo/image:latest" > "bar/image:latest" > "bizz/image:latest"` but the equality relationship of `"foo/image:latest" == "bar/image:latest" == "bizz/image:latest"` (which ideally would be provable by the engine, but is unlikely if different engines simply cannot read for a different repository)

patmagee:
@rhpvorderman thanks for the extremely thorough review!!

Most tooling that existing for WDL uses the `version development` and not develop to reference the current WIP release. I would not suggest we deviate from this since that would likely cause a world of issues everywhere. Additionally, the version is named `development` in the folder hierarchy: https://github.com/openwdl/wdl/tree/main/versions/development

When It comes to wrapping, I am generally a :-1:  on manual wrapping. Most modern editors handle wrapping splendidly so there is no issue there. When it comes to reading changes in large paragraphs, I have found that GH does a pretty good job of highlighting the actual changes in the text for you. I am actually a fine of showing the whole paragraph as a change though. When it comes to text context matters, and the only way you can evaluate a change contextually is to read the entire paragraph that it is a part of

rhpvorderman:
Ah yes, sorry. I tried to signal that the top version in the changelog should always be 'development' since that was what we decided on not '2.0.0' or some other version. I forgot that it was actually 'development' and not 'develop'. Sorry!

About the wrapping I am convinced now and I will drop the issue :slightly_smiling_face: .

versions/1.1/SPEC.md
rhpvorderman: Ah okay. Well since this was already discussed elsewhere so no need for me to challenge this decision. If this solves problems for people I am happy it is there.

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5775 @@
...
+  "wf.infile": "/file.txt"
+}
+```
+
+Running the `wf` workflow with these inputs would yield the following command line from the call to the `hello` task:
+
+```sh
+egrep '^[a-z]+$' '/file.txt'
+```
+
```
vortexing, line 253: A quick subheading here would help 
### Parallelizing Hello World

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5775 @@
...
+The inputs to this workflow would be `wf_parallel.files` and `wf_parallel.pattern`. For example:
+
+```json
+{
+  "wf_parallel.pattern": "^[a-z]+$",
+  "wf_parallel.infile": ["/file1.txt", "/file2.txt"]
+}
+```
+
```
vortexing, line 288: Before moving on, it seems like I'm not sure what the goal of this first section was. It's too specific to just be a general orientation to the reader to what WDL is supposed to be.  Perhaps include here a link to other resources for finding example workflows, that is out of scope for this document, then move on to defining the spec?  Such as: https://github.com/openwdl/learn-wdl

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5775 @@
...
+    - [Task Scope](#task-scope)
+    - [Workflow Scope](#workflow-scope)
+    - [Cyclic References](#cyclic-references)
+    - [Namespaces without Scope](#namespaces-without-scope)
+  - [Evaluation Order](#evaluation-order)
+
+## Introduction
+
+WDL is meant to be a *human readable and writable* way to express tasks and workflows. Here is the "Hello World" tool in WDL:
+
```
vortexing, line 188: Does someone have an actual diagram of the essential parts of a WDL workflow that would be useful here to orient someone new to the spec?  Starting off with an example written IN WDL assumes that means something to the person reading it.  But if they're new to the spec and reading the intro still, they likely aren't.  No masterpiece required, but surely someone has a handy portion of a slide they've used in presentations or something?

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5775 @@
...
+  #You can have comments anywhere in the workflow
+  call test
+  
+  output { #You can also put comments after braces
+    String result = test.result
+  }
+}
+```
+
+### Keywords
```
vortexing, line 389: Perhaps consider making this "Reserved Keywords" instead of just "Keywords"?  Seems more clear even if you scan the text that you should pay attention to this section even if you otherwise wouldn't.

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5775 @@
...
+* `Directory`
+* `hints`
+
+### Types
+
+A [declaration](#declarations) is a name that the user reserves in a given [scope](#appendix-b-wdl-namespaces-and-scopes) to hold a value of a certain type. In WDL *all* declarations (including inputs and outputs) must be typed. This means that the information about the type of data that may be held by each declarations must be specified explicitly.
+
+In WDL *all* types represent immutable values.
+
+- Even a `File` represent a logical "snapshot" of the file at the time when the value was created.
```
vortexing, line 412: Typo?  this isn't a sentence in my mind:  Even a `File` represent a logical "snapshot" of the file at the time when the value was created.` >>Perhaps change to: >> A `File` represent**s** a logical "snapshot" of the file at the time when the value was created.

vortexing: Also, why are these two sentences bulleted and not just part of the paragraphs above? Are they for some reason different?

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5775 @@
...
+
+##### Array[X]
+
+An `Array` represents an ordered list of elements that are all of the same type.
+
+An array value can be declared with comma-separated values in brackets (`[]`), and a specific (zero-based) index of an `Array` can be accessed by placing the index in brackets after the declaration name. Accessing a non-existent index of an `Array` results in an error.
+
+```wdl
+Array[File] files = ["/path/to/file1", "/path/to/file2"]
+File f = files[0]
```
vortexing, line 450: Add commented example here, like you do for `empty[0]`:
File f = files[0] # evaluates to: "/path/to/file1"

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5775 @@
...
+
+Array[Int] empty = []
+empty[0]  # error!
```
vortexing, line 453: To be more consistent, would this little example be:
`Int I = empty[0] # error!`

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5775 @@
...
+```
+
+An `Array` may be empty, unless it is declared using the non-empty postfix quantifier `+`, in which case it must contain at least one element. Note that the `+` and `?` postfix quantifiers can be combined to declare an array that is either undefined or non-empty. Attempting to assign an empty array literal to a non-empty `Array` declaration results in an error. Otherwise, the non-empty assertion is only checked at runtime: binding an empty array to an `Array[T]+` input or function argument is a runtime error.
```
vortexing, line 456: Is there a more concrete example of when you'd want to declare an undefined array?  Or a non-empty one?  Like a one-liner explaining this?

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5775 @@
...
+Array[Int]+? nonempty4
+Array[Int]+? nonempty5 = [0.0]
+
+Array[Boolean]+ nonempty3 = []  # error!
+Array[Int]+? nonempty6 = []     # error!
+```
+
+##### Pair[X, Y]
+
+A `Pair` represents two associated values, which may be of different types In other programming languages, a `Pair` might be called a "two-tuple".
```
vortexing, line 475: Typo (lowercase I in In).  OR  perhaps this is better:

`A `Pair` represents two associated values, which may be of different types.  In other languages, a `Pair` might be called a "two-tuple".`

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5775 @@
...
+
+A `Pair` can be declared with comma-separated values in parentheses (`()`). The components of a pair value are accessed using its `left` and `right` accessors.
+
+```wdl
+Pair[Int, Array[String]] data = (5, ["hello", "goodbye"])
+Int i = p.left
```
vortexing, line 481: Provide quick comments:
Int i = p.left    # evaluates to the integer 5
String hello = data.right[0]  # evaluates to the string "goodbye"

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5775 @@
...
+Array[Float]+ nonempty = [0.0]
+# array that must contain at least one Int?
+# (which may have an undefined value)
+Array[Int?]+ nonempty2 = [None, 1]
+# array that can be undefined or must contain
+# at least one Int
+Array[Int]+? nonempty4
+Array[Int]+? nonempty5 = [0.0]
+
+Array[Boolean]+ nonempty3 = []  # error!
```
vortexing, line 469: All of these commented examples of what the value WOULD be would all be more clear to a wider audience if you wrote them all the same way:  So `# error!` and ` # => error!` (seen later on in the doc) throughout the document would be replaced with `# evaluates to error!`

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5775 @@
...
+Map[File, Array[Int]] file_to_ints = {
+  "/path/to/file1": [0, 1, 2],
+  "/path/to/file2": [9, 8, 7]
+}
+
+string_to_int["b"]  # => 2
+string_to_int["c"]  # error!  
+```
+
+A `Map` is insertion ordered, meaning the order in which elements are added to the `Map` is preserved, for example when [✨ converting a `Map` to an array of `Pair`s](#-arraypairp-y-as_pairsmapp-y).
```
vortexing, line 503: Is this the place to add that if you use something to add to a map, it'll be added at the end?   Shouldn't the same concern by highlighted for `Array` too?  It works the same way right?  Nothing was mentioned regarding insertion order for Arrays.

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5775 @@
...
+
+```wdl
```
vortexing, line 505: These types of examples might be more clear as well if you did something like this:
```wdl
## Variable definition:
Map[Int, String] int_to_string = { 2: "hello", 1: "goodbye" }

## Variable evaluation at runtime:
as_pairs(int_to_string)  # => [(2, "hello"), (1, "goodbye")]
```
Because technically you wouldn't just WRITE the evaluation statement in the WDL in the same section as the variable definition.  Might help folks understand the timing of when to write what sort of format.  Also you haven't introduced what `as_pairs` actually is, b/c the library section hasn't been read yet. ;) I think that's sort of fine, but call out that that's a runtime thing.

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5775 @@
...
+Object f = object {
+  a: 10,
+  b: 11
+}
+Int i = f.a
+```
+
+Due to the lack of explicitness in the typing of `Object` being at odds with the goal of being able to know the type information of all WDL declarations, **the `Object` type, the `object` literal syntax, and all of the [standard library functions](#-object-read_objectstringfile) with `Object` parameters or return values have been deprecated and will be removed in the next major version of the WDL specification**.
```
vortexing, line 529: Might be nice here to make a suggestion for folks: 
"Objects can be better defined by using custom types (Structs, see next section), or more clearly typed declarations such as Array, Pair, and Map when appropriate."

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5775 @@
...
+
+##### Custom Types (Structs)
+
+WDL provides the ability to define custom compound types called [`struct`s](#struct-definition). Structs are defined directly in the WDL document and are usable like any other type. 
+
+A `struct` is declared using the `struct` keyword, followed by a unique name, followed by member declarations within braces. A declaration with a custom type can be assigned a value using struct literal syntax, which is identical to the [objects literal syntax](#-object) with the exception that the `struct` name is used instead of the `object` keyword.
```
vortexing, line 535: Do you want to be referring here to the objects syntax given that it will be deprecated?  You'll have to fix this later.  Perhaps now is a good time to put whatever text you were hoping the readers to read here, HERE, even if it's slightly duplicated in the short term.

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5775 @@
...
+  b: 11
+}
+Int i = f.a
+```
+
+Due to the lack of explicitness in the typing of `Object` being at odds with the goal of being able to know the type information of all WDL declarations, **the `Object` type, the `object` literal syntax, and all of the [standard library functions](#-object-read_objectstringfile) with `Object` parameters or return values have been deprecated and will be removed in the next major version of the WDL specification**.
+
+##### Custom Types (Structs)
+
+WDL provides the ability to define custom compound types called [`struct`s](#struct-definition). Structs are defined directly in the WDL document and are usable like any other type.
```
vortexing, line 533: WDL provides the ability to define custom compound types called [`struct`s](#struct-definition). `Struct`s are defined directly in the WDL document and are usable like any other type.   `Struct`s can be made up of any type or compound types and are defined by a keyword before the opening brace (`{`), much like a `map` declaration.

The value of a specific member of an `struct` value can be accessed by placing a `.` followed by the member identifier after the `struct` keyword/declaration name.

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5775 @@
...
+
+##### Custom Types (Structs)
+
+WDL provides the ability to define custom compound types called [`struct`s](#struct-definition). Structs are defined directly in the WDL document and are usable like any other type. 
+
+A `struct` is declared using the `struct` keyword, followed by a unique name, followed by member declarations within braces. A declaration with a custom type can be assigned a value using struct literal syntax, which is identical to the [objects literal syntax](#-object) with the exception that the `struct` name is used instead of the `object` keyword.
+
+```wdl
+# a struct is a user-defined type; it can contain members of any types
+struct BamAndIndex {
```
vortexing, line 539: Perhaps give an example here that couldn't realistically be written as a Pair (in this case) that would make it more clear WHY you'd bother defining the struct vs just using some existing compound type.  
```struct BamAndIndexAndParams {
    File bam
    File bam_index
    Array[Int] params_to_test
}```

File: versions/1.1/SPEC.md [hunk 34]
```diff
@@ -0,0 +1,5775 @@
...
+
+BamAndIndex b_and_i = BamAndIndex { 
+  bam: "NA12878.bam",
+  bam_index: "NA12878.bam.bai" 
+}
+```
+
+#### Optional Types and None
+
+A type may have a `?` postfix quantifier, which means that it is an optional type, and its value is allowed to be undefined. It can only be used in calls or functions that accept optional values.
```
vortexing, line 552: To me this sounds like you mean that the type is optional for the variable (like, I think `File? foo` should be a file, but  if foo was an Int, that'd be ok too).  Perhaps what you want here is something more like: 

The type of a variable may have a `?` postfix quantifier, which means that its value is allowed to be undefined without throwing an error at runtime. It can only be used in calls or functions that accept optional values.

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5775 @@
...
+
+|Target Type |Source Type     |Notes/Constraints |
+|------------|----------------|------------------|
+|`File`|`String`||
+|`Float`|`Int`|May cause overflow error|
+|`Y?`|`X`|`X` must be coercible to `Y`|
+|`Array[Y]`|`Array[X]`|`X` must be coercible to `Y`|
+|`Map[X,Z]`|`Map[W,Y]`|`W` must be coercible to `X` and `Y` must be coercible to `Z`|
+|`Pair[X,Z]`|`Pair[W,Y]`|`W` must be coercible to `X` and `Y` must be coercible to `Z`|
+|`Object`|`Map[String,Y]`||
```
vortexing, line 601: Do you want to separate the `Object` coersions to the bottom and put some *** in there to highlight that `object` coersions will be irrelevant once deprecated?

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5775 @@
...
+There are some pairs of WDL types for which there is an obvious, unambiguous conversion from one to the other. In these cases, WDL provides an automatic conversion (called "coercion") from one type to the other, such that a value of one type can be used anywhere the other type is expected.
+
+For example, file paths are always represented as strings, making the conversion from `String` to `File` obvious and unambiguous.
+
+```wdl
+String path = "/path/to/file"
+File f = path
+```
+
+The table below lists all globally valid coercions. The "target" type is the type being coerced to (this is often called the "left-hand side" or "LHS" of the coercion) and the "source" type is the type being coerced from (the "right-hand side" or "RHS").
```
vortexing, line 591: Move line 608 to here.  

"The table below lists all globally valid coercions, beyond the coercion of all primitive types which can always be converted to `String` using [string interpolation](#expression-placeholders-and-string-interpolation). The "target" type is the type being coerced to (this is often called the "left-hand side" or "LHS" of the coercion) and the "source" type is the type being coerced from (the "right-hand side" or "RHS")."

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5775 @@
...
+|`Struct`    |`Map[String,Y]`|`Map` keys must match `Struct` member names, and all `Struct` members types must be coercible from `Y`|
+|`Map[String,Y]`|`Struct`|All `Struct` members must be coercible to `Y`|
+|`Object`|`Struct`||
+|`Struct`|`Object`|`Object` keys must match `Struct` member names, and `Object` values must be coercible to `Struct` member types|
+
+Note that primitive types can always be converted to `String` using [string interpolation](#expression-placeholders-and-string-interpolation).
+
+```wdl
+Int i = 5
+String istring = "~{i}"
```
vortexing, line 612: This is the first time you've used `~{}`.  Just FYI.  It's an example that may not add much value as written in part b/c the reader doesn't technically know what this means yet, but I have a hard time giving you a better suggestion.

versions/1.1/SPEC.md
jdidion: Thanks - I reorganized the intro a bit to hopefully make it clearer that it's just meant as a quick example of what you can do with WDL.

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5775 @@
...
+
+This constraint propagates into compound types. For example, an `Array[T?]` can contain both optional and non-optional elements. This facilitates the common idiom `select_first([expr, default])`, where `expr` is of type `T?` and `default` is of type `T`, for converting an optional type to a non-optional type. However, an `Array[T?]` could not be passed to the [`sep`](#-string-sepstring-arraystring) function, which requires an `Array[T]`.
+
+The two exceptions where coercion from `T?` to `T` are allowed are:
+* [String concatenation in expression placeholders](#concatenation-of-optional-values)
+* [Equality and inequality comparisons](#equality-and-inequality-comparison-of-optional-types)
+
+###### Struct/Object coercion from Map
+
+`Struct`s and `Object`s can be coerced from map literals, but beware the following behavioral difference:
```
vortexing, line 629: There's no example here using `Struct` and given that `Object` will be deprecated, seems odd to do this via Object examples.

versions/1.1/SPEC.md
jdidion: Thanks, I didn't know about that repo! Please read the revised version and see if it makes more sense.

versions/1.1/SPEC.md
jdidion: That's a good suggestion - I don't have anything but maybe @vdauwera does?

versions/1.1/SPEC.md
jdidion: done

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5775 @@
...
+```
+
+A fully qualified name is the unique identifier of any particular call, input, or output, and has the following structure:
+
+* For calls: `<parent namespace>.<call alias>`
+* For inputs and outputs: `<parent namespace>.<input or output name>`
+
+The `parent namespace` here will equal the fully qualified name of the workflow containing the call, or the workflow or task containing the input or output. For the top-level workflow this is equal to the workflow name.
+
+For example:
```
vortexing, line 697: Preface this with a link to the text about workflow imports first.

versions/1.1/SPEC.md
jdidion: good suggestion

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5775 @@
...
+
+Here, `ns.ns2.task` is a namespace identifier (see the [Call Statement](#call-statement) section for more details). Namespace identifiers, like fully-qualified names are left-associative, which means `ns.ns2.task` is interpreted as `((ns.ns2).task)`, which means `ns.ns2` has to resolve to a namespace so that `.task` can be applied. If `ns2` is instead a task definition within `ns`, then the namespaced identifier `ns.ns2.task` is invalid.
+
+### Declarations
+
+```txt
+$declaration = $type $identifier ('=' $expression)?
+```
+
+A declaration reserves a name that can be referenced anywhere in the [scope](#appendix-b-wdl-namespaces-and-scopes) where it is declared. A declaration has a type, a name, and an optional initialization. Each declaration must be unique within its scope, and may not collide with a reserved WDL keyword (e.g. `workflow`, or `input`).
```
vortexing, line 807: maybe add a link here to the reserved keyword section?

versions/1.1/SPEC.md
jdidion: Yes, that's true. Generally I've found the examples to be a bit looser on correct syntax when it is just statements shown outside the context of a task or workflow, but they should probably still be valid WDL.

versions/1.1/SPEC.md
jdidion: done

versions/1.1/SPEC.md
jdidion: I tried to make this more clear. Please read the revised version and see if it make sense.

versions/1.1/SPEC.md
jdidion: done

versions/1.1/SPEC.md
jdidion: done

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5775 @@
...
+import "other.wdl" as ns
+
+workflow wf {
+  call ns.ns2.task
+}
+```
+
+Here, `ns.ns2.task` is a namespace identifier (see the [Call Statement](#call-statement) section for more details). Namespace identifiers, like fully-qualified names are left-associative, which means `ns.ns2.task` is interpreted as `((ns.ns2).task)`, which means `ns.ns2` has to resolve to a namespace so that `.task` can be applied. If `ns2` is instead a task definition within `ns`, then the namespaced identifier `ns.ns2.task` is invalid.
+
+### Declarations
```
vortexing, line 801: I wish this section could be much higher in the text.  I think the rest of the above text would make more sense if you'd defined what a declaration is, and what does them (tasks and workflows).  That's a big structural issue though, so that might be overwhelming and not worth even more of your effort.

versions/1.1/SPEC.md
jdidion: I think that's a bit confusing - it doesn't evaluate to `error!` - rather the evaluation fails with an error.

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5775 @@
...
+task mytask {
+  input {
+    String s  # required input declaration
+    Int? i    # optional input declaration
+    Float f   # input delcaration with an initialization
+  }
+
+  Boolean b = true  # declaration in the task body
+
+  command <<< ... >>>
```
vortexing, line 831: This <<< ... >>> format is the first mention of it.  In fact, it's the first mention of the command section at all.  perhaps start with the command {...} format and then have the <<<...>>> as an optional one?

versions/1.1/SPEC.md
jdidion: Sure I can add that. Generally it's assumed that Arrays are ordered (otherwise they'd be Sets), but that might only be obvious to someone coming from a CS background.

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5775 @@
...
+|--------|--------|--------|------|
+|`Array`|`==`|`Array`|`Boolean`|
+|`Array`|`!=`|`Array`|`Boolean`|
+|`Map`|`==`|`Map`|`Boolean`|
+|`Map`|`!=`|`Map`|`Boolean`|
+|`Pair`|`==`|`Pair`|`Boolean`|
+|`Pair`|`!=`|`Pair`|`Boolean`|
+|`struct`|`==`|`struct`|`Boolean`|
+|`struct`|`!=`|`struct`|`Boolean`|
+|`Object`|`==`|`Object`|`Boolean`|
```
vortexing, line 1022: Perhaps add *** to be deprecated here?

versions/1.1/SPEC.md
jdidion: Right - these need to be cleaned up throughout the whole spec. I'll work on that.

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5775 @@
...
+|            | Greater Than          | left-to-right | x>y                  |
+|            | Greater Than Or Equal | left-to-right | x>=y                 |
+| 3          | Equality              | left-to-right | x==y                 |
+|            | Inequality            | left-to-right | x!=y                 |
+| 2          | Logical AND           | left-to-right | x&&y                 |
+| 1          | Logical OR            | left-to-right | x\|\|y               |
+
+#### Member Access
+
+The syntax `x.y` refers to member access. `x` must be a call in a workflow, or a struct or object value. A call can be thought of as a struct where the members are the outputs of the called task.
```
vortexing, line 1092: Slight formatting/caveat change:
"The syntax `x.y` refers to member access. `x` must be either a call in a workflow, a `struct` or an `object` (until it is deprecated). A call can be thought of as a `struct` where the members are the outputs of the called task."

versions/1.1/SPEC.md
vortexing: Yeah, the `evaluates to` makes more sense for these other examples I'm finding in the text.  For errors, perhaps write: `# Throws an error at runtime` or something to that effect.  Just use more words and make them the SAME words throughout the doc.

versions/1.1/SPEC.md
vortexing: ASS-u-me.  You ARE writing a spec here, and it's, what, like 1 sentence of additional length?  hahahah.  ;)

versions/1.1/SPEC.md
vortexing: Maybe throw in a trash can icon?

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5804 @@
...
+
+Struct z
+String a = z.b
+```
+
+#### Ternary operator (if-then-else)
+
+This is an operator that takes three arguments: a condition expression, an if-true expression, and an if-false expression. The condition is always evaluated. If the condition is true then the if-true value is evaluated and returned. If the condition is false, the if-false expression is evaluated and returned. The if-true and if-false expressions must return values of the same type, such that the value of the if-then-else is the same regardless of which side is evaluated.
+
+Examples:
```
vortexing, line 1106: These examples are not feeling useful to me.  They need more context to really understand what they are trying to tell the reader.  I don't think you've introduced the runtime section yet.  Also, making your examples closer to valid WDL would be good.  The `Boolean morning = ...` especially - what are you intending here?  What actual example could make this more clear?  Maybe put 
```
# in input to workflow, the value of morning is set to either true or false
Boolean morning = true # or false
String greeting = "good " + if morning then "morning" else "afternoon"
# Runtime evaluation if morning is true
greeting = "good morning"
# Runtime evaluation if morning is false
greeting = "good afternoon"
```

OR perhaps something else.

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5804 @@
...
+* `' ` - literal string
+* `~{input}` - identifier expression - replaced with the value of `input`
+
+Placeholders may contain other placeholders to any level of nesting, and placeholders are evaluated recursively in a depth-first manner. Placeholder expressions are anonymous - they have no name and thus cannot be referenced by other expressions, but they can reference declarations and call outputs.
+
+```wdl
+Int i = 3
+Boolean b = true
+# s evaluates to "4", but would be "0" if b were false
+String s = "~{if b then '${1 + 3}' else 0}"
```
vortexing, line 1157: Did you mean this:
```
Int i = 3
Boolean b = true
# s evaluates to "4", but would be "0" if b were false
String s = "~{if b then '${1 + i}' else 0}"
```

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5804 @@
...
+```
+
+##### Expression Placeholder Coercion
+
+The expression result in a placeholder must ultimately be converted to a string in order to take the place of the placeholder in the command script. This is immediately possible for WDL primitive types due to automatic conversions ("coercions") that occur only within the context of string interpolation:
+
+- `String` is substituted directly.
+- `File` is substituted as if it were a `String`.
+- `Int` is formatted without leading zeros (unless the value is `0`), and with a leading `-` if the value is negative.
+- `Float` is printed in the style `[-]ddd.ddd`, with 6 digits after the decimal point.
```
vortexing, line 1167: Excluding the decimal point?  or did you mean `[-]ddd.dddddd`?

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5804 @@
...
+  - The first file is always placed according to the rules above.
+  - Subsequent files that would otherwise overwrite this file are instead placed in a subdirectory named for the version.
+
+For example, imagine two versions of file `fs://path/to/A.txt` are being localized (labeled version `1.0` and `1.1`). The first might be localized as `/execution_dir/path/to/A.txt`. The second must then be placed in `/execution_dir/path/to/1.1/A.txt`
+
+#### Input Type Constraints
+
+Recall that a type may have a quantifier:
+* `?` means that the parameter is optional. A user does not need to specify a value for the parameter in order to satisfy all the inputs to the workflow.
+* `+` applies only to `Array` types and it represents a constraint that the `Array` value must contain one-or-more elements.
```
vortexing, line 1516: This is far more clear/succinct than I'd seen it before.... or wait, did this ever get mentioned before?  These two bullets need to be in multiple places perhaps.  Or not here, in the other places where they are mentioned.

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5804 @@
...
+```
+
+### Command Section
+
+The `command` section is the only required task section. It defines the command template that is evaluated and executed when the task is called. Specifically, the commands are executed after all of the inputs are staged and before the outputs are evaluated.
+
+There are two different syntaxes that can be used to define the command section:
+
+```wdl
+# HEREDOC style - this way is preferred
```
vortexing, line 1667: Oh, I stand corrected.  Then I would go back through and make sure all example command sections have HEREDOC styling with this exception here.

versions/1.1/SPEC.md
jdidion: done

versions/1.1/SPEC.md
jdidion: done

versions/1.1/SPEC.md
jdidion: Yes, given that the object literal syntax has already been discussed. In the 2.0 spec when objects have been removed, we'll also remove references to objects and object literal syntax throughout the spec.

jdidion: Actually, this is a good point. I think it makes better sense to talk about custom types first and objects second.

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5846 @@
...
+#### Expression Placeholders
+
+The body of the command section (i.e. the command "template") can be though of as a single string expression, which (like all string expressions) may contain placeholders.
+
+There are two different syntaxes that can be used to define command expression placeholders, depending on which style of command section definition is used:
+
+|Command Definition Style|Placeholder Style|
+|---|---|
+|`command { ... }`|`~{}` (preferred) or `${}`|
+|`command <<< >>>`|`~{}` only|
```
vortexing, line 1711: minutia:  put the <<< >>> option first so it looks more like it's the preferred option for the format.  Then maybe put a caveat for the  {  } format to remind folks that when that is used you might run into bash variable problems.

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5846 @@
...
+A common pattern is to use a placeholder in a string expression to construct a file name as a function of the task input. For example:
+
+```wdl
+task example {
+  input {
+    String prefix
+    File bam
+  }
+
+  command {
```
vortexing, line 1840: If `command <<< >>>` is truly the preferred way, then the basic examples when you are not likely to run into bash variable confusion, should all have this format.  ONLY if the example would have issues due to variable definition formats should the `command { }` format be used. Perhaps a find/replace for all the command sections used in the doc might be semi-painless but worth it to set folks on The Good Path?

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5846 @@
...
+There are two purely optional sections that can be used to store metadata with the task: `meta` and `parameter_meta`. These sections are designed to contain metadata that is only of interest to human readers. The engine can ignore these sections with no loss of correctness. The extra information can be used, for example, to generate a user interface. Any attributes that may influence execution behavior should go in the `runtime` section.
+
+Both of these sections can contain key/value pairs. Metadata values are different than in `runtime` and other sections:
+
+* Only string, numeric, and boolean primitives are allowed.
+* Only array and "meta object" compound values are allowed.
+* The special value `null` is allowed for undefined attributes.
+* Expressions are not allowed.
+
+A meta object is similar to a regular WDL object, except:
```
vortexing, line 2260: Here you're referring to a soon-to-be deprecated object.  Perhaps shift to `struct` here just so you don't have to fix it later and start to get your references to only things that you WANT people using?

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5846 @@
...
+* The `object` keyword is not required.
+* Its values must conform to the same metadata rules defined above.
+
+Note that, unlike the WDL `Object` type, metadata objects are not deprecated and will continue to be supported in future versions.
```
vortexing, line 2264: If you discuss these as `structs` you don't have to include this confusing sentence.

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5846 @@
...
+    # workflow-level metadata can go here
+  }
+
+  parameter_meta {
+    # metadata about each input/output parameter can go here
+  }
+}
+```
+
+A workflow can be thought of as a directed acyclic graph (DAG) of transformations that convert the input data to the desired outputs. Rather than explicitly specifying the sequence of operations, A WDL workflow instead describes the connections between the steps in the workflow (i.e. between the nodes in the graph). It is the responsibility of the execution engine to determine the proper ordering of the workflow steps, and to orchestrate the execution of the different steps.
```
vortexing, line 2507: I'd put this above the code chunk above.  You'd just scroll through the chunk, then read the text, then scroll back up.

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5846 @@
...
+struct Sample {
+  String id
+  Person donor
+  Map[String, Array[File]] assay_data
+}
+```
+
+### Struct Literals
+
+A struct literal is similar to an [`object` literal](#-object), except that the struct name is used in place of the `object` keyword. The members of a struct value can be validated against the struct's definition at the time of creation, which makes WDL code written using `struct`s more amenable to static type analysis.
```
vortexing, line 3247: Beware referring to to-be-deprecated text without mentioning it's status.

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5846 @@
...
+|     |key_2    |"value_2"|
+|     |key_3    |"value_3"|
+
+## X read_json(String|File)
+
+Reads a JSON file into a WDL value whose type depends on the file's contents. The mapping of JSON type to WDL type is:
+
+|JSON Type|WDL Type|
+|---------|--------|
+|object|`Object`|
```
vortexing, line 3877: Why does this not have `structs` in it.  I keep harping on this b/c deprecating Object leaves some ways to get data in and out via object-like var's missing.  Is the idea NOT to read a json into a struct, but that we can still WRITE a struct into a json?

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5846 @@
...
+  ./myscript -i ~{p.left} -s ~{p.right}
+  >>>
+
+  output {
+    Pair[Int, String] out = (read_int("int.txt"), read_string("str.txt")) 
+  }
+}
+```
+
+Alternatively, a `struct` or `Object` can be created with `left` and `right` members and then serialized by the appropriate function.
```
vortexing, line 5482: Perhaps remove this example using Object and make it use struct instead?

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5846 @@
...
+    Map[String, String] my_map = read_json(stdout())
+  }
+  
+  runtime {
+    container: "my_image:latest"
+  }
+}
+```
+
+`my_map` will be a `Map[String, String] with members:
```
vortexing, line 5454: You're missing a ` after : `Map[String, String]

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5846 @@
...
+  input {
+    String? s
+  }
+  if (defined(s)) {
+    call say_hello { input: name = select_first([s]) }
+  }
+}
+```
+
+## X select_first(Array[X?]+)
```
vortexing, line 4756: Instead of using X here as the return type (since you usually use it as part of the variable itself), consider using `R` as you mention at the top of the standard library section:
"The signature of each function is given as `R func_name(T1, T2, ...)`, where `R` is the return type "

Replace with: ## R select_first(Array[X?]+)

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5846 @@
...
+|     |key_2    |"value_2"|
+|     |key_3    |"value_3"|
+|1    |key_1    |"value_1"|
+|     |key_2    |"value_2"|
+|     |key_3    |"value_3"|
+|2    |key_1    |"value_1"|
+|     |key_2    |"value_2"|
+|     |key_3    |"value_3"|
+
+## X read_json(String|File)
```
vortexing, line 3871: Sounds like this:
## X read_json(String|File)

should actually be: 
## R read_json(String|File)

as per the other comment I left - you mention arbitrary Types as `R` in your standard library intro, so shouldn't use `X` here as the type.

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5846 @@
...
+}
+```
+
+This task would assign the array with elements `"foo"` and `"bar"` to `my_array`.
+
+If the echo statement was instead `echo '{"foo": "bar"}'`, the engine MUST fail the task for a type mismatch.
+
+### Struct and Object
+
+`Struct`s and `Object`s are serialized identically. A JSON object is always deserialized to a WDL `Object`, which can then be coerced to a `struct` type if necessary.
```
vortexing, line 5259: You don't provide any deserialization to a struct examples.  The section above and below both have a `write_json` and a `read_json` section.  Perhaps state why this section does not.    Also, consider removing Object here or mark with trash can for deprecation.

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5846 @@
...
+  "wf.t2.runtime.disks": "100",
+  "wf.t2.runtime.container": "mycontainer:latest"
+}
+```
+
+## JSON Serialization of WDL Types
+
+### Primitive Types
+
+All WDL types serialize naturally to JSON values:
```
vortexing, line 4939: All **primitive** WDL types serialize naturally to JSON values:

versions/1.1/SPEC.md
jdidion: good suggestion

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5846 @@
...
+
+JSON also does not have a specific type for filesystem paths, but a WDL `String` may be coerced to a `File` if necessary.
+
+### Array
+
+Arrays are represented naturally in JSON using the `array` type. Each array element is serialized recursively into its JSON format.
+
+When a JSON `array` is deserialized to WDL, each element of the array must be coercible to a common type.
+
+### Struct and Object
```
vortexing, line 4960: This seems out of order.  Better to keep the order in the text as array, pair, map then struct (leave out Object in header but retain in text with trash can). 

Also, I'm thinking you're going to want to rephrase this text too now that `object` as a WDL type is being deprecated, and becuase there are lots of references to `o/Bject`s here but in multiple contexts.  What will this text mean once `object` is gone?  How are `struct`s going to be represented in JSON?  

I think you mean something like, "`Struct`s (and the to be deprecated `Object`) themselves are represented naturally by JSON `object`s.   Each WDL `struct` (and also `Object`) member value is serialized recursively into its specific JSON type.  "

Then I don't know that I'm understanding more from the next sentence.  

"When a JSON `object` is deserialized, each value is deserialized to its most likely WDL type. " Then, state how JSON `object`s will be coerced into a `struct`.  Will this be only if the struct you want is specified?  Must each element of the JSON `object` be coercible to the type specified in the `struct` definition?  Leave out mention of `Object` which won't be a thing.

versions/1.1/SPEC.md
jdidion: Actually I don't think that section belongs in type coercion at all since it's a conversion, not a coercion. I'm moved it to its own section.

versions/1.1/SPEC.md
jdidion: Good point

versions/1.1/SPEC.md
jdidion: I think this section is out of place. I'll move it to after discussing imports.

versions/1.1/SPEC.md
jdidion: done

versions/1.1/SPEC.md
jdidion: I've revised this section. Please read the new version and see if it's more clear.

versions/1.1/SPEC.md
jdidion: Good point - I don't think it's necessary to refer to the command section at all here. I've just elided it.

versions/1.1/SPEC.md
jdidion: I turned them into task examples

versions/1.1/SPEC.md
jdidion: yep good catch

versions/1.1/SPEC.md
jdidion: yep good catch

versions/1.1/SPEC.md
jdidion: done

versions/1.1/SPEC.md
jdidion: done

versions/1.1/SPEC.md
jdidion: done

versions/1.1/SPEC.md
jdidion: fixed

versions/1.1/SPEC.md
jdidion: This is a good point. The concept of an object value will need to be kept around for this specific example. The reason is that the read_json function doesn't know the type of an object, so it can't return a struct literal. It instead returns an object that needs to be coerced to a struct with a specific type. I will add clarification.

versions/1.1/SPEC.md
jdidion: done

versions/1.1/SPEC.md
jdidion: good catch

versions/1.1/SPEC.md
jdidion: No, X is a generic type. It means that the return type is the same as the Array element type.

versions/1.1/SPEC.md
jdidion: yep, you're right here

versions/1.1/SPEC.md
jdidion: Same explanation as above.

versions/1.1/SPEC.md
jdidion: good catch

versions/1.1/SPEC.md
jdidion: done

jdidion:
Thank you @vortexing for the close reading of the spec! I believe I've incorporated most if not all of your suggestions. I think it reads much better now.

versions/1.1/SPEC.md
```diff
@@ -665,149 +690,37 @@ String a = "beware"
 String b = "key"
 String c = "lookup"
 
+struct Words {
+  Int a
+  Int b
+  Int c
+}
+
 # What are the keys to this object?
```
vortexing, line 699: Do you mean: "# What are the keys to this **`Struct`**?"

versions/1.1/SPEC.md
```diff
@@ -665,149 +690,37 @@ String a = "beware"
...
-Object object_syntax = object {
+Words literal_syntax = Words {
   a: 10,
   b: 11,
   c: 12
 }
 
 # What are the keys to this object?
```
vortexing, line 706: Same here.  Do you mean: "# What are the keys to this **`Struct`**?"

versions/1.1/SPEC.md
jdidion: yep good catch

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5967 @@
...
+🗑 It is also possible to assign an `Object` or `Map[String, X]` value to a `Struct` declaration. In the either case:
+* The `Object`/`Map` must not have any members that are not declared for the struct.
+* The value of each object/map member must be coercible to the declared type of the struct member.
+* The `Object`/`Map` must at least contain values for all of the struct's non-optional members.
+
+Note that the ability to assign non-`Struct` values to `Struct` declarations is deprecated and will be removed in WDL 2.0.
+
+### Struct Namespacing
+
+Although a `struct` is a top-level element, it is not considered a member of the WDL document's namespace the way that other top-level elements (`task`s and `workflow`s) are. Instead, when a WDL document is imported all of its `structs` are added to a global struct namespace. This enables structs to be used by their name alone, without the need for any any `namespace.` prefix.
```
DavyCats, line 3366:
```suggestion
Although a `struct` is a top-level element, it is not considered a member of the WDL document's namespace the way that other top-level elements (`task`s and `workflow`s) are. Instead, when a WDL document is imported all of its `structs` are added to a global struct namespace. This enables structs to be used by their name alone, without the need for any `namespace.` prefix.
```

jdidion:
Voting is open. Obviously I'm +1.

patmagee:
Huge and enthusiastic +1 for me
:+1:

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5967 @@
...
+The `command` section is the only required task section. It defines the command template that is evaluated and executed when the task is called. Specifically, the commands are executed after all of the inputs are staged and before the outputs are evaluated.
+
+There are two different syntaxes that can be used to define the command section:
+
+```wdl
+# HEREDOC style - this way is preferred
+command <<< ... >>>
+
+# older style - may be preferable in some cases
+command <<< ... >>>
```
mlin, line 1610:
```suggestion
command { ... }
```

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5967 @@
...
+task write_array {
+  input {
+    Array[String] str_array
+  }
+  command <<<
+    # the "write_lines" function writes each string in an array
+    # as a line in a temporary file, and returns the path to that
+    # file, which can then be referenced by other commands such as 
+    # the unix "cat" command
+    cat write_lines(str_array)
```
mlin, line 1658:
```suggestion
    cat ~{write_lines(str_array)}
```

mlin (approved):
Two trivial corrections noted, otherwise :+1: from me thanks!

geoffjentry:
:+1:

rhpvorderman:
:+1: From me as well. Thanks for tackling this huge task!

DavyCats:
:+1:

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5967 @@
...
+  "example.kessel_run_parsecs": 11.98,
+  "example.sample_swap_detected": false,
+  "example.sample_variants": ["/data/patient1.vcf", "/data/patient2.vcf"],
+  "example.droids": {"C": 3, "D": 2, "P": 0, "R": 2}
+}
+```
+
+It is recommended (but not required) that JSON outputs be "pretty printed" to be more human-readable.
+
+## Specifying / Overriding Runtime Attributes
```
pieterlukasse, line 5019: This is in! I'm happy 👍

aednichols:
👍

cjllanwarne:
👍  phenomenal work. Thanks so much to @jdidion in particular and to everyone else who has put in so much time and effort contributing and reviewing to get this WDL 1.1 SPEC to this state!

vdauwera:
👍  Amazing! Huge props to @jdidion and everyone who participated, this is going to be so great for the community. 💯

vortexing:
👍🏼
👏🏼👏🏼👏🏼👏🏼👏🏼

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5967 @@
...
+The runtime container may use a non-standard Bash shell that supports more complex glob strings, such as allowing expansions that include `a_inner.txt` in the example above. To ensure that a WDL is portable when using `glob`, a container image should be provided and the WDL author should remember that `glob` results depend on coordination with the Bash implementation provided in that container.
+
+## String basename(String|File, [String])
+
+Returns the "basename" of a file - the name after the last directory separator in the file's path. The optional second parameter may specify a suffix to remove from the file name.
+
+**Parameters**
+
+1. (`String`|`File`): Path of the file to read. If the argument is a `String`, it is assumed to be a local file path relative to the current working directory of the task.
+2. `[String]`: Suffix to remove from the file name.
```
rhpvorderman, line 3722: In cromwell the second string has as a regex possibility. This is useful for certain types of files. For example: `\.ya?ml` matching `.yml` and `.yaml`. 

Should that be codified in the spec as well?

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5967 @@
...
+
+**Example**
+
+```wdl
+String chocolike = "I like chocolate when it's late"
+
+String chocolove = sub(chocolike, "like", "love") # I love chocolate when it's late
+String chocoearly = sub(chocolike, "late", "early") # I like chocoearly when it's early
+String chocolate = sub(chocolike, "late$", "early") # I like chocolate when it's early
+String chocoearlylate = sub(chocolike, "[^ ]late", "early") # I like chocoearly when it's late
```
rhpvorderman, line 3594: @DavyCats mentioned in the regex PR that `^[ ]late` matches `olate` not just `late` so the resulting string should be `I like chocearly when it's late`

versions/1.1/SPEC.md
```diff
@@ -0,0 +1,5967 @@
...
+    # these two expressions are equivalent
+    Float max1 = if value1 > value2 then value1 else value2
+    Float max2 = max(value1, value2)
+  }
+}
+``` 
+
+## String sub(String, String, String)
+
+Given 3 String parameters `input`, `pattern`, `replace`, this function replaces all non-overlapping occurrences of `pattern` in `input` by `replace`. `pattern` is a [regular expression](https://en.wikipedia.org/wiki/Regular_expression) that will be evaluated as a [POSIX Extended Regular Expression (ERE)](https://en.wikipedia.org/wiki/Regular_expression#POSIX_basic_and_extended).
```
rhpvorderman, line 3574: How should the regex string be interpreted regarding escape characters?

I.e. in WDL a string with a backslash is `\\`. Where the former backslash escapes the second. So now for regex there are two possibilities.

1. The regex is written as a WDL string. In POSIX ERE a backslash needs to be escaped, so it is `\\`. When the regex is written as a WDL string it needs to be written as `\\\\`. The engine then converts the string properly to `\\`. The regex interpreter then knows it should match `\`.
2. The regex is written as a raw string. Meaning `\\` will translate to a regex `\\`, which matches `\`. 

I prefer option 2. But there is some ambiguity about this. See the issue @mlin posted on biowdl/tasks: https://github.com/biowdl/tasks/issues/270.

jdidion: Speaking from the perspective of the parser, option 1 is much cleaner. Otherwise we need to put in a special case for every function that accepts a regexp OR have a special raw string prefix like in python (e.g. r""). For now I say we go with option 1 and consider the change for option 2 in the 2.0 spec.

versions/1.1/SPEC.md
jdidion: good catch

versions/1.1/SPEC.md
rhpvorderman: The point is that this changed between two releases of cromwell at some point. Whichever option is chosen it should be unambiguous from the spec. 
I can understand the parser perspective, but for the user writing `\\\\` to mean `\` is quite a nightmare. 

> The Workflow Description Language (WDL) is a way to specify data processing workflows with a **human-readable and -writeable syntax**. WDL makes it straightforward to define analysis tasks, chain them together in workflows, and parallelize their execution. The language makes common patterns **simple to express**, while also admitting uncommon or complicated behavior; and strives to achieve portability not only across execution platforms, but also different types of users. Whether one is an **analyst, a programmer, an operator of a production system, or any other sort of user**, WDL should be accessible and understandable. 

emphasis mine. I think it would be more fitting with the WDL philosophy to go for option 2. People can look up how regexes work, and their copy-pasted regex examples will work as intended without needing to worry about escaping their regex escape characters. 
Currently cromwell does support the raw string option (at least it seems to work that way). So it would be a shame to lose that. But that is just my 2 cents.

jdidion: Yes, but Cromwell != WDL.

I agree that it should be unambiguous, but I don't think it's appropriate to add it in this version of the spec.

rhpvorderman: Option 1 is far better than leaving it ambiguous. And if it is easier to implement it will mean we will get v1.1 implementation sooner which is also desirable.
Adding raw strings later is a very good idea. A raw string construct makes it easier for the parser as well as the user, while being explicit as well.

versions/1.1/SPEC.md
jdidion: I think we can consider this. Is the regexp a full POSIX ERE or a simple glob expression?

The larger question is whether we're willing to add new functionality in this version of the spec or if it should go through the PR process. The decision should also apply to the new `unzip` function.

versions/1.1/SPEC.md
jdidion: Great - do you want to make a PR for raw strings?

rhpvorderman: Yup, I will put it on my todo list :slightly_smiling_face: .

versions/1.1/SPEC.md
mlin: If I write `basename(path, ".yml")` does the `.` match any character? (seems counterintuitive for this example of the common usage)

jdidion: Good question/point. If a regexp is allowed, then the '.' always has to be escaped. The alternative would be to introduce a different syntax for regular expressions, e.g. `/regexp/` that can only be used in specific situations.

rhpvorderman: For the people who need `\.ya?ml` there is already `sub(string, string, string)`. So I am inclined to vote for `basename(path, extension)` being kept simple. Even if that breaks some of the files in biowdl/tasks (what do you think @DavyCats ?) . Maybe: basename(path, extension, regex=false) is better. Because regex=true can then be explicitly set. But that would be for a later version.

jdidion: I agree. Let's leave the regexp option out in 1.1 and discuss whether to add a specific regexp syntax in 2.0 or support the third `basename` parameter.

PR #432, opened 2020-12-25
fixed file type in description of read_json()

notestaff:
### Checklist
- [X] Pull request details were added to CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@ Keep the changelog pleasant to read in the text editor:
 
 version 2.0.0
 ---------------------------
++ Fixed file type in the description of read_json.
+  [PR 432](https://github.com/openwdl/wdl/pull/432) by @notestaff
+
 + Abbreviated syntax for call inputs bound to workflow-scoped values by name:
   `{input: x, y=b, z}` is shorthand for `{input: x=x, y=b, z=z}`
   [PR 365](https://github.com/openwdl/wdl/pull/365) by @mlin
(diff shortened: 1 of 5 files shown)


geoffjentry:
I'm curious what other folks think here. Traditionally we've taken the stance to not retroactively fix issues like this - i.e. official versions of the spec aren't living documents.That does of course bring some cons to the party along with the pros it has

notestaff:
Could also make a separate "errata" document, and add to the original only footnotes with a hyperlink to the erratum.

But it seems that the "official version of the spec" never really specified that read_json() argument should be a TSV file, it just left implementers to figure out that there's a typo.

jdidion:
I like the idea of an errata document.

geoffjentry:
@jdidion likewise

notestaff:
If an errata document, should an unofficial version with the errata applied (and highlighted?) be maintained as well?  Should the original be annotated with links/footnotes to errata?

aofarrel:
Any updates on this? There are a few others with the WDL 1.0 spec which need to be addressed, but if neither an errata nor updates to the spec go through, then unfortunately there is no good resource for anyone wishing to write WDLs compatible with Cromwell/Terra. I get that the spec isn't officially tied to any one executor, but my understanding is that WDL 1.0 is still very widely used and its spec deserves to be fixed for that reason.

I think the best way to do this would be to update the original in the case of clear typos and adding examples, but failing that, I lean towards annotating the original. At the very least there has to be a link on the original where someone can find the errata; if it's not linked in the current spec than it is unlikely people will find it.

jdidion:
@notestaff @aofarrel if you'd still like to see this fixed in earlier versions of WDL (1.0 and earlier) please create a PR adding/updating an ERRATA doc adjacent to those versions of the spec. Thanks

PR #438, opened 2021-02-19
Add errata for v1.1

jdidion:
### Checklist


patmagee (approved):
This looks good to me. I like thee strong cautio against using this approach. Changing this in 1.1 could potentially prove backwards incompatible for some people, so stating this is a special case is a good compromise.

Issue #348, opened 2019-12-14
Hex code of single quote and double quote is upside-down


yihming:
I just found that the hex equivalent of single quote and double quote in [development SPEC](https://github.com/openwdl/wdl/blob/master/versions/development/SPEC.md) should be switched. I.e. single quote is \x27, while double quote is \x22.

yihming:
@vdauwera Thank you for taking care of this issue.

PR #451, opened 2021-04-01
Fix hex codes for single/double quotes in spec

vdauwera:
Addresses #348 reported by @yihming


patmagee (approved):

patmagee:
@vdauwera should we fix this in any other of the spec versions?

vdauwera:
Oh I didn't think about fixing them retroactively, but yeah that's a good idea

patmagee (approved):

jdidion:
We need to backport this to 1.1.1

PR #520, opened 2022-12-14
Fix typo in sample code

jdavcs:
Just fixing a typo: identifier should correspond to `inc` task output. 
Not sure if this should be added to `CHANGELOG.md`; is so, I can add a commit.

### Checklist
- [ ] Pull request details were added to CHANGELOG.md

PR #521, opened 2022-12-14
Fix typo in sample code

jdavcs:
Just fixing a typo: identifier should correspond to `inc` task output.
Not sure if this should be added to `CHANGELOG.md`; is so, I can add a commit.
(this replaces #520)

NOTE: just noticed this is for version 1.0  which is no longer current. Please ignore if this is no longer relevant.

### Checklist


jdidion:
@jdavcs This example has been fixed/changed in WDL 1.1. If you'd still like to see this fixed in earlier versions of WDL (1.0 and earlier) please create a PR adding/updating an ERRATA doc adjacent to those versions of the spec. Thanks

jdavcs:
@jdidion Given that it's no longer current, +1 to leave it as is. Thanks for looking into this!

PR #524, opened 2022-12-27
make 'input:' optional in call body

mlin:
Just scratching a holiday itch....the `input:` heading in call bodies is useless clutter AFAICT. Make it optional (for backwards compatibility).

Perhaps the idea was for call bodies to have additional sections (e.g. `runtime:` for workflow-level overrides). Those could still be added in the future, whilst keeping `input:` implied if not present.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,9 @@ version development
   consist of multiple files.
   [PR 241](https://github.com/openwdl/wdl/pull/241) by @cjllanwarne.
 
++ Make `input:` optional in call bodies.
+  [PR 524](https://github.com/openwdl/wdl/pull/524) by @mlin.
+
 version 1.1.0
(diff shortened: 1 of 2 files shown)


rhpvorderman:
I like this idea. Less is more. I like how it is optional too, having maximal backwards-compatibility where possible is the only way forward. When WDL 2.0  eventually launches I don't want it to become the python 3.0 disaster where only the core devs are on the new stuff and the rest of the world waits for 10+ years.

patmagee:
I love this idea too. I have always felt that in lieu of other properties it was just added boilerplate to the call definition. @mlin are you able to take a stab at updating the grammar bychance?

patmagee (approved):
LGTM!

jdidion:
@mlin please change target to `wdl-1.2`

PR #527, opened 2023-01-10
fix: parameter_meta example

j23414:
Tried uploading the example to Terra.bio, but needed to make the following changes before the wdl worked:

* add missing input brackets
* remove breaking commas from the parameter_meta field

### Checklist

### DIFF
--- a/versions/1.0/SPEC.md
+++ b/versions/1.0/SPEC.md
@@ -1188,17 +1188,19 @@ This purely optional section contains key/value pairs where the keys are names o
 For example:
 ```wdl
 task wc {
-  File f
-  Boolean l = false
-  String? region
+  input {
+    File f
+    Boolean l = false
+    String? region
+  }
   parameter_meta {
-    f : { help: "Count the number of lines in this file" },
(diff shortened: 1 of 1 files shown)


j23414:
In case it helps, here is the full script uploaded to  https://portal.firecloud.org/?return=firecloud#methods

```
version 1.0

task wc {
  input {
    File f
    Boolean l = false
    String? region
  }
  parameter_meta {
    f : { help: "Count the number of lines in this file" }
    l : { help: "Count only lines" }
    region: {help: "Cloud region", suggestions: ["us-west", "us-east", "asia-pacific", "europe-central"]}
  }
  command <<<
    wc ${true='-l', false=' ' l} ~{f}
  >>>
  output {
    String retval = read_string(stdout())
  }
  runtime {
    docker : 'ubuntu:latest'
  }
}

workflow WC_WORKFLOW {
  input {
    File f
    Boolean l = false
    String? region
  }

  call wc {
    input: 
      f=f,
      l=l,
      region=region
  }
  output {
    String stdout = wc.retval
  }
}
```

PR #548, opened 2023-03-28
WDL 1.1.1

jdidion:
I wanted people to be able to start looking at this. Still a bit of work to be done to update all the examples to the new format. I also don't want to merge this until we get the new repo structure figured out.


SPEC.md
```diff
@@ -490,415 +677,1104 @@ An `Array` represents an ordered list of elements that are all of the same type.
...
+The order of precedence is:
 
-| Operator | RHS Type  | Result    |
-| -------- | --------- | --------- |
-| `-`      | `Float`   | `Float`   |
-| `-`      | `Int`     | `Int`     |
-| `!`      | `Boolean` | `Boolean` |
+1. `Int`, `Float`: `Int` coerces to `Float`
+2. `X`, `Y`: For any primitive types `X` != `Y`, both are coerced to `String`
+3. If applying `+` to two values of the same type that cannot otherwise be summed/concatenated (i.e., `Boolean`, `File`, `Directory`), both values are first coerced to `String`
```
patmagee, line 1372: This stanza is a little ambiguous to me and does not clearly define the order of oeprations. Could we potentially add an example or several examples?

SPEC.md
```diff
@@ -490,415 +677,1104 @@ An `Array` represents an ordered list of elements that are all of the same type.
...
+  "fail": true
+}
+```
+</p>
+</details>
+
+### Expressions
+
+```txt
+$expression = '(' $expression ')'
```
patmagee, line 1636: I dont know if removing this would be considered a breaking change, but we could always link to the grammar instead of embedding this expressions

SPEC.md
```diff
@@ -935,3871 +1811,7403 @@ In operations on mismatched numeric types (e.g. `Int` + `Float`), the `Int` type
...
+#### Reserved `runtime` hints
 
-```wdl
-Int i = 3
-String s = "~{1 + i}"  # s == "4"
-```
+The following attributes are considered "hints" rather than requirements. They are optional for execution engines to support. The purpose of reserving these attributes is to encourage interoperability of tasks and workflows between different execution engines.
 
-As another example, consider how the following expression would be parsed:
+In WDL 2.0, these attributes will move to the new `hints` section.
```
patmagee, line 4558: we are introducing hints in 1.2, so these attributes will then be made available there will they not?

SPEC.md
```diff
@@ -935,3871 +1811,7403 @@ In operations on mismatched numeric types (e.g. `Int` + `Float`), the `Int` type
...
+The following guidelines should be followed when using hints:
+
+* A hint should never be required.
+* Less is more. Before adding a new hint key, ask yourself "do I really need another hint?", or "is there a better way to specify the behavior I require". If there is, then adding a hint is likely not the right way to achieve your aims.
+* Complexity is killer. By allowing any arbitrary keys, it is possible that the `runtime` section can get quite unruly and complex. This should be discouraged, and instead a pattern of simplicity should be stressed.
+* Sharing is caring. People tend to look for similar behavior between different execution engines. It is strongly encouraged that execution engines agree on common names and accepted values for hints that describe common usage patterns. A good example of hints that have conventions attached to them is cloud provider specific details:
+    ```wdl
+    task foo {
+      .... 
+      runtime {
```
patmagee, line 4721: was the decision for our last call to rename `hints` to `runtime` and make them functionally equivalent? we added the new `requirements` section

patmagee (commented):
@jdidion this is a heroic effort, thank you so much for taking this on! The approach you took for the testing and examples is quite clever and will help build a compliance suite for engines moving forward. I think the structure is simple, easy to enforce and easy to generate

Overall, I am in favor of the changes that were proposed. I have left a few comments and have one main question around the use of `runtime` as `hints`. IIRC we are introducing introducing `hints` explicitly in `1.2`, but I noticed the documentation now refers to keys in the `runtime` block as `hints`, was this intentional?

tests/README.md
```diff
@@ -0,0 +1,215 @@
...
+|_ ...
+|_ test_config.json
+```
+
+The `test_config.json` file contains a JSON array with one element for each test, where each element is an object with the test inputs, outputs, and configuration parameters. For example:
+
+```json
+[
+  {
+    "id": "foo",
```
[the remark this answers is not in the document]
jdidion: Good catch - yes it defaults to `target` but can be overridden.

.github/workflows/codespell.yml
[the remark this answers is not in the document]
jdidion: Wasn't my idea :) It was submitted in one of the PRs by our mysterious typo hunters.

SPEC.md
jdidion: I'm not sure what the right thing to do is. But honestly I haven't reviewed these grammar sections closely - there are probably errors or inconsistencies with the official grammar. If we keep them in the spec we should at least switch to a standard notation like BNF.

SPEC.md
jdidion: Good catch

SPEC.md
jdidion: In 1.2 we will have `requirements` and `hints`. `runtime` will be deprecated but still supported until `2.0`. The use of `runtime` is orthogonal to the use of `requirements` + `hints` - i.e. you can do one or the other but not mix them.

SPEC.md
geoffjentry: TBH I never really liked the pseudo-grammar blocks. I thought I had removed them at some point long ago but I guess not. Perhaps I just nicked a few of them 🤷 

I would advocate for removing them and then as @jdidion said, perhaps look at providing a BNF as an appendix or something

geoffjentry: Looking at some grammars being presented prompted the thought -> I think someone already **has** done a BNF of WDL, or at least a partial one. Will try to dig it up, assuming it wasn't a hallucination on my part

geoffjentry: Aha - top of the google search `WDL BNF` - @cjllanwarne can say more to what rough edges are here: https://github.com/broadinstitute/winstanley/blob/44896516cef2dadac35e7af0fc0d5c4dc8de2ff2/src/winstanley/wdl.bnf#L4

jdidion: Nice...I'll open an issue to add that as an appendix or otherwise clone it into openwdl

patmagee (approved):
This looks great @jdidion!

patmagee (approved):
This looks great @jdidion!

jdidion:
Hi all - in the Sept 27 governance call, those present voted to approve this PR, and to merge it and release WDL 1.1.1 in one week's time contingent on their being no objections from any of the members who weren't present. So speak now or forever hold your peace!

rhpvorderman:
I was not present due to timeslot issues, but I am in full agreement. Thanks @jdidion for all this work!

Issue #540, opened 2023-03-23
Add new `requirements` section, deprecate `runtime`


jdidion:
In the 3/10/23 governance call we decided that, rather than backport the changes to `runtime` that currently appear in the development version, we will instead deprecate the `runtime` section and add a new `requirements` section that will have the semantics of the development `runtime` section.

DavyCats:
Changing from `runtime` to `requirements` would constitute a breaking change, so this would have to be part of 2.0, right? Unless `requirements` gets added as a synonym in 1.2 while `runtime` remains available as well.

jdidion:
Yes, sorry, should have added more details. runtime will be deprecated but still supported in 1.2. It will be removed in 2.0.

patmagee:
@DavyCats as @jdidion mentioned, we are not going to be breaking the spec here but will introduce an explicit deprecation of the `runtime` block in favour of the `requirements` section. 

They will co-exist up until the 2.0 release. `requirements` provide a stricter vocabulary hopefully helping with portability

vsmalladi:
Im not sure this is necessary. This seems unnecessary to move from `runtime` to `requirements` what is the value we are getting by doing this?

jdidion:
Originally, we had planned in WDL 2.0 to change the semantics of `runtime` (in a breaking way) to only allow specific attributes. All optional attributes were to move to `hints`.

We then decided to implement this change in WDL 1.2, but since we cannot introduce a breaking change, we decided instead to deprecate `runtime` in favor of `requirements` and `hints`. The name `requirements` is also more clear than `runtime` regarding the intent of that section, and it also matches the term used in `CWL`.

PR #595, opened 2023-12-11
feat: requirements and hints

jdidion:
This PR adds the new `requirements` and `hints` sections, and deprecates `runtime`.

There are a few additions here that were not in the original spec for this feature but which I think are needed to make it complete:

* Added the `fpga` requirement - similar to `gpu`, it is a `Boolean` that specifies whether the task requires a FPGA. The popularity of DRAGEN, and the increasing number of other FPGA accelerated bioinformatics tools, make this essential.
* Added the `gpu` and `fpga` hints to enable more detailed requests for hardware accelerators (e.g., number and type).
* Added the `disks` hint to enable more detailed requests for disk volumes (e.g., class and permissions).
* Stipulated compute environment-specific hints can override task-specific hints
* Deprecated the previously allowed behavior implied by setting `allowNestedInputs: true` where required task/subworkflow inputs could be left unsatisfied. Now all inputs either need to have a default value or have their value specified in the call inputs. Only optional task/subworkflow inputs that are not explicitly set in the call inputs may have their value set at runtime.

### Checklist

Closes issues: #540, #541 #543

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -1451,10 +1451,11 @@ A hidden type is one that may only be instantiated by the execution engine, and 
 
 ##### Union
 
-The `Union` type is used for a value that may have any one of several concrete types. A `Union` value must always be coerced to a concrete type. The `Union` type is used in two contexts:
+The `Union` type is used for a value that may have any one of several concrete types. A `Union` value must always be coerced to a concrete type. The `Union` type is used in the following contexts:
 
 * It is the type of the special [`None`](#optional-types-and-none) value.
 * It is the return type of some standard library functions, such as [`read_json`](#read_json).
+* It is the type of some reserved [`requirements`](#✨-requirements-section) and [`hints`](#✨-hints-section) attributes.
 
 #### Scoped Types
 
--- a/SPEC.md
+++ b/SPEC.md
@@ -37,6 +37,7 @@ Revisions to this specification are made periodically in order to correct errors
         - [Custom Types (Structs)](#custom-types-structs)
       - [Hidden Types](#hidden-types)
         - [Union](#union)
+      - [Scoped Types](#scoped-types)
       - [Type Conversion](#type-conversion)
         - [Primitive Conversion to String](#primitive-conversion-to-string)
         - [Type Coercion](#type-coercion)
@@ -84,7 +85,7 @@ Revisions to this specification are made periodically in order to correct errors
     - [Evaluation of Task Declarations](#evaluation-of-task-declarations)
     - [✨ Requirements Section](#-requirements-section)
       - [Units of Storage](#units-of-storage)
-      - [`requirements` attributes](#requirements-attributes)
+      - [Requirements attributes](#requirements-attributes)
         - [`container`](#container)
         - [`cpu`](#cpu)
         - [`memory`](#memory)
@@ -92,12 +93,11 @@ Revisions to this specification are made periodically in order to correct errors
         - [`disks`](#disks)
         - [`max_retries`](#max_retries)
         - [`return_codes`](#return_codes)
-    - [🗑 Runtime Section](#-runtime-section)
-    - [Meta Values](#meta-values)
     - [✨ Hints Section](#-hints-section)
+      - [Hints-scoped types](#hints-scoped-types)
       - [Reserved Task Hints](#reserved-task-hints)
         - [`max_cpu`](#max_cpu)
-        - [✨ `accelerators`](#-accelerators)
+        - [✨ `gpus` and ✨ `fpgas`](#-gpus-and--fpgas)
         - [`max_memory`](#max_memory)
         - [✨ `volumes`](#-volumes)
         - [`short_task`](#short_task)
@@ -106,7 +106,9 @@ Revisions to this specification are made periodically in order to correct errors
         - [`outputs`](#outputs)
       - [Compute Environments](#compute-environments)
       - [Conventions and Best Practices](#conventions-and-best-practices)
+    - [🗑 Runtime Section](#-runtime-section)
     - [Metadata Sections](#metadata-sections)
+      - [Meta Values](#meta-values)
       - [Task Metadata Section](#task-metadata-section)
       - [Parameter Metadata Section](#parameter-metadata-section)
     - [Advanced Task Examples](#advanced-task-examples)
@@ -1453,6 +1455,12 @@ The `Union` type is used for a value that may have any one of several concrete t
 
 * It is the type of the special [`None`](#optional-types-and-none) value.
 * It is the return type of some standard library functions, such as [`read_json`](#read_json).
+
+#### Scoped Types
+
+A scoped type is one that may only be defined by the execution engine within a specific scope, but which may be instantiated by the user within that scope.
+
+Currently, scoped types limted to the [`hints`](#✨-hints-section) section.
 
 #### Type Conversion
 
@@ -3299,7 +3307,7 @@ A task has a required [`command`](#command-section) that is a template for a Bas
 
 Tasks explicitly define their [`input`s](#task-inputs) and [`output`s](#task-outputs), which is essential for building dependencies between tasks and workflows. The value of an input declaration may be supplied by the caller. Tasks may have additional "private" declarations within the task body. All task declarations may be initialized with hard-coded literal values, or may have their values constructed from expressions. Input and private declarations can be referenced in the command template.
 
-A task may also specify runtime environment [requirements](#requirements-section) (such as the amount of RAM or number of CPU cores) that must be satisfied in order for its commands to execute properly, and [hints](#hints-section) that should be satisfied if possible.
+A task may also specify runtime environment [requirements](#requirements-section) (such as the amount of RAM or number of CPU cores) that must be satisfied in order for its commands to execute properly, and [hints](#✨-hints-section) that should be satisfied if possible.
 
 There are two optional metadata sections: the [`meta`](#metadata-sections) section, for task-level metadata, and the [`parameter_meta`](#parameter-metadata-section) section, for parameter-level metadata.
 
@@ -4366,7 +4374,7 @@ Optional whitespace is allowed between the number/expression and the suffix. For
 
 The decimal and binary units may be shortened by omitting the trailing "B". For example, "K" and "KB" are both interpreted as "kilobytes".
 
-#### `requirements` attributes
+#### Requirements attributes
 
 The following attributes must be supported by the execution engine. The value for each of these attributes must be defined - if it is not specified by the user, then it must be set to the specified default value.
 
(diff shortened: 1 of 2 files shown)


rhpvorderman (commented):
Excellent write-up. As a general discussion point, I strongly disagree with not allowing expressions in the hints section. 

The SLURM cluster for instance requires jobs to be given a maximum runtime. This can be dynamically derived from the filesize when it is an expression. Currently we use `time_minutes` in the `runtime` section. Under the new scheme this would classify as a hint and thus expressions cannot be used. 
Another solution is to add time to the requirements section, but I think this strategy is untenable. There are quite a lot of execution backends and as a consequence there will be unforeseen requirements that need to be solved in the hints section. Therefore allowing expressions in the hints section is preferable, so hints can be tuned to the input data. Since I work with widely disparate filesizes, that would make my life a lot easier. Furthermore, the machinery for expression evaluation is already present in the execution engines, so why not use it?

jdidion:
@rhpvorderman I agree with you. I don't really remember the argument for `hints` being like `meta` rather than `runtime` but from an implementation perspective I don't feel like it would be adding that much additional complexity, and as you say it would make hints much more useful.

It would be great if others can weigh in on this PR, especially @mlin and someone from the Cromwell/Terra team (@cjllanwarne?). Hopefully we can discuss at the next meeting if there is disagreement.

jdidion:
Two challenges with allowing expressions in the `hints` section:

* Unlike `requirements`, some hints are expected to have `object` values. The `inputs` and `outputs` hints are the best example. If we are deprecating the `Object` type, then how does the user actually create these? Maybe we allow object literals only within `hints`?
* Hints can be specified at the workflow level. Do we allow expressions there? Are expressions evaluated in the context of the workflow, or the task that inherits them?

jdidion:
I addressed this by introducing a (yet another) new concept: scoped types. These are similar to hidden types in that they may only be declared by the execution engine, but different in that a) they are only declared in a specific scope and b) they may be instantiated by the user.

I introduced three types scoped to the hints section: `input`, `output`, and `hints`. The use of reserved words for naming is intentional, so that they don't conflict with existing user-defined types.

SPEC.md
```diff
@@ -3933,28 +3952,26 @@ Input and private declarations may appear in any order within their respective s
 
 Declarations in the output section may reference any input and private declarations, and may also reference other output declarations.
 
-### Runtime Section
-
-The `runtime` section defines a set of key/value pairs that represent the minimum requirements needed to run a task and the conditions under which a task should be interpreted as a failure or success. 
+### ✨ Requirements Section
 
-During execution of a task, resource requirements within the `runtime` section must be enforced by the engine. If the engine is not able to provision the requested resources, then the task immediately fails. 
+The `requirements` section defines a set of key/value pairs that represent the minimum requirements needed to run a task and the conditions under which a task should be interpreted as a failure or success. The `requirements` section is limited to the attributes defined in this specification. Arbitrary key/value pairs are not allowed in the `requirements` section, and must instead be placed in the [`hints`](#✨-hints-section) section.
```
markjschreiber, line 4313: Nit - should probably remove the word minimum as some requirements are maximums.

SPEC.md
```diff
@@ -4002,28 +4019,28 @@ Optional whitespace is allowed between the number/expression and the suffix. For
...
+The user is strongly suggested to specify a `container` for every task. If `container` is not specified, or is specified with the special `"*"` value, the execution behavior is determined by the execution engine. A task that depends on the engine to determine the execution environment should be careful to only use built-in Bash operations and tools specified as mandatory by the [POSIX standard](https://unix.stackexchange.com/questions/228888/what-are-the-posix-mandatory-utilities).
 
-The format of a container URI string is `protocol://location`, where protocol is one of the protocols supported by the execution engine. Execution engines must, at a minimum, support the `docker://` protocol, and if no protocol is specified, it is assumed to be `docker://`. An execution engine should ignore any URI with a protocol it does not support.
+The format of a container URI is `protocol://location`, where `protocol` is one of the protocols supported by the execution engine. Execution engines must, at a minimum, support the `docker` protocol. If only `location` is specified, the protocol is assumed to be `docker`. An execution engine should ignore any URI with a protocol it does not support.
```
markjschreiber, line 4395: `An execution engine should ignore any URI with a protocol it does not support.`

If there isn't at lease one it supports in almost all cases I suspect this will lead to a workflow failure. Should we call that out here?

Because the language doesn't have any constructs for saying how to authenticate with the repository, there is also the assumption that the engine will do this based on some configuration or roles (or fail if it doesn't know how to do this).

SPEC.md
```diff
@@ -4002,28 +4019,28 @@ Optional whitespace is allowed between the number/expression and the suffix. For
...
 
-Container source locations should use the syntax defined by the individual container repository. For example an image defined as `ubuntu:latest` would conventionally refer a docker image living on `DockerHub`, while an image defined as `quay.io/bitnami/python` would refer to a `quay.io` repository.
+A container location uses the syntax defined by the container repository. For example, the URI `ubuntu:latest` refers to a Docker image hosted on `DockerHub`, while the URI `quay.io/bitnami/python` refers to an image in a `quay.io` repository.
```
markjschreiber: Can we state that URI patterns with SHA sums are allowed (perhaps even encouraged when repository tags are mutable)?

SPEC.md
```diff
@@ -4586,31 +4618,58 @@ Example output:
...
 
-##### `maxCpu`
+##### `max_cpu`
+
+* Accepted types:
+    * `Int`
+    * `Float`
+* Alias: `maxCpu`
 
 Specifies the maximum CPU to be provisioned for a task. The value of this hint has the same specification as [`requirements.cpu`](#cpu).
```
markjschreiber: I think there needs to be some guidance around what max means (or could mean depending on the engine). For example if I set a max CPU (or memory) should I expect that hint to result in a (probably hard) limit on a container or cgroup or should I expect that this means that the compute instance that gets the task may have no more than X cpu or memory (which might lead to conflicts in the ability to provision the minimums). If we leave it completely up to the engines to decide then different interpretations by different engines would reduce or remove any possible value of the hint.

It's fine to say that the hint can be ignored but if it isn't what should a user expect?

SPEC.md
```diff
@@ -4586,31 +4618,58 @@ Example output:
...
+
+A `Boolean` value for which `true` indicates that that this task is not expected to take long to execute. The execution engine can interpret this as permission to attempt to optimize the execution of the task. For example, the engine may batch together multiple `short_task`s, or it may use the cost-optimized instance types that many cloud vendors provide, e.g., `preemptible` instances on `GCP` and `spot` instances on `AWS`.
+
+##### `localization_optional`
+
+* Accepted types: `Boolean`
+* Default value: `false`
+* Alias: `localizationOptional`
 
 A `Boolean` value, for which `true` indicates that, if possible, any `File` inputs for this task should not be (immediately) localized. For example, a task that processes its input file once in linear fashion could have that input streamed (e.g., using a `fifo`) rather than requiring the input file to be fully localized prior to execution. This directive must not have any impact on the success or failure of a task (i.e., a task must run the same with or without localization).
```
markjschreiber: > This directive must not have any impact on the success or failure of a task (i.e., a task must run the same with or without localization).

I think this means that the script of the task (and the applications used by the task) must be able to work in either case (e.g. they can use a local file path or a s3:// or gs:// style path). Or should this be interpreted to mean that the engine must be able to detect the capabilities of the task and called applications and do the right thing? Maybe we need to say who is responsible for ensuring this directive doesn't impact success (the WDL workflow author or the workflow engine).

SPEC.md
```diff
@@ -5373,18 +5452,108 @@ The following fully-qualified names exist when calling `workflow main` in `main.
...
+#### Reserved Workflow Hints
+
+The following hints are reserved. An implementation is not required to support these attributes, but if it does support a reserved attribute it must enforce the semantics and allowed values defined below. The purpose of reserving these hints is to encourage interoperability of tasks and workflows between different execution engines.
+
+##### `allow_nested_inputs`
+
+* Allowed type: `Boolean`
+* Alias: `allowNestedInputs`
+
+By default, when a workflow calls a subworkflow or task it must provide inputs for all of the subworkflow/task inputs. However, if the runtime engine supports the `allow_nested_inputs` hint and it is set to `true` in a workflow's `hints` section, then that workflow is allowed to leave any subworkflow/task inputs unsatisfied. Any unsatisfied inputs must be specified by at runtime. If a runtime engine does not support `allow_nested_inputs` or if any inputs remain unsatisfied at runtime then the workflow fails with an error.
```
markjschreiber: > specified by at runtime

Typo (specified at runtime)

markjschreiber: > If a runtime engine does not support `allow_nested_inputs` or if any inputs remain unsatisfied at runtime then the workflow fails with an error

Isn't this counter to the statement above "A workflow execution never fails due to the inability of the execution engine to recognize or satisfy a hint"

markjschreiber (commented):
Looks great. Might be some places where we can be more explicit that I have mentioned in comments.

SPEC.md
jdidion: `max*` attributes are all in `hints`

SPEC.md
jdidion: Good point. If no value is specified then the default is `*`, but I don't think it's a good idea to fall back to the default if the user does specify value(s) but none are valid/supported. I made this distinction explicit.

I think the second part of your comment is about private Docker repositories - yes, it's up to the engine to specify the mechanism for how credentials are provided.

SPEC.md
markjschreiber: Got it.

SPEC.md
jdidion: good suggestion

SPEC.md
jdidion: In this case I think the hint is really just meant as a potential optimization. I.e. it's communicating to the engine that the task will use no more than `maxCpu` cpus, so it is free to limit the provisioned resources. But maybe others have a different view.

SPEC.md
jdidion: This hint is saying that the task is able to handle a sequential read-only input file, i.e., it's not going to perform random access. Thus, the engine is free to e.g. provide a FIFO rather than a real file, and stream the input to that FIFO.

SPEC.md
jdidion: thx

jdidion: yes :)

Maybe we need a workflow requirements section as well/instead?

jdidion: As per discussion in the governance meeting today, this hint now only allows setting at runtime the values of optional inputs that are not explicitly specified in the call. Thus, there is never the case that a task/subworkflow input is unsatisfied, and so the workflow can execute regardless of the value of `allowed_nested_inputs`. I still provided the option for the workflow engine to refuse to execute a workflow if `allowed_nested_inputs` is `false` and the user tries to set a nested value anyway, but I consider that to be different than failing.

SPEC.md
markjschreiber: Could we reword "_Specifies the maximum CPU to be provisioned for a task_" to something like "_A hint to the engine that the task expects to use no more than maxCpu. The engine is free to make any optimization it deems beneficial based on this information_".

SPEC.md
```diff
@@ -4586,31 +4620,62 @@ Example output:
...
 
-##### `maxMemory`
+##### `max_memory`
+
+* Accepted types:
+    * `Int`: Bytes of RAM.
+    * `String`: A decimal value with, optionally with a unit suffix.
+* Alias: `maxMemory`
 
 Specifies the maximum memory provisioned for a task. The value of this hint has the same specification as [`requirements.memory`](#memory).
```
markjschreiber: Same comment for maxCpu.

markjschreiber (changes requested):
Request for minor rewording around the maxCPu andMaxMemory hints.

SPEC.md
jdidion: I updated the language for all the hints using your first sentence as a template. The second sentence I instead put as a general statement at the top level of the hints section.

markjschreiber (approved):

PR #603, opened 2024-01-30
Resolve Contains_key Closes #305

### LINKED ISSUE #305: map contains_key() function
This is related to issue #174 and is a solution suggested in that thread.
It would empower terser more maintainable WDLs if we could check if a map has a certain key. In my use case the map would *USUALLY* be totally empty, but has a very large number of possible keys.
```
input {
  Map[String, MyOverrideStruct] = overrides_map
}
MyOverrideStruct? runtime_override = if contains_key(overrides_map, "MyTask") then overrides_map["MyTask"] else null;
call MyTask {
 input:
   foo="foo",
  runtime_override=runtime_override
}
```
The use case is that the SV team is building a big pipeline that calls a lot of complex 3rd-party code. It is inevitable that a few difficult samples will cause some tasks to crash due to e.g. out of memory error. To enable pushing these difficult samples through the pipeline, we pass structures with optional overrides to the task, allowing us to bump up memory, increase disk size, etc. However this induces a scaling problem that the overrides for *EVERY* relevant task must be passed through sub-workflows, making our WDLs look mainly like a long list of optional overrides. The present alternative would be to create an enormous struct of overrides, which could at least hide some of the mess, but still presents an annoying maintenance problem: if someone renames tasks, adds them, removes them then the struct file needs to be changed too.

Draft implementation: https://github.com/openwdl/wdl/tree/305-contains-key

vsmalladi:
### Checklist


SPEC.md
```diff
@@ -9209,6 +9210,165 @@ Example output:
 </p>
 </details>
 
+### ✨ `contains_key`
+
+```
+Map[P, Y], P), Boolean contains_key(Map[P?, Y], P?)
```
jdidion: Looks like a bad merge here. Check the original diff: https://github.com/openwdl/wdl/compare/cf597dc07eb88d2fe727bbeb10473569f7b412d8..1af834fa053360f9c84dea734abcfc6bf202bb0c

jdidion (changes requested):

SPEC.md
```diff
@@ -9209,6 +9210,96 @@ Example output:
 </p>
 </details>
 
+### ✨ `contains_key`
+
+```
+Map[P, Y], Boolean contains_key(Map[P?, Y], P?)
```
jdidion: Still something weird going on here. See for e.g. `min` and `max` for other functions with multiple variants.

Should be:

```
* Boolean contains_key(Map[P, Y], P)
* Boolean contains_key(Object, String)
* Boolean contains_key(Map[String, Y]|Struct|Object, Array[String])
```

jdidion (changes requested):
Still something messed up with the signature codeblock. Other than that lgtm.

jdidion (approved):

PR #606, opened 2024-02-05
Runtime access to requirements and meta

jdidion:
### Checklist


SPEC.md
```diff
@@ -5286,6 +5295,97 @@ Example output:
 </p>
 </details>
 
+### Runtime Access to Requirements, Hints, and Metadata
+
+The `requirements` and `hints` sections comprise resource requests to the execution engine. But these requests can be [specified or overridden at runtime](#specifying--overriding-requirements-and-hints), and the execution engine has some latitude in whether and how it fulfills them. Thus, the workflow developer may wish to know exactly what resources are available at runtime, such as:
+
+* What are the actual resource allocations. For example, a task may request at least `8 GB` of memory but may be able to use more memory if it is available.
```
markjschreiber: Nit - I think we should prefer `GiB` in most places in the spec as we often see people using `GB` and then wondering why they don't get enough memory.

SPEC.md
```diff
@@ -5286,6 +5295,97 @@ Example output:
...
+* `id`: A `String` with the unique ID of the task. The execution engine may choose the format for this ID, but it is suggested to include at least the following information:
+    * The task name
+    * The task alias, if it differs from the task name
+    * The index of the task instance, if it is within a scatter statement
+* `container`: The URI `String` of the container in which the task is executing, or `None` if the task is being executed in the host environment. 
+* `cpu`: The allocated number of cpus as a `Float`. Must be greater than `0`.
+* `memory`: The allocated memory in bytes as an `Int`. Must be greater than `0`.
+* `gpu`: An `Array[String]` with one specification per allocated GPU. The specification is execution engine-specific. If no GPUs were allocated, then the value must be an empty array.
+* `fpga`: An `Array[String]` with one specification per allocated FPGA. The specification is execution engine-specific. If no FPGAs were allocated, then the value must be an empty array.
+* `disks`: A `Map[String, Int]` with one entry for each disk mount point. The key is the mount point and the value is the amount of disk space allocated in bytes.
```
markjschreiber: Should this be just for the mount points requested in the task definition or actually all of the mount points the container or process can see with something like `df`. My preference would be for the latter because other mount points may have special purposes and shouldn't be used by the task.

SPEC.md
```diff
@@ -5286,6 +5295,97 @@ Example output:
 </p>
 </details>
 
+### Runtime Access to Requirements, Hints, and Metadata
```
markjschreiber, line 5298: Should we put in a recommendation that if these values are used in a way that significantly modifies the behavior of a task then it may lead to a workflow that will have reduced portability?

markjschreiber (commented):
A couple of comments with suggestions for clarification. Overall I'm a bit worried that people may be tempted to use these values in ways that significantly modify a tasks behavior and that doing that would make a workflow less portable. We should probably warn against this.

SPEC.md
jdidion: I think it should be the former. Given the strong suggestion to use a `container` (which becomes a requirement in WDL 2.0), this map should just contain the volumes that are mounted when starting the container.

But I'd also be fine with leaving it more vague - that it needs to contain at least the mount points requested but may contain additional.

SPEC.md
jdidion: I'm good with that. But perhaps you can give some examples of what types of usages you're worried about?

markjschreiber: The use of metadata in some Cromwell workflows to specify things that would be hints in WDL 1.2 comes to mind. We don't want people to try and use metadata as a substitute for hints and change behavior accordingly.

SPEC.md
markjschreiber: Ah, I missed the part where it was mount points only and not volumes in the container. I think that is mainly OK. There are a few edge cases I can think of. In the case of WDL on HealthOmics we mount a disk at `/mnt/workflow` but in general only the specific working directory of the task is `r/w` other places in this mount are read only by the task, some locations are read and write only by the engine. I would prefer people write to the working directory using a relative path rather than using absolute paths to the mount (or assuming the whole mount is available). 

We also have some implementations where a mount size can increase dynamically so potentially relying on this figure might break something.

These might both also be good cases where making too many decisions in a task based on these values might break between implementations

jdidion: Oh good point about autoscaling volumes. I think I should update it to say: at a minimum the map must contain the *initial* capacities of the requested volumes. A disk's capacity may increase (e.g. autoscaling) so if the user needs accurate numbers they should determine them by calling the appropriate system command.

SPEC.md
vsmalladi: We should just not allow it right? And force that behavior into hints?

jdidion: The metadata sections are free-form, but they should definitely not be used for hints, especially after we introduce a dedicated `hints` section. That said, I'm not sure we can force the Cromwell folks to do/not do anything, but we can strongly recommend against it in the spec.

markjschreiber: Agreed, we don't want this to become a way to "officially" implement what have to date been workarounds or engine specific optimizations

SPEC.md
```diff
@@ -5286,6 +5295,97 @@ Example output:
...
+
+The `requirements` and `hints` sections comprise resource requests to the execution engine. But these requests can be [specified or overridden at runtime](#specifying--overriding-requirements-and-hints), and the execution engine has some latitude in whether and how it fulfills them. Thus, the workflow developer may wish to know exactly what resources are available at runtime, such as:
+
+* What are the actual resource allocations. For example, a task may request at least `8 GiB` of memory but may be able to use more memory if it is available.
+* The task metadata, to avoid duplication. For example, the task may wish to write log messages with the task's name and description without having to duplicate the information in the task's `meta` section.
+* The runtime engine may also choose to provide additional information at runtime.
+
+This information is provided by the `task` variable, which is implicitly defined by the execution engine. The type of `task` is a [scoped type](#scoped-types) with the following members:
+
+* `name`: The task name.
```
rhpvorderman, line 5308: Should we add `time` to the list? This is relevant for SLURM. I guess it also matters for the cloud backends. Can somebody who uses cloud backends comment on this? If it is only relevant for SLURM then there is no need to add it.

markjschreiber: As in the maximum runtime for a task? It could be useful to prevent tasks getting stuck and running much longer than expected but it might be something that isn't very portable. A task that should take no more than 15 minutes in one environment might legitimately take 40 minutes in a different environment. It might be something that would be more useful if added as a hint for that environment?

rhpvorderman: Yes, you are right. A hint is much better.

jdidion: This PR is about what information is both useful and can be provided at runtime, regardless if it originates from a requirement, hint, or neither (I.e. an inherent property of the task or the runtime environment).

If time (start time? max allowed runtime? Both? What are the units/type/format?) is useful then we should include it.

vsmalladi: Max run time might be useful in Seconds. But that might be more of a hint.

vsmalladi (approved):

markjschreiber (approved):

SPEC.md
jdidion: I added the following: `end_time`: An `Int?` whose value is the time by which the task must be completed, as a [Unix time stamp](https://en.wikipedia.org/wiki/Unix_time). A value of `0` means that the execution engine does not impose a time limit. A value of `None` means that the execution engine cannot determine whether the runtime of the task is limited. A positive value is a guarantee that the task will be preempted at the specified time, but is *not* a guarantee that the task won't be preempted earlier.

My reasoning is that "max time" is less precise because there could be a delay between when the clock starts and when the execution starts, whereas with a fixed point in time both the execution engine and the executing task will be on the same page as to when the task will be killed. The max runtime can be (approximately) back-calculated by subtracting the current time from `end_time` as the first step of the task.

Thoughts?

PR #615, opened 2024-02-10
Clarify behavior of read_* and some write_* functions when the input is empty

jdidion:
The general principal is to return an empty value/write an empty file when possible, and raise an error otherwise.

### Checklist


rhpvorderman (approved):
This is a very nice clarification.

PR #617, opened 2024-02-10
Replace emojis with their text codes

jdidion:
This fixes linking behavior in IDEs and leaves behavior in GitHub unaffected.

Note that this will break any direct links to section headings that contain emojis!

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,8 @@ version 1.1.2
 
 + Remove some syntax sections that were missed in 1.1.1.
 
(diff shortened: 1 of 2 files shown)


rhpvorderman:
Emojis are valid unicode, text codes are implementation dependent. So I am more inclined to file a bug with the IDEs rather than fix it here. I would not care very much, if it wasn't for the breaking of third-party links. That is not a good idea for a specification in my opinion.

jdidion:
I think I agree. I have found this annoying for a long time and finally decided to try to fix it. But I think the fix is more negative than positive.

PR #618, opened 2024-03-28
Add `values` function for getting values from a `Map`

jdidion:
I also added the `Array[String] keys(Struct|Object)` function variant for getting the member names for a struct or object. I think this is useful, but also necessary to be consistent with the signatures of the `contains_key` function.

Also fixed the example for `contains_key`.

Also added the missing entry for multi-line strings to the changelog.

### Checklist

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -10522,7 +10522,7 @@ There is no natural or unambiguous serialization of a `Map` with a non-`String` 
 
 #### Map to Struct
 
-A `Map[X, Y]` can be converted to a `Struct` with two array members: `Array[X] keys` and `Array[Y] values`. This is the suggested approach.
+A `Map[P, Y]` can be converted to a `Struct` with two array members: `Array[X] keys` and `Array[Y] values`. This is the suggested approach.
 
 <details>
 <summary>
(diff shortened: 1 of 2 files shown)


SPEC.md
```diff
@@ -9883,15 +9884,20 @@ Test config:
 
 ```
 Array[P] keys(Map[P, Y])
+Array[String] keys(Struct|Object)
 ```
 
-Creates an `Array` of the keys from the input `Map`, in the same order as the elements in the map.
+Given a key-value type collection (`Map`, `Struct`, or `Object`), returns an `Array` of the keys from the input collection, in the same order as the elements in the collection.
```
markjschreiber, line 10001: If the key of the input is optional then does the returned Array contain a concrete String?

For example:

```
Struct Name {
    last: String,
    first: String,
    middle: String?
}

...
Name name = {
    first: "John",
    last: "Doe"
}

Array[String] the_keys = keys(name).
```
Does `the_keys` contain the `String middle` or not?

markjschreiber (commented):
How should optional keys be handled when they are not set?

SPEC.md
jdidion: Yes, for a struct all members are always present, and optional members default to `None` if a value is not specified.

markjschreiber (approved):
LGTM

PR #619, opened 2024-03-28
Clarify that `read_boolean` is case-insensitive

jdidion:
The example already implied that this was the case. I just made it explicit in the function description.

### Checklist

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -7037,7 +7037,7 @@ Example output:
 Boolean read_boolean(File)
 ```
 
-Reads a file that contains a single line containing only a boolean value and (optional) whitespace. If the line contains "true" or "false", that value is returned as a `Boolean`. If the file is empty or does not contain a single boolean, an error is raised. The comparison is case- and whitespace-insensitive.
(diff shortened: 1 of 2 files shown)


vsmalladi (approved):
LTGM

SPEC.md
```diff
@@ -7037,7 +7037,7 @@ Example output:
 Boolean read_boolean(File)
 ```
 
-Reads a file that contains a single line containing only a boolean value and (optional) whitespace. If the line contains "true" or "false", that value is returned as a `Boolean`. If the file is empty or does not contain a single boolean, an error is raised.
+Reads a file that contains a single line containing only a boolean value and (optional) whitespace. If the line contains "true" or "false", that value is returned as a `Boolean`. If the file is empty or does not contain a single boolean, an error is raised. The comparison is case- and whitespace-insensitive.
```
illusional: This says _contains_, what does that mean for `truefalse` or `this is true` or `false or true`? Or should we change it to "exactly" true or false whitespace and case insensitive?

jdidion: Thanks - yes the intention is just true/false plus optional whitespace. I'll clarify that.

jdidion:
```suggestion
Reads a file that contains a single line containing only a boolean value and (optional) whitespace. If the non-whitespace content of the line is "true" or "false", that value is returned as a `Boolean`. If the file is empty or does not contain a single boolean, an error is raised. The comparison is case- and whitespace-insensitive.
```

PR #620, opened 2024-03-28
Generalize the `length` function

jdidion:
Added `Map`, `Object`,  `Struct`, and `String` as allowed arguments to `length`. Also moved the `length` function to the "other" section of the standard library since it now accepts more than just an array argument.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,7 +64,7 @@ version 1.2.0
 
 + Added `find`, and `matches` functions.
 
-+ Generalized `length` function to also accept `Map`, `Object`, `Struct`, and `String` arguments.
++ Generalized `length` function to also accept `Map`, `Object`, and `String` arguments.
 
 + Added parameters to `read_tsv` that enable it to read field names from a header row or an `Array[String]` and return an `Array[Object]`. [PR 627](https://github.com/openwdl/wdl/pull/627)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,8 @@ version 1.2.0
 
 + Added `find`, and `matches` functions.
 
++ Generalized `length` function to also accept `Map`, `Object`, `Struct`, and `String` arguments.
+
 version 1.1.1
 ---------------------------
 
--- a/SPEC.md
+++ b/SPEC.md
@@ -168,7 +168,6 @@ Revisions to this specification are made periodically in order to correct errors
     - [`squote`](#squote)
     - [`sep`](#sep-1)
   - [Generic Array Functions](#generic-array-functions)
-    - [`length`](#length)
     - [`range`](#range)
     - [`transpose`](#transpose)
     - [`cross`](#cross)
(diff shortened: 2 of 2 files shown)


SPEC.md
```diff
@@ -10164,6 +10112,83 @@ Example output:
...
+**Returns**: The length of the collection/string as an `Int`.
+
+<details>
+<summary>
+Example: test_length.wdl
+
+```wdl
+version 1.2
+
+struct Name {
```
markjschreiber: Could you make one of the members of `Name` optional so that the example effectively checks that optional members are also counted?

markjschreiber (commented):
LGTM.

Suggested addition to the example to verify that optional struct members are counted.

markjschreiber (approved):
Ship it!

SPEC.md
```diff
@@ -10164,6 +10112,84 @@ Example output:
...
+
+Example output:
+
+```json
+{
+  "test_length.xlen": 3,
+  "test_length.ylen": 3,
+  "test_length.zlen": 0,
+  "test_length.mlen": 2,
+  "test_length.slen": 2,
```
vsmalladi: Shouldn't `test_length.slen` be `5` and `test_length.name_len` be `2`?

vsmalladi (changes requested):
Seems like some issues in the tests

SPEC.md
wleepang: +1

jdidion: Yes `slen` should be `5`. For `name_length`, since `name` is a struct instance, all members are always present; an optional member whose value is not initialized is set to `None`.

I'm not sure if allowing struct arguments to `contains_key` and `length` is actually useful. I thought it made sense to add it since `Object` is supported, but maybe it just adds more confusion than it's worth.

vsmalladi: I think it might just be more confusing if all arguments are always present.

jdidion: Done - removed `Struct` as an allowed argument

vsmalladi (approved):
LGTM

SPEC.md
```diff
@@ -10164,6 +10112,78 @@ Example output:
...
+
+Returns the length of the input argument as an `Int`:
+
+* For an `Array[X]` argument: the number of elements in the array.
+* For a `Map[X, Y]` argument: the number of items in the map.
+* For an `Object` argument: the number of key-value pairs in the object.
+* For a `String` argument: the number of characters in the string.
+
+Note that every struct of a given type will always have the same length, equal to the number of members in the struct definition, regardless of whether any of the members are optional.
+
```
DavyCats: Since struct's are not allowed as an argument, this comment seems unnecessary.

SPEC.md
```diff
@@ -10164,6 +10112,78 @@ Example output:
...
+Example: test_length.wdl
+
+```wdl
+version 1.2
+
+struct Name {
+  String first
+  String last
+  String? middle
+}
```
DavyCats: This struct definition is also not getting used anymore.

SPEC.md
jdidion: good catch, thanks

SPEC.md
jdidion: fixed

CHANGELOG.md
```diff
@@ -64,6 +64,8 @@ version 1.2.0
 
 + Added `find`, and `matches` functions.
 
++ Generalized `length` function to also accept `Map`, `Object`, `Struct`, and `String` arguments.
```
DavyCats:
```suggestion
+ Generalized `length` function to also accept `Map`, `Object`, and `String` arguments.
```

DavyCats (approved):
Just one more stray struct mention. Otherwise LGTM

PR #623, opened 2024-03-28
Add `chunk` function for splitting an array into equal-sized chunks

jdidion:
### Checklist

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -174,6 +174,7 @@ Revisions to this specification are made periodically in order to correct errors
     - [`cross`](#cross)
     - [`zip`](#zip)
     - [`unzip`](#unzip)
+    - [✨ `chunk`](#-chunk)
     - [`flatten`](#flatten)
     - [`select_first`](#select_first)
     - [`select_all`](#select_all)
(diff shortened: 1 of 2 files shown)


SPEC.md
```diff
@@ -9481,6 +9482,72 @@ Example output:
...
+
+```wdl
+version 1.2
+workflow chunk_array {
+  Array[String] s1 = ["a", "b", "c", "d", "e", "f"]
+  Array[String] s2 = ["a", "b", "c", "d", "e"]
+  Array[String] s3 = ["a", "b"]
+  Array[String] s4 = []
+  
+  scatter (a in split(s1, 3)) {
```
markjschreiber: Not sure if this scatter block is needed to demonstrate the chunk function?

markjschreiber (changes requested):
Is the scatter at line 9514 required?

SPEC.md
wleepang: should this be:
```
scatter (a in chunk(s1, 3)) { ...
```

jdidion: yes thanks

jdidion: I'm open to suggestions if you think another example would be better.

markjschreiber (approved):

SPEC.md
```diff
@@ -9592,6 +9593,72 @@ Example output:
...
 
+### ✨ `chunk`
+
+```
+Array[Array[X]] chunk(Array[X], Int)
+```
+
+Given an array and a length *n*, splits the array into consecutive, non-overlapping arrays of *n* elements. If the length of the array is not a multiple *n* then the final sub-array will have `length(array) % n` elements.
+
+This function is the inverse of [`flatten`](#flatten); i.e., `flatten`ing a `chunk` array will always yield the original array.
```
illusional: I'm meh on this line, I don't know if it adds much.

SPEC.md
```diff
@@ -9592,6 +9593,72 @@ Example output:
...
+
+**Parameters**
+
+1. `Array[X]`: The array to split. May be empty.
+2. `Int`: The desired length of the sub-arrays. Must be > 0.
+
+**Returns**: An array of sub-arrays, where each sub-array is the same length except possibly the last one.
```
illusional:
```suggestion
**Returns**: An array of sub-arrays, where each sub-array is of length `N` except possibly the last one.
```

patmagee (commented):
@jdidion what is the behaviour if the number of requested chunks is greater then the length of the array? An error?

jdidion:
@patmagee the parameter is the requested chunk size, not the number of chunks. And it is stated that the last chunk can be shorter than the requested size. So e.g. if the array is of length 1 and I request a chunk size of 5, then the returned value will be an array of one array that contains one element.

SPEC.md
jdidion: Removed

PR #624, opened 2024-03-28
Add `join_paths` function to correctly join file paths

jdidion:
### Checklist

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -7547,7 +7547,7 @@ Example output:
 ### ✨ `join_paths`
 
 ```
-File join_paths(File|String, String)
+File join_paths(File, String)
 File join_paths(File, Array[String]+)
 File join_paths(Array[String]+)
 ```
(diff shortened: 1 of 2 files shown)


SPEC.md
```diff
@@ -7367,6 +7368,93 @@ Example output:
 </p>
 </details>
 
+### ✨ `join_paths`
+
+```
+File join_paths(File|String, String)
```
markjschreiber: This probably depends on the rules of casting between `File` and `Directory` but should the definitions be

```
File join_paths(Directory|String, String)
```
etc

markjschreiber (commented):
Question about `File` vs `Directory` in the function definition.

SPEC.md
jdidion: We haven't added `Directory` support to 1.2 yet :)

markjschreiber (approved):
LGTM. One recommendation that we could make is that engines should normalize the final path produced by `join_paths` because when a workflow output location is specified by using `join_paths` and the output will be going to `s3` or `gs` or similar object stores then `foo//baa/./baz` is not the same as `foo/baa/baz` even though they are when the output location is a POSIX compliant FS.

SPEC.md
```diff
@@ -7367,6 +7368,93 @@ Example output:
...
+File join_paths(File, Array[String]+)
+File join_paths(Array[String]+)
+```
+
+Joins together two or more paths into an absolute path in the host filesystem.
+
+There are three variants of this function:
+
+1. `File join_paths(File|String, String)`: Joins together exactly two paths. The first path may be either absolute or relative and must specify a directory; the second path is relative to the first path and may specify a file or directory.
```
illusional: If File is coercible to a String, is this needed, or just helpful to imply?

SPEC.md
```diff
@@ -7367,6 +7368,93 @@ Example output:
...
+2. `File join_paths(File, Array[String]+)`: Joins together any number of relative paths with a base path. The first argument may be either an absolute or a relative path and must specify a directory. The paths in the second array argument must all be relative. The *last* element may specify a file or directory; all other elements must specify a directory.
+3. `File join_paths(Array[String]+)`: Joins together any number of paths. The array must not be empty. The *first* element of the array may be either absolute or relative; subsequent path(s) must be relative. The *last* element may specify a file or directory; all other elements must specify a directory.
+
+An absolute path starts with `/` and indicates that the path is relative to the root of the container in which the task is executed. A relative path does not start with `/` and indicates the path is relative to its parent directory. It is up to the execution engine to determine which directory to use as the parent when resolving relative paths; by default it is the working directory in which the task is executed.
```
illusional: Do we want to prescribe the behaviour of supplying an absolute path as a later parameter, eg: Python will ignore all previous paths - or should this be an exception?

SPEC.md
```diff
@@ -7367,6 +7368,93 @@ Example output:
...
+    File rel_dir_file = "mydir"
+    String rel_str = "mydata.txt"
+  }
+
+  # these are all equivalent to '/usr/bin/echo'
+  File bin1 = join_paths(abs_file, [rel_str_dir, rel_file])
+  File bin2 = join_paths(abs_str, [rel_str_dir, rel_file])
+  File bin3 = join_paths([abs_str, rel_str_dir, rel_file])
+  
+  # this resolves to '<working dir>/mydir/mydata.txt'
```
illusional: Depends how pedantic the spec should be, but the above states:

> It is up to the execution engine to determine which directory to use as the parent when resolving relative paths;

So maybe "the default behaviour is this should resolve to ..."?

SPEC.md
jdidion: Good point. `String` is coercible to `File` so we only need `File`.

SPEC.md
jdidion: I don't really like that behavior in python tbh. IMO if we're going to the principle of least surprising behavior, then that should be an error.

illusional (approved):
Awesome!! Thanks for driving this :)

PR #625, opened 2024-03-28
Add `parameter_meta` to struct definition

jdidion:
### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,8 @@ version 1.2.0
 
 + Added `find`, and `matches` functions.
 
++ Add `parameter_meta` section to struct definition.
+
 version 1.1.1
 ---------------------------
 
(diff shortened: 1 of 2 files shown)


vsmalladi:
@jdidion should this be merged to the 1.2 branch?

jdidion:
@vsmalladi yes fixed now

illusional (commented):
Is it an error if a parameter_meta contains extra fields?

jdidion:
@illusional yes. In the task and workflow `parameter_meta` sections it says "This section contains metadata specific to input and output parameters. Any key in this section *must* correspond to a task input or output." So I think it's only consistent to put a similar requirement on `struct` `parameter_meta`.

patmagee (approved):
This looks god to me, Straight forward.

@jdidion For completeness do you think we should also allow a meta section? 

Also IIRC, meta and parameter_meta can both reference expressions (or maybe I am misremember). Are expressions allowed? Can they reference the values within

jdidion:
@patmagee yes good point - no reason not to also allow `meta`.

Neither `meta` nor `parameter_meta` allow expressions. You might be thinking of `hints`, which we originally proposed would be like `meta` sections and not allow expressions, but have decided to instead make it like `requirements` and allow expressions.

patmagee (approved):
LGTM

PR #626, opened 2024-03-28
Remove restriction on extra keys when coercing Map/Object to struct

jdidion:
### Checklist

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -1446,10 +1446,9 @@ Example output:
 
 🗑 It is also possible to assign an `Object` or `Map[String, X]` value to a `Struct` declaration. In the either case:
 
-* The value of each object/map member must be coercible to the declared type of the struct member.
+* The value of each  `Object`/`Map` member must be coercible to the declared type of the struct member.
 * The `Object`/`Map` must at least contain values for all of the struct's non-optional members.
-
-Keys in the map/object that do not match any struct members are ignored.
+* A key in the `Object`/`Map` that does not correspond to a member of the struct is ignored.
 
 Note that the ability to assign values to `Struct` declarations other than struct literals is deprecated and will be removed in WDL 2.0.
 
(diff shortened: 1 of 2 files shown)


SPEC.md
```diff
@@ -1442,10 +1442,11 @@ Example output:
 
 🗑 It is also possible to assign an `Object` or `Map[String, X]` value to a `Struct` declaration. In the either case:
 
-* The `Object`/`Map` must not have any members that are not declared for the struct.
 * The value of each object/map member must be coercible to the declared type of the struct member.
 * The `Object`/`Map` must at least contain values for all of the struct's non-optional members.
 
+Keys in the map/object that do not match any struct members are ignored.
```
markjschreiber: Is the cast destructive? For example if I cast a Map to a struct do the keys from the map that don't belong to the struct disappear from the Map or are they just inaccessible from the Struct pointer? What if I cast back again? Do they re-appear?

markjschreiber (changes requested):
I think it would be good to have more clarity on what happens to the keys (and pairs) that don't match.

SPEC.md
jdidion: The new struct value is a copy, so it is not destructive in that sense. If you cast the Map to a struct without first saving the Map to a variable then it's gone. Casting from struct back to Map will not restore those extra members.

jdidion: It's a good point though - the spec is not clear on whether type conversion is destructive. I'll add some clarification in that seciton.

markjschreiber (approved):

SPEC.md
illusional: Maybe saying _Only keys in the map / object that match struct members are stored in the struct_. But agree, the example Map -> Struct -> Map will not have the original fields.

illusional (approved):

PR #633, opened 2024-04-08
Add array `contains` function

jdidion:
### Checklist


patmagee:
@jdidion do we need to make any comments here on how equality works and whether type coerced objects are equal?

Ie would the following be truthy?

```wdl

File Foo = "foo"
Array[String] bar = ["foo"]
```

jdidion:
@patmagee yes type coercion should apply here. `String` -> `File` is one of the officially supported coercions, but not `File` -> `String`.

```
Array[File] files = ["/foo/bar"]
contains(files, "/foo/bar")  # true

Array[String] strings = ["/foo/bar"]
File foobar = "/foo/bar"
contains(strings, foobar)  # error - `foobar` is not of type `String` and is not coercible to `String`
contains(strings, "~{foobar}")  # true
```

vsmalladi (approved):
LGTM

Did you want to add that example you ad in the discussion?

PR #650, opened 2024-05-23
Merge WDL 1.1.2 changes, fix broken links

jdidion:
I used `lychee --include-fragments` to find broken links/anchors.

I removed `development` from the changelog since those changes should only be reflected in the `development` branch.

I alphabetized the list of governance members and added @wleepang. We should review the list again and make sure all the people listed are still considered (and still want to be) part of governance, and that we're not missing anyone.


vsmalladi (approved):
LGTM

PR #660, opened 2024-06-06
Use `echo` instead of `cat` for number that isn't a file path

### LINKED ISSUE #654: WDL comment test does not run due to passing in a nonpath string into cat
https://github.com/openwdl/wdl/blob/fc24d91a11355ffb74961704ea255f74986f7913/SPEC.md?plain=1#L422

This should probably be `echo` instead of `cat`

stxue1:
Closes #654

`~{number * 2}` outputs a number, but `cat ~{number *2}` tries to read the file named that number, which doesn't work as that file does not exist. This test isn't testing reading from a file but from stdout, so `echo` should be more appropriate in printing the number to stdout.

I didn't add anything to the changelog as this is more of a bug fix.

### Checklist

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -419,7 +419,7 @@ There is no special syntax for multi-line comments - simply use a `#` at the sta
     # This comment will not be included within the command
     command <<<
       # This comment WILL be included within the command after it has been parsed
-      cat ~{number * 2}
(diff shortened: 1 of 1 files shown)


jdidion (approved):
Rebasing to `wdl-1.1.3`

PR #669, opened 2024-06-26
Fix many issues with examples

### LINKED ISSUE #653: Example WDL test for gpu's does not work depending on default executor container
In the `test_gpu_task.wdl` [example test](https://github.com/openwdl/wdl/blob/fc24d91a11355ffb74961704ea255f74986f7913/SPEC.md?plain=1#L4226-L4244), it depends on the [`lspci` command](https://github.com/openwdl/wdl/blob/fc24d91a11355ffb74961704ea255f74986f7913/SPEC.md?plain=1#L4233) from the [default executor container](https://github.com/openwdl/wdl/blob/fc24d91a11355ffb74961704ea255f74986f7913/SPEC.md?plain=1#L4029) as [no container is specified](https://github.com/openwdl/wdl/blob/fc24d91a11355ffb74961704ea255f74986f7913/SPEC.md?plain=1#L4240-L4242).

For MiniWDL and Toil, the default container is Ubuntu 20.04:
https://github.com/chanzuckerberg/miniwdl/blob/a34aa902ec1be2f411db1e38daf51bae4f839f42/WDL/runtime/config_templates/default.cfg#L144-L146

But the base Ubuntu 20.04 doesn't support `lspci`:
```
$ docker run -it ubuntu:20.04 bash
root@d3b58db61322:/# lspci
bash: lspci: command not found
```

I think some container with `lspci` preinstalled should be specified. Ex:
```Dockerfile
FROM ubuntu:22.04

RUN apt-get -y update && apt-get -y install pciutils
```
and
```wdl
  runtime {
    gpu: true
    container: "quay.io/stxue/ubuntu_lspci:latest"
  }
```
which allows the test to run:
```
stxue@mustard:/private/groups/patenlab/toil-dev/wdl-conformance-tests$ srun -c 1 --mem 1G --time=0:10:00 --partition gpu --gres=gpu:1 --pty bash -i
stxue@phoenix-05:/private/groups/patenlab/toil-dev/wdl-conformance-tests$ toil-wdl-runner test_gpu_task.wdl 
/usr/lib/python3/dist-packages/paramiko/transport.py:236: CryptographyDeprecationWarning: Blowfish has been deprecated and will be removed in a future release
  "class": algorithms.Blowfish,
[2024-05-29T13:39:17-0700] [MainThread] [I] [toil] Running Toil version 7.1.0a1-e9d0b62d7d62af9672946858fdd6fb587b1dd563 on host phoenix-05.prism.
/usr/lib/python3/dist-packages/paramiko/transport.py:236: CryptographyDeprecationWarning: Blowfish has been deprecated and will be removed in a future release
  "class": algorithms.Blowfish,
[2024-05-29T13:39:18-0700] [MainThread] [I] [toil.leader] 0 jobs are running, 0 jobs are issued and waiting to run
[2024-05-29T13:39:18-0700] [MainThread] [I] [toil.leader] Issued job 'WDLTaskJob' wf.test_gpu.command kind-WDLTaskJob/instance-52hhmugy v1 with job batch system ID: 2 and disk: 2.0 Gi, memor
y: 2.0 Gi, cores: 1, accelerators: [{'count': 1, 'kind': 'gpu'}], preemptible: False
/usr/lib/python3/dist-packages/paramiko/transport.py:236: CryptographyDeprecationWarning: Blowfish has been deprecated and will be removed in a future release
  "class": algorithms.Blowfish,
/usr/lib/python3/dist-packages/paramiko/transport.py:236: CryptographyDeprecationWarning: Blowfish has been deprecated and will be removed in a future release
  "class": algorithms.Blowfish,
[2024-05-29T13:39:25-0700] [MainThread] [I] [toil.leader] Finished toil run successfully.

Workflow Progress 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 3/3 (0 failures) [00:06<00:00, 0.48 jobs/s]
{"wf.out": true}
[2024-05-29T13:39:25-0700] [MainThread] [I] [toil.common] Successfully deleted the job store: FileJobStore(/data/tmp/tmpxfvwtsc1/tree)
```
```
stxue@phoenix-05:/private/groups/patenlab/toil-dev/wdl-conformance-tests$ MINIWDL__SCHEDULER__CONTAINER_BACKEND=singularity miniwdl run test_gpu_task.wdl 
2024-05-29 13:40:34.114 wdl.w:wf workflow start :: name: "wf", source: "test_gpu_task.wdl", line: 2, column: 1, dir: "/private/groups/patenlab/toil-dev/wdl-conformance-tests/20240529_134034_wf"
2024-05-29 13:40:34.118 wdl.w:wf miniwdl :: version: "v1.12.0", uname: "Linux phoenix-05.prism 5.15.0-78-generic #85-Ubuntu SMP Fri Jul 7 15:25:09 UTC 2023 x86_64"
2024-05-29 13:40:34.146 wdl.w:wf ready :: job: "call-test_gpu", callee: "test_gpu"
2024-05-29 13:40:34.147 wdl.w:wf.t:call-test_gpu task setup :: name: "test_gpu", source: "test_gpu_task.wdl", line: 9, column: 1, dir: "/private/groups/patenlab/toil-dev/wdl-conformance-tests/20240529_134034_wf/call-test_gpu", thread: 139860603147840
    0:00:00 elapsed    ▬▬▬        tasks finished: 0, ready: 1, running: 0/usr/lib/python3/dist-packages/paramiko/transport.py:236: CryptographyDeprecationWarning: Blowfish has been deprecated and will be removed in a future release
  "class": algorithms.Blowfish,
2024-05-29 13:40:34.369 wdl.w:wf.t:call-test_gpu Singularity runtime initialized (BETA) :: singularity_version: "singularity-ce version 3.10.3"
2024-05-29 13:40:34.377 wdl.w:wf.t:call-test_gpu singularity run :: pid: 3168507, log: "/private/groups/patenlab/toil-dev/wdl-conformance-tests/20240529_134034_wf/call-test_gpu/singularity.log.txt"
2024-05-29 13:40:34.651 wdl.w:wf.t:call-test_gpu done
2024-05-29 13:40:34.651 wdl.w:wf finish :: job: "call-test_gpu"
2024-05-29 13:40:34.652 wdl.w:wf done
{
  "dir": "/private/groups/patenlab/toil-dev/wdl-conformance-tests/20240529_134034_wf",
  "outputs": {
    "wf.out": true
  }
}
```

### LINKED ISSUE #654: WDL comment test does not run due to passing in a nonpath string into cat

### LINKED ISSUE #661: Expected output in test_struct.wdl unit test does not exist
The `test_struct.wdl` unit test has an expected output of [`test_struct.person`](https://github.com/openwdl/wdl/blob/9c0b9cf4586508a9e6260cc5c5e562e21f625aac/SPEC.md?plain=1#L1151), but the workflow itself does not have an output variable named `person`:
https://github.com/openwdl/wdl/blob/9c0b9cf4586508a9e6260cc5c5e562e21f625aac/SPEC.md?plain=1#L1124-L1136

I think instead of expecting `test_struct.person`, the right output is `test_struct.john`, as that output from the workflow is missing in the expected output json.

### LINKED ISSUE #662: No output section in map_to_struct.wdl unit test
The `map_to_struct.wdl` unit test has no `output` section:
https://github.com/openwdl/wdl/blob/9c0b9cf4586508a9e6260cc5c5e562e21f625aac/SPEC.md?plain=1#L1422-L1440

For most runners, having no output section will result in no output. For example:
```
(venv3.12) heaucques@pop-os:~/Documents/wdl/wdl-tests/scripts/tests$ miniwdl run no_output.wdl
2024-06-17 14:37:58.328 wdl.w:wf workflow start :: name: "wf", source: "no_output.wdl", line: 3, column: 1, dir: "/home/heaucques/Documents/wdl/wdl-tests/scripts/tests/20240617_143758_wf"
2024-06-17 14:37:58.329 wdl.w:wf miniwdl :: version: "v1.12.0", uname: "Linux pop-os 6.8.0-76060800daily20240311-generic #202403110203~1715181801~22.04~aba43ee SMP PREEMPT_DYNAMIC Wed M x86_64"
2024-06-17 14:37:58.344 wdl.w:wf done
{
  "dir": "/home/heaucques/Documents/wdl/wdl-tests/scripts/tests/20240617_143758_wf",
  "outputs": {}
}
```

I'm unsure if the spec says execution engines should be able to output things when there is no output section in a workflow, but for consistency with the other unit tests, I think [the struct declarations](https://github.com/openwdl/wdl/blob/9c0b9cf4586508a9e6260cc5c5e562e21f625aac/SPEC.md?plain=1#L1428-L1439) should be wrapped in an output section.

### LINKED ISSUE #663: Missing output variable in python_strip_task.wdl unit test
The `python_strip_task.wdl` unit test's output has the variable [`lines`](https://github.com/openwdl/wdl/blob/9c0b9cf4586508a9e6260cc5c5e562e21f625aac/SPEC.md?plain=1#L3618), but the expected json only has the namespace identifier:
https://github.com/openwdl/wdl/blob/9c0b9cf4586508a9e6260cc5c5e562e21f625aac/SPEC.md?plain=1#L3640

I think the proper name should be `python_strip.lines`

### LINKED ISSUE #664: Missing input in allow_nested.wdl unit test
For the `allow_nested.wdl` unit test, these lines call a task and omits a required input:
https://github.com/openwdl/wdl/blob/9c0b9cf4586508a9e6260cc5c5e562e21f625aac/SPEC.md?plain=1#L5655-L5658

[This line](https://github.com/openwdl/wdl/blob/9c0b9cf4586508a9e6260cc5c5e562e21f625aac/SPEC.md?plain=1#L5700) clarifies that with `allowNestedInputs` directive, this input can be supplied elsewhere:
> `repeat2.i` may be set by the caller of the workflow, e.g., by including `"allow_nested.repeat2.i": 2,` in the input JSON.

However, for the unit test, I don't see this input being passed through:
https://github.com/openwdl/wdl/blob/9c0b9cf4586508a9e6260cc5c5e562e21f625aac/SPEC.md?plain=1#L5677-L5685

Instead, there is an expected output that wants the same variable:
https://github.com/openwdl/wdl/blob/9c0b9cf4586508a9e6260cc5c5e562e21f625aac/SPEC.md?plain=1#L5693

As I don't see where `repeat2.i` is defined in the workflow/input JSON, it looks like that line in the example output JSON should be moved to the example input JSON; the spec also doesn't say (at least to my knowledge) that the execution engine should output certain task inputs from nested calls when there is a output section.

### LINKED ISSUE #665: Call statements have a syntax error in if_else.wdl unit test
These call statements in the `if_else.wdl` unit test are missing the `input:` statement:
https://github.com/openwdl/wdl/blob/9c0b9cf4586508a9e6260cc5c5e562e21f625aac/SPEC.md?plain=1#L6057
https://github.com/openwdl/wdl/blob/9c0b9cf4586508a9e6260cc5c5e562e21f625aac/SPEC.md?plain=1#L6062

The [spec](https://github.com/openwdl/wdl/blob/9c0b9cf4586508a9e6260cc5c5e562e21f625aac/SPEC.md?plain=1#L5397) says that the only element a call statement can have is the `input:` keyword followed by a list. It looks like the call statements in the unit test are trying to pass in the list of inputs, and is just missing the `input:` keyword.

### LINKED ISSUE #666: Mismatched variable names in test_prefix.wdl unit test
The `test_prefix.wdl` unit test has a workflow output `env_prefixed`:
https://github.com/openwdl/wdl/blob/9c0b9cf4586508a9e6260cc5c5e562e21f625aac/SPEC.md?plain=1#L8093

But the expected output is `env1_prefixed`:
https://github.com/openwdl/wdl/blob/9c0b9cf4586508a9e6260cc5c5e562e21f625aac/SPEC.md?plain=1#L8110

### LINKED ISSUE #667: Undefined behavior for string to file coercion when file does not exist
It's unclear if a string being coerced to a file should error or not depending on if the file exists and that the file is not outputted in the WDL.

For example:
```wdl
version 1.1
workflow test_map {
  File f = "path/to/file"

  output {
  }
}
```
It's unclear whether this workflow should error on `File f` if the file does not exist.

MiniWDL seems to not error on these cases, but only errors if the file is to be outputted, for example:
```wdl
version 1.1

workflow test_map {
  File f = "path/to/file"

  output {
    File out = f
  }
}
```

A lot of the SPEC tests also depend on this behavior, for example:
https://github.com/openwdl/wdl/blob/664adc323f536ab1fac0c41bf3fc85f4f15eb52b/SPEC.md?plain=1#L1190-L1208

In particular, [this section about type coercion](https://github.com/openwdl/wdl/blob/9c0b9cf4586508a9e6260cc5c5e562e21f625aac/SPEC.md?plain=1#L1286-L1288) talks about string to file coercion, but doesn't mention whether the file must exist at the coercion step.

Should execution engines ensure that files must exist at coercion or only at a workflow/task output?

### LINKED ISSUE #668: The provided correct example for stripping leading whitespace is a Python heredoc that does not run
The 1.1 spec gives this as a correct answer for stripping leading whitespace:

```sh
python <<CODE
  with open("/path/to/file") as fp:
    for line in fp:
      if not line.startswith('#'):
        print(line.strip())
CODE
```

https://github.com/openwdl/wdl/blob/9c0b9cf4586508a9e6260cc5c5e562e21f625aac/SPEC.md?plain=1#L3648-L3655

But that Bash command will always fail. The Python script is indented by 2 spaces; Bash does not itself strip leading whitespace inside the heredoc. So Python will raise an `IndentationError` when parsing the script:

```
$ python <<CODE
  with open("/path/to/file") as fp:
    for line in fp:
      if not line.startswith('#'):
        print(line.strip())
CODE
  File "<stdin>", line 1
    with open("/path/to/file") as fp:
IndentationError: unexpected indent
```

The spec goes on to act as if this command can ever succeed, including by giving example output for the task containing it, which can never be obtained.

jdidion:
Closes #653.
Closes #654.
Closes #661.
Closes #662.
Closes #663.
Closes #664.
Closes #665.
Closes #666.
Closes #667.
Closes #668.

### Checklist

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -1,14 +1,15 @@
 # Workflow Description Language (WDL)
 
-This is version 1.1.2 of the Workflow Description Language (WDL) specification. It describes WDL `version 1.1`. It introduces a number of new features (denoted by the ✨ symbol) and clarifications to the [1.0](https://github.com/openwdl/wdl/blob/main/versions/1.0/SPEC.md) version of the specification. It also deprecates several aspects of the 1.0 specification that will be removed in the [next major WDL version](https://github.com/openwdl/wdl/blob/wdl-2.0/SPEC.md) (denoted by the 🗑 symbol).
+This is version 1.1.3 of the Workflow Description Language (WDL) specification. It describes WDL `version 1.1`. It introduces a number of new features (denoted by the ✨ symbol) and clarifications to the [1.0](https://github.com/openwdl/wdl/blob/main/versions/1.0/SPEC.md) version of the specification. It also deprecates several aspects of the 1.0 specification that will be removed in the [next major WDL version](https://github.com/openwdl/wdl/blob/wdl-2.0/SPEC.md) (denoted by the 🗑 symbol).
 
 ## Revisions
 
 Revisions to this specification are made periodically in order to correct errors, clarify language, or add additional examples. Revisions are released as "patches" to the specification, i.e., the third number in the specification version is incremented. No functionality is added or removed after the initial revision of the specification is ratified.
 
-* [1.1.2](https://github.com/openwdl/wdl/tree/release-1.1.2/SPEC.md): 2024-04-12
-* [1.1.1](https://github.com/openwdl/wdl/tree/release-1.1.1/SPEC.md): 2023-10-04
-* [1.1.0](https://github.com/openwdl/wdl/tree/release-1.1.0/SPEC.md): 2021-01-29
+* [v1.1.3](https://github.com/openwdl/wdl/releases/tag/v1.1.3): 2025-01-23
+* [v1.1.2](https://github.com/openwdl/wdl/releases/tag/v1.1.2): 2024-04-15
+* [v1.1.1](https://github.com/openwdl/wdl/releases/tag/v1.1.1): 2023-10-04
+* [v1.1.0](https://github.com/openwdl/wdl/releases/tag/v1.1.0): 2021-01-29
  
 ## Table of Contents
 
@@ -367,7 +368,7 @@ WDL also provides features for implementing more complex workflows. For example,
   
   ```json
   {
-    "hello.all_matches": [["hi_world"], ["hi_pal"]]
+    "hello_parallel.all_matches": [["hi_world"], ["hello"]]
   }
   ```
   </p>
@@ -419,7 +420,7 @@ There is no special syntax for multi-line comments - simply use a `#` at the sta
     # This comment will not be included within the command
     command <<<
       # This comment WILL be included within the command after it has been parsed
-      cat ~{number * 2}
+      echo ~{number * 2}
     >>>
 
     output {
@@ -534,7 +535,12 @@ The following primitive types exist in WDL:
 * A `File` represents a file (or file-like object).
     * A `File` declaration can have a string value indicating a relative or absolute path on the local file system.
     * Within a WDL file, literal values for files may only be local (relative or absolute) paths.
+    * The path assigned to a `File` is not required to be valid unless and until it is accessed.
+      * To read from a file, it must exist and be assigned appropriate permissions.
+      * To write to a file, the parent directory must be assigned appropriate permissions.
     * An execution engine may support other ways to specify [`File` inputs (e.g. as URIs)](#input-and-output-formats), but prior to task execution it must [localize inputs](#task-input-localization) so that the runtime value of a `File` variable is a local path.
+      * Remote files must be treated as read-only.
+      * A remote file is only required to be vaild at the time that the execution engine needs to localize it.
 
 <details>
   <summary>
@@ -656,7 +662,8 @@ An optional declaration has a default initialization of `None`, which indicates 
     "optionals.test_defined": false,
     "optionals.test_defined2": true,
     "optionals.test_is_none": true,
-    "optionals.test_not_none": false
+    "optionals.test_not_none": false,
+    "optionals.test_non_equal": true
   }
   ```
   </p>
@@ -768,7 +775,7 @@ task sum {
   }
   
   command <<<
-  printf ~{sep(" ", ints)} | awk '{tot=0; for(i=1;i<=NF;i++) tot+=$i; print tot}'
+  printf "~{sep(" ", ints)}" | awk '{tot=0; for(i=1;i<=NF;i++) tot+=$i; print tot}'
   >>>
   
   output {
@@ -833,8 +840,8 @@ Example output:
 {
   "non_empty_optional.nonempty1": [0.0],
   "non_empty_optional.nonempty2": [null, 1],
-  "non_empty_optional.nonempty3": [],
-  "non_empty_optional.nonempty4": [0.0]
+  "non_empty_optional.nonempty3": null,
+  "non_empty_optional.nonempty4": [0]
 }
 ```
 </p>
@@ -1148,7 +1155,7 @@ Example output:
 
 ```json
 {
-  "test_struct.person": {
+  "test_struct.john": {
     "name": "John",
     "account": {
       "account_number": "123456",
@@ -1295,14 +1302,17 @@ Example: string_to_file.wdl
 version 1.1
 
 workflow string_to_file {
-  String path1 = "/path/to/file"
-  File path2 = "/path/to/file"
+  input {
+    File infile
+  }
+
+  String path1 = "~{infile}"
 
   # valid - String coerces unambiguously to File
-  File path3 = path1
+  File path2 = path1
 
   output {
-    Boolean paths_equal = path2 == path3
+    Boolean paths_equal = infile == path2
   }
 }
 ```
@@ -1311,7 +1321,9 @@ workflow string_to_file {
 Example input:
 
 ```json
-{}
+{
+  "string_to_file.infile": "hello.txt"
+}
 ```
 
 Example output:
@@ -1424,18 +1436,20 @@ workflow map_to_struct {
   String b = "key"
   String c = "lookup"
 
-  # What are the keys to this Struct?
-  Words literal_syntax = Words {
-    a: 10,
-    b: 11,
-    c: 12
-  }
-
-  # What are the keys to this Struct?
-  Words map_coercion = {
-    a: 10,
-    b: 11,
-    c: 12
+  output {
+    # What are the keys to this Struct?
+    Words literal_syntax = Words {
+      a: 10,
+      b: 11,
+      c: 12
+    }
+  
+    # What are the keys to this Struct?
+    Words map_coercion = {
+      a: 10,
+      b: 11,
+      c: 12
+    }
   }
 }
 ```
@@ -1561,7 +1575,7 @@ task count_lines {
   }
 
   command <<<
-    wc -l ~{write_lines(array)}
+    wc -l < ~{write_lines(array)}
   >>>
   
   output {
@@ -1864,8 +1878,8 @@ Example output:
 {
   "array_map_equality.is_true1": true,
   "array_map_equality.is_true2": true,
-  "array_map_equality.is_false1": true,
-  "array_map_equality.is_false2": true
+  "array_map_equality.is_false1": false,
+  "array_map_equality.is_false2": false
 }
 ```
 </p>
@@ -2048,7 +2062,7 @@ version 1.1
 struct Experiment {
   String id
   Array[String] variables
-  Map[String, Float] data
+  Map[String, String] data
 }
 
 workflow nested_access {
@@ -2081,7 +2095,7 @@ Example input:
       "variables": ["name", "height"],
       "data": {
         "name": "Pinky",
-        "height": 7
+        "height": "10"
       }
     },
     {
@@ -2089,7 +2103,7 @@ Example input:
       "variables": ["name", "weight"],
       "data": {
         "name": "Porky",
-        "weight": 1000
+        "weight": "100"
       }
     }
   ]
@@ -2209,7 +2223,7 @@ Example input:
 {
   "placeholders.start": "h",
   "placeholders.end": "o",
-  "placeholders.input": "hello"
+  "placeholders.instr": "hello"
 }
 ```
 
@@ -2217,7 +2231,8 @@ Example output:
 
 ```json
 {
-  "placeholders.cmd": "grep 'h...o' hello"
+  "placeholders.cmd": "grep 'h...o' hello",
+  "placeholders.s": "4"
 }
 ```
 </p>
@@ -2296,12 +2311,15 @@ Example: placeholder_coercion.wdl
 version 1.1
 
 workflow placeholder_coercion {
-  File x = "/hij"
+  input {
+    File x
+  }
+  String x_as_str = x
   Int? i = None
 
   output {
     Boolean is_true1 = "~{"abc"}" == "abc"
-    Boolean is_true2 = "~{x}" == "/hij"
+    Boolean is_true2 = "~{x}" == x_as_str
     Boolean is_true3 = "~{5}" == "5"
     Boolean is_true4 = "~{3.141}" == "3.141000"
     Boolean is_true5 = "~{3.141 * 1E-10}" == "0.000000"
@@ -2315,7 +2333,9 @@ workflow placeholder_coercion {
 Example input:
 
 ```json
-{}
+{
+  "placeholder_coercion.x": "hello.txt"
+}
 ```
 
 Example output:
@@ -2429,7 +2449,7 @@ Example output:
 
 ```json
 {
-  "flags.num_matches": 2
+  "flags.num_matches": "2"
 }
 ```
 </p>
@@ -2746,7 +2766,7 @@ task greet_person {
   printf "Hello ~{person.name.first}! You have ~{length(assay_array)} test result(s) available.\n"
 
   if ~{defined(person.income)}; then
-    if [ "~{select_first([person.income]).amount}" -gt 1000 ]; then
+    if [ ~{round(select_first([person.income]).amount)} -gt 1000 ]; then
       currency="~{select_first([select_first([person.income]).currency, "USD"])}"
       printf "Please transfer $currency 500 to continue"
     fi
@@ -2764,7 +2784,7 @@ Example input:
 
 ```json
 {
-  "person_struct.person": {
+  "greet_person.person": {
     "name": {
       "first": "Richard",
       "last": "Rich"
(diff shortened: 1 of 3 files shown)


jdidion:
@stxue1 @adamnovak would also appreciate your reviews

SPEC.md
```diff
@@ -534,7 +535,12 @@ The following primitive types exist in WDL:
 * A `File` represents a file (or file-like object).
     * A `File` declaration can have a string value indicating a relative or absolute path on the local file system.
     * Within a WDL file, literal values for files may only be local (relative or absolute) paths.
+    * The path assigned to a `File` is not required to be valid unless and until it is accessed.
+      * To read from a file, it must exist and be assigned appropriate permissions.
+      * To write to a file, the parent directory must be assigned appropriate permissions.
```
adamnovak, line 540: WDL `File` objects are snapshots, right? Any WDL function that writes a `File` returns the written `File` and doesn't take a `File` as input to overwrite.

Or does this mean within a task, if you substitute a `File`'s name into a command that tries to overwrite it, the parent directory that the `File` initially came from should be writable?

SPEC.md
```diff
@@ -1295,14 +1302,17 @@ Example: string_to_file.wdl
 version 1.1
 
 workflow string_to_file {
-  String path1 = "/path/to/file"
-  File path2 = "/path/to/file"
+  input {
+    File infile
+  }
+
+  String path1 = "~{infile}"
```
adamnovak, line 1309: This makes the test now depend on file to string conversion as well as string to file conversion. But I don't think the test framework lets you set up actual file fixtures to make this work a different way.

Or could we just use `hello.txt`, since it's available to use in the input JSON?

SPEC.md
```diff
@@ -2080,16 +2094,14 @@ Example input:
...
+        "name": "Pinky"
       }
     },
     {
       "id": "pig_weight",
       "variables": ["name", "weight"],
       "data": {
-        "name": "Porky",
-        "weight": 1000
+        "name": "Porky"
```
adamnovak: `variables` no longer holds the map keys; we should either cut the extra `variables` entries or change these numbers into quoted strings instead of removing them.

SPEC.md
```diff
@@ -2972,14 +2991,16 @@ workflow import_structs {
...
+  "import_structs.infile": "hello.txt"
+}
 ```
 
 Example output:
 
 ```json
 {
-  "import_structs.bill": 175000
+  "import_structs.bill": 175000.0
```
adamnovak, line 3005: At least under [the interpretation of JSON that JSON Schema uses](https://json-schema.org/understanding-json-schema/reference/numeric#integer), these are the JSON value.

Are we expecting the runner to produce one serialization here and not the other? Or @stxue1 are you doing something that needs to key on actually having a decimal point?

SPEC.md
```diff
@@ -4854,23 +4877,23 @@ version 1.1
...
 
   command <<<
     mkdir index
-    tar -C index -xzf ~{index}
+    tar -C index -xzf ~{index_tar_gz}
     hisat2 \
       -p ~{threads} \
       ~{if defined(max_reads) then "-u ~{select_first([max_reads])}" else ""} \
-      -x index/~{index_id} \
+      -x index/grch38/genome \
```
adamnovak: This looks like we're hardcoding for the name of the genome that actually happens to be in this file; we also no longer use `index_id` at all.

Maybe we meant to generate this string from the filename?

```suggestion
      -x index/"$(echo ~{index_id} | tr '_' '/')" \
```

adamnovak (approved):
This looks like everything should work.

SPEC.md
jdidion: It means the directory in the execution environment where the file is to be written must have writable permissions.

jdidion: For example, if I define `File outfile = "outfile"` and then in the command do `printf "hello" > ~{outfile}`, the directory where the execution engine resolves `outfile` (presumably, the working directory, but it doesn't have to be) must be writable.

jdidion: There is not really any reason to use a `File` for this purpose, and (obviously) it adds a lot of confusion. Hence the impetus to disallow relative paths.

SPEC.md
jdidion: Fixed

SPEC.md
jdidion: JSON doesn't make a distinction between int and float. The test framework should compare these values as equal.

SPEC.md
jdidion: I'm guessing the tgz file contains the reference nested two levels deep. Rather than hard-code the path, I used the `--strip-components 2` argument to `tar`.

SPEC.md
stxue1: > Are we expecting the runner to produce one serialization here and not the other? Or @stxue1 are you doing something that needs to key on actually having a decimal point?

I currently do try to key on the decimal point to test for floats, but since the JSON schema doesn't actually support a distinction for floats and ints, there isn't much reason to actually key on decimal points, so I'll remove the check from our test suite.

SPEC.md
```diff
@@ -8548,6 +8576,9 @@ workflow test_transpose {
   
   output {
     Boolean is_true = transpose(input_array) == expected_output_array
+    Array[Array[Int]] out = transpose(input_array)
+    Array[Array[Int]] expected = expected_output_array
```
stxue1, line 8574: The example output section is missing the 2 new outputs.
The example output json below is currently defined as:
```json
{
  "test_transpose.is_true": true
}
```
and should become
```json
{
  "test_transpose.expected": [[0, 3], [1, 4], [2, 5]],
  "test_transpose.out": [[0, 3], [1, 4], [2, 5]],
  "test_transpose.is_true": true
}
```

SPEC.md
```diff
@@ -1295,14 +1302,17 @@ Example: string_to_file.wdl
...
+
+  String path1 = "~{infile}"
 
   # valid - String coerces unambiguously to File
-  File path3 = path1
+  File path2 = path1
 
   output {
-    Boolean paths_equal = path2 == path3
+    Boolean paths_equal = infile == path3
```
stxue1: `path3` is undefined, I think it should be replaced with `path2`?

SPEC.md
```diff
@@ -3916,7 +3938,9 @@ Example output:
 ```json
 {
   "optional_output.example2": null,
-  "optional_output.file_array_len": 1
+  "optional_output.file_array_len": 1,
+  "optional_output.example1": "example1.txt",
```
stxue1, line 3942: The `exclude_output` section below specifies both of these outputs to not be validated. Is this intentional or should one or the other be removed?

stxue1: The unit tests with ids `outputs`, `glob`, `relative_and_absolute`, `change_extension`, and `gen_files` all omit the outputs when specified in `exclude_output`. I think one or the other should be removed here.

stxue1:
The unit test `test_hints` `wc` command is malformed:
https://github.com/openwdl/wdl/blob/20ff54d261e44e903d94149eb1f5c4dd141d26ea/SPEC.md?plain=1#L4584

And should be:
```wdl
  wc -l < ~{foo}
```

The expected output should change from `3` to `2` as well:
https://github.com/openwdl/wdl/blob/20ff54d261e44e903d94149eb1f5c4dd141d26ea/SPEC.md?plain=1#L4619
```json
{
  "test_hints.num_lines": 2
}
```

stxue1:
`input_hint` is missing an output:
https://github.com/openwdl/wdl/blob/20ff54d261e44e903d94149eb1f5c4dd141d26ea/SPEC.md?plain=1#L4703-L4705

I'm not too certain what the expected output should be. MiniWDL currently returns
```json
{
    "input_hint.experience": []
}
```

SPEC.md
claymcleod: It appears this has been addressed.

SPEC.md
claymcleod: On this one, I don't see an `exclude_output` section for this test?

stxue1: I think my PR removed exclude_outputs for this test, so this is now resolved

claymcleod: I think quite a few of the comments here were fixed by later commits. Based on the fact that I don't see any exclude outputs now _and_ that every other comment you made seems to have been addressed, I'm going to assume this comment is out of date now and resolve it. If you feel it should be revisited, please open another issue.

claymcleod:
> `input_hint` is missing an output:
> 
> https://github.com/openwdl/wdl/blob/20ff54d261e44e903d94149eb1f5c4dd141d26ea/SPEC.md?plain=1#L4703-L4705
> 
> I'm not too certain what the expected output should be. MiniWDL currently returns
> 
> ```json
> {
>     "input_hint.experience": []
> }
> ```

This has also been resolved.

Based on this, I think the PR is good to go, and I'm going to merge it. This PR has sat for a second, so I did the best I could with making sure nothing was missed, but feel free to file another issue if anything wasn't pulled in correctly.

claymcleod (approved):

PR #670, opened 2024-06-26
WDL 1.2.1

### LINKED ISSUE #672: Semantics for `runtime` `disks` mount points are confusing
The `disks` key in the `runtime` section is documented as providing "persistent volumes" at certain mount point paths. Those paths are specified as being "in the host environment" in WDL 1.2, and I think "on the host machine" in 1.1.
https://github.com/openwdl/wdl/blob/664adc323f536ab1fac0c41bf3fc85f4f15eb52b/SPEC.md?plain=1#L5072-L5088

If the mount point is a *host* side path, then where is the storage expected to be mounted in the *container* where the WDL `command` section gets run? If the storage is meant to be mounted at the given path in the *container*, then why does that path need to also exist on the host?

Is this meant to just let the WDL task get access to a particular directory on the host, by mounting that path on the host into the container at the same path?

What kind of "persistence" specifically is supposed to be available? If two tasks run one after the other, and they both have a `disks` entry with a given mount point and size, should the second task be guaranteed to see files there written by the first task? Or can the "persistent" volume be a fresh empty directory for each task? Or is some kind of opportunistic sharing expected?

If a task requests a 100 GiB persistent volume, does it have to deal with the possibility that, upon being mounted, it already had 50 GiB of files in it left over from previous tasks and only has 50 GiB free space?

If two tasks run at the same time, can they ever share a persistent volume? Or does a task get exclusive ownership of a persistent volume while it is mounted?

We're trying to implement this in Toil in https://github.com/DataBiosphere/toil/pull/5001 and so far we've come up with an implementation that just mounts the specified path from the host into the container. But I think it really makes more sense to mount fresh empty directories with the given amount of reserved space into the container instead, since that matches what I would imagine a workflow would actually want. But that completely ignores the "persistent" part of the spec.

Are there any workflow examples that use mount points, beyond the test examples in the spec that just measure their size? What kind of behavior do they expect w.r.t. persistence or the relationship between in-container and host-side paths?

jdidion:
### Checklist

Closes #672


jdidion:
@stxue1 @adamnovak would also appreciate your reviews

SPEC.md
```diff
@@ -848,23 +849,39 @@ Single- and double-quotes do not need to be escaped within a multi-line string.
...
+    # always false.
+    if [ -f "~{f2}" ]; then
+      echo "~{f2}"
+    fi
+  >>>
 }
 ```
 
-An execution engine may support [other ways](#input-and-output-formats) to specify `File` and `Directory` inputs (e.g., as URIs), but prior to task execution it must [localize inputs](#task-input-localization) so that the runtime value of a `File`/`Directory` variable is a local path.
+Within a WDL file, literal values for files and directories may only be paths that are local to the execution environment.
```
adamnovak: So an engine is not allowed to support something like:
```
inputs {
    File genome_reference = "https://ftp.ncbi.nlm.nih.gov/genomes/all/GCA/000/001/405/GCA_000001405.15_GRCh38/seqs_for_alignment_pipelines.ucsc_ids/GCA_000001405.15_GRCh38_no_alt_analysis_set.fna.gz"
}
```
?

SPEC.md
```diff
@@ -5081,9 +5117,11 @@ Test config:
...
+The `disks` attribute provides a way to request one or more persistent volumes, each of which has a minimum size and is mounted at a specific location with both read and write permissions. When the `disks` attribute is provided, the execution engine must guarantee the requested resources are available or immediately fail the task prior to instantiating the command.
+
+If the mount point is omitted, it is assumed to be a persistent volume mounted at the root of the execution directory within a task.
 
-If a mount point is specified, then it must be an absolute path to a location in the host environment. If the mount point is omitted, it is assumed to be a persistent volume mounted at the root of the execution directory within a task.
+If a mount point is specified, then it must be an absolute path to a location in the host environment (i.e., within the container). The specified path either must not already exist in the host environment, or it must be empty and have at least the requested amount of space available. The mount point should be assumed to be ephemeral, i.e., it will be deleted after the task completes.
```
adamnovak: I've always seen "host environment" in the context of containerized applications used to mean the system *outside* the container; not the system *within* the container. If I am on a machine booted into Fedora and I run a WDL workflow that calls for an `ubuntu:22.04` container to run, the "host" as I understand it would be the Fedora environment.

Also, I don't think that containers have a notion of some amount of space being available at a path in the container. I don't think I can make an image that has 20 GiB of free space at `/mnt/freespace` and upload it to Docker Hub, as far as I know. So I don't think saying `must be empty and have at least the requested amount of space available` is different from just saying `must be empty`. You can mount at paths that don't exist in your container image, or you can mount at paths that exist as empty directories, but you can't mount at paths that are files or paths that are directories that have contents.

SPEC.md
```diff
@@ -10017,7 +10055,9 @@ workflow test_transpose {
   Array[Array[Int]] expected_output_array = [[0, 3], [1, 4], [2, 5]]
   
   output {
-    Boolean is_true = transpose(input_array) == expected_output_array
+    Array[Array[Int]] out = transpose(input_array) 
+    Array[Array[Int]] expected = expected_output_array
```
adamnovak, line 9973: There values will get output but aren't listed in the example output.

SPEC.md
jdidion: Good point - I've updated the spec to say that the execution engine is only required to support paths - the implication being it's also free to support URLs but that may not be portable across engines.

I'm open to expanding this in WDL 1.3 to require engines to support http(s) URLs as well. This would be consistent with the requirement to support http(s) URLs in imports.

SPEC.md
jdidion: Yes, we should be more consistent with usage. I've updated the spec such that all uses of "execution environment" refers to the context in which the task is executed, and "host" refers to the system that hosts the execution environment.

SPEC.md
jdidion: Fixed

SPEC.md
```diff
@@ -3386,7 +3419,7 @@ A document is imported using it's [URI](https://en.wikipedia.org/wiki/Uniform_Re
 * `https://`
 * 🗑 `file://` - Using the `file://` protocol for local imports can be problematic. Its use is deprecated and will be removed in WDL 2.0.
 
-In the event that there is no protocol specified, the import is resolved **relative to the location of the current document**. In the primary WDL document, a protocol-less import is relative to the host file system. If a protocol-less import starts with `/` it is interpreted as relative to the root of the host in the resolved URI.
+In the event that there is no protocol specified, the import is resolved **relative to the location of the current document**. In the primary WDL document, a protocol-less import is relative to the folder that contains the primary WDL file. If a protocol-less import starts with `/` it is interpreted as relative to the root of the file system that contains the primary WDL file.
```
patmagee, line 3492: The language here makes an assumption that path resolution happens on disk. This is not always true, as would be the case for resolving the imports of a Dockstore workflow. Its common practice to use relative import uris, but when resolving the workflow you would resolve it against the `https://` context

SPEC.md
```diff
@@ -5081,9 +5119,11 @@ Test config:
     * `Array[String]` - An array of disk specifications.
 * Default value: `1 GiB`
 
-The `disks` attribute provides a way to request one or more persistent volumes, each of which has a minimum size and is mounted at a specific location. When the `disks` attribute is provided, the execution engine must guarantee the requested resources are available or immediately fail the task prior to instantiating the command.
+The `disks` attribute provides a way to request one or more persistent volumes, each of which has a minimum size and is mounted at a specific location with both read and write permissions. When the `disks` attribute is provided, the execution engine must guarantee the requested resources are available or immediately fail the task prior to instantiating the command.
```
patmagee, line 5196: > read and write permissions

I am worried this requirement constitutes a breaking change. I am not 1--% clear how cromwell works with data-disks and other existing disks but I would not be surprised if it treated some mounts as read-only

SPEC.md
```diff
@@ -5081,9 +5119,11 @@ Test config:
...
+
+If the mount point is omitted, it is assumed to be a persistent volume mounted at the root of the execution directory within a task.
 
-If a mount point is specified, then it must be an absolute path to a location in the host environment. If the mount point is omitted, it is assumed to be a persistent volume mounted at the root of the execution directory within a task.
+If a mount point is specified, then it must be an absolute path to a location in the execution environment (i.e., within the container). The specified path either must not already exist in the execution environment, or it must be empty and have at least the requested amount of space available. The mount point should be assumed to be ephemeral, i.e., it will be deleted after the task completes.
```
patmagee, line 5200: > The specified path either must not already exist in the execution environment, or it must be empt

I think this is technically hard to achieve when using a container technology like Docker, which simply allows overriding paths.

SPEC.md
```diff
@@ -7957,7 +7996,7 @@ File join_paths(File, Array[String]+)
 File join_paths(Array[String]+)
 ```
 
-Joins together two or more paths into an absolute path in the host filesystem.
+Joins together two or more paths into an absolute path in the execution environment's filesystem.
```
patmagee, line 7900: We should define these somewhere. Execution environment can mean multiple different things in different contexts. IE in this case it could be
- the trusted compute environment ie a cloud project
- a VM
- The running docker container

SPEC.md
```diff
@@ -11257,6 +11301,12 @@ Here is an example JSON input file for a workflow `wf`:
 
 WDL implementations are only required to support workflow execution, and not necessarily task execution, so a JSON input format for tasks is not specified. However, it is strongly suggested that if an implementation does support task execution, that it also supports this JSON input format for tasks. It is left to the discretion of the WDL implementation whether it is required to prefix the task input with the task name, i.e., `mytask.infile` vs. `infile`.
 
+### File/Directory Inputs
+
+It is up to the execution engine to resolve input files and directories and stage them into the execution environment. The execution engine is free to specify the values that are allowed for `File` and `Directory` parameters, but at a minimum it is required to support POSIX absolute file paths (e.g., `/path/to/file`).
+
+It is strongly recommended that input files and directories be specified as absolute paths to local files or as URLs. If relative paths are allowed, then it is suggested that they be resolved relative to the directory that contains the input JSON file (if a file is provided) or to the working directory in which the workflow is initially launched.
```
patmagee, line 11233: Should change the suggestion from `URL to `URI` to capture use cases of cloud URIs

claymcleod:
I'm closing this in favor of just merging the WDL 1.2.1 branch manually.

PR #689, opened 2024-11-01
Apply some fixes to WDL examples

stxue1:
These are fixes for some of the comments I left at #669 

I'm merging into `wdl-1.1.3` instead so the eventual merge into `wdl-1.1` by #669 will be cleaner.

There are only two tests left that don't work: `hisat2` and `gatk_haplotype_caller`. I haven't look at them yet as I don't have much experience with running the hisat2 or gatk pipeline.

Other than those two, I believe the rest of the examples work.

### Checklist

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -4693,7 +4693,9 @@ Example input:
 Example output:
 
 ```json
-{}
+{
+  "input_hint.experience": []
+}
 ```
 </p>
 </details>
--- a/SPEC.md
+++ b/SPEC.md
@@ -1312,7 +1312,7 @@ workflow string_to_file {
   File path2 = path1
 
   output {
-    Boolean paths_equal = infile == path3
+    Boolean paths_equal = infile == path2
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


SPEC.md
```diff
@@ -4616,7 +4608,7 @@ Example output:
 
 ```json
 {
-  "test_hints.num_lines": 3
```
claymcleod, line 4619: It looks like [`greetings.txt`](https://github.com/openwdl/wdl-tests/blob/main/spec/wdl-1.1/data/greetings.txt) has 3 lines.

claymcleod (changes requested):
Otherwise looks good to me.

SPEC.md
stxue1: Running `wc -l` on the file returns 2 since wc counts the number of newline characters, so I think 2 is correct here.

claymcleod: It appears you're correct, today I learned something new.

PR #695, opened 2025-01-15
feat: adds enumerations

claymcleod:
This commit adds enumerations to WDL. 

Based on my discussions and browsing previous threads surround the idea, it sounds like the value of adding enumerations to WDL is broadly agreed upon as an improvement regarding the user experience for limiting assignment to a set of valid values within a particular context; it's simply the details that need to be hashed out. 

In this PR, enumerations in WDL are valued, meaning that they have an assigned valued for each variant therein. These values can either be explicitly or implicitly typed. When it comes to the details, I try to take a common sense approach and a "middle of the road" position on the spectrum of conciseness versus flexibility. In particular:

- The inner type can be elided from the enum's signature if it is trivially computed (leans towards conciseness).
- When you'd like to be explicit with the inner type to resolve ambiguity, you can be.
- Enumeration variants with no assigned value are assumed to be typed as a `String` with an assigned value equivalent to the variant name (leans towards common sense and conciseness).
- Enumerations that are not of the same type cannot be compared, as that is equivalent to a type mismatch (leans towards common sense approach, as comparing two enums of different types implies different contexts, even if the inner values of the variants are the same).

I'll leave the details to what I've written in the changeset.

### Checklist

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -1491,15 +1491,16 @@ Note that the ability to assign values to `Struct` declarations other than struc
 
 ##### Enumeration Types (Enums)
 
-An enumeration (or "enum") is a closed set of alternatives that are considered semantically valid in a specific context. An enum is [defined]() at the top-level of the WDL document and can be used as a declaration type anywhere in the document.
+An enumeration (or "enum") is a closed set of enumerated values (known as "variants") that are considered semantically valid in a specific context. An enum is defined at the top-level of the WDL document and can be used as a declaration type anywhere in the document.
 
-An enum is defined using the `enum` keyword, followed by a globally unique name, followed by a comma-delimited list of identifiers in braces. The value for an enum cannot be created; rather, it must be selected from the list of available values using the `<name>.<value>` syntax.
+An enum is defined using the `enum` keyword, followed by a globally unique name, followed by a comma-delimited list of identifiers—optionally tagged with values—in braces. When referring to a variant within an enum, for example, when assigning to an enum declaration, the `<name>.<variant>` syntax should be used.
 
 ```wdl
 enum FileKind {
   FASTQ,
   BAM
 }
+
 task process_file {
   input {
     File infile
@@ -1516,6 +1517,7 @@ workflow process_files {
     Array[File] files
     FileKind kind
   }
+
   scatter (file in files) {
     call process_file {
       input:
@@ -1526,7 +1528,49 @@ workflow process_files {
 }
 ```
 
-As an example, consider a workflow that processes different types of NGS files and has a `file_kind` input parameter that is expected to be either "fastq" or "bam". Using `String` as the type of `file_kind` is not ideal - if the user specifies an invalid value, the error will not be caught until runtime, perhaps after the workflow has already run for several hours. Alternatively, using an `enum` type for `file_kind` restricts the allowed values such that the execution engine can validate the input prior to executing the workflow.
+As an example, consider a workflow that processes different types of NGS files and has a `file_kind` input parameter that is expected to be either "FASTQ" or "BAM". Using `String` as the type of `file_kind` is not ideal - if the user specifies an invalid value, the error will not be caught until runtime, perhaps after the workflow has already run for several hours. Alternatively, using an `enum` type for `file_kind` restricts the allowed values such that the execution engine can validate the input prior to executing the workflow.
+
+Enums are valued, meaning that each variant within an enum has an associated value. To assign a type to the values therein, enums can either be _explicitly_ or _implicitly_ typed.
+
+* Explicitly typed enums take an explicit type assigment within square brackets after the enum's identifier that declares the type of the value. Where possible, explicitly typed enums should resolve ambiguities via coercion.
+* Implicitly typed enums are enums where the values can be unambiguously mapped to a single [primitive type](#primitive-types). Enums that are implicitly typed and for which no values are assigned are assumed to be `String` valued with values matching the variant names. All values within an enum must have the same type or an error is thrown.
+
+```wdl
+# An explicitly typed enum that is `String`-valued.
+enum FruitColors[String] {
+  Banana = "yellow",
+  Orange = "orange",
+  Apple = "red",
+}
+
+# An explicitly typed enum that is `Float`-valued. Because the enum is
+# explicitly typed, the `ThreePointOh` variant can be coerced to a `Float`,
+# which is a valid enumeration definition.
+enum FavoriteFloat[Float] {
+  ThreePointOh = 3,
+  FourPointOh = 4.0
+}
+
+# ERROR: because the enum is implicitly typed, the type cannot be unambiguously
+# resolved, which results in an error.
+enum FavoriteNumber {
+  ThreePointOh = 3,
+  FourPointOh = 4.0
+}
+
+# An implicitly typed enum that is `String`-valued.
+enum Whitespace {
+  Tab = "\t",
+  Space = " "
+}
+
+# An implicitly typed enum that is implied to be `String`-valued with the
+# values "FASTQ" and "BAM" respectively.
+enum FileKind {
+  FASTQ,
+  BAM
+}
+```
 
 #### Hidden and Scoped Types
 
@@ -3397,7 +3441,7 @@ struct Invalid {
 
 ## Enum Definition
 
-An `Enum` is an enumerated type. Enums enable the creation of types that represent closed sets of alternatives (called "variants") that are semantically valid in a specific context. Once defined, an `Enum` type can be used as the type of a declaration like any other type. However, new variants of an `Enum` cannot be created. Instead, a declaration having an `Enum` type must be assigned one of the variants created as part of the `Enum`'s definition.
+An `enum` is an enumerated type. Enums enable the creation of types that represent closed sets of alternatives (called "variants") that are semantically valid in a specific context. Once defined, an `enum` type can be used as the type of a declaration like any other type. However, new variants of an `enum` cannot be created. Instead, a declaration having an `enum` type must be assigned one of the variants created as part of the `enum`'s definition.
 
 An enum definition is a top-level WDL element, meaning it is defined at the same level as tasks, workflows, and structs, and it cannot be defined within a task or workflow body. An enum is defined using the `enum` keyword, followed by a name that is unique within the WDL document, and a body containing a comma-delimited list of varaints in braces (`{}`).
 
@@ -3411,8 +3455,8 @@ enum Color {
 
 An enum can be thought of as two different constructs sharing the same name:
 
-* A `Struct` with one `String`-type member, `name`, whose value is the "stringified" version of the variant's identifier. For example, the name of `Color.RED` is `"RED"`. Unlike structs, it is not possible to create new instances of an `Enum` outside of the enum's definition.
-* A global variable of type `Object` whose members have names equal to the enum variant names, and whose values are instances of the `Enum`.
+* A `Struct` with one `String`-type member, `name`, whose value is the "stringified" version of the variant's identifier. For example, the name of `Color.RED` is `"RED"`. Unlike structs, it is not possible to create new instances of an `enum` outside of the enum's definition.
+* A global variable of type `Object` whose members have names equal to the enum variant names, and whose values are instances of the `enum`.
 
 For example, the above definition of `Color` can be thought of as shorthand for:
 
@@ -3442,13 +3486,13 @@ Keep in mind that the above example is illustrative - it is not actually possibl
 
 ### Enum Usage
 
-An `Enum`'s variants are [accessed](#member-access) using a `.` to separate the variant name from the `Enum`'s identifier. Each variant has single member, `String name`, that can be accessed through a `.` separator.
+An `enum`'s variants are [accessed](#member-access) using a `.` to separate the variant name from the `enum`'s identifier. Each variant has single member, `String name`, that can be accessed through a `.` separator.
 
-A declaration with an `Enum` type can only be initialized by referencing a variant directly or by assigning it to the value of another declaration of the same `Enum` type.
+A declaration with an `enum` type can only be initialized by referencing a variant directly or by assigning it to the value of another declaration of the same `enum` type.
 
-Two enum values can be tested for equality (i.e., using `==` or `!=`). To be equal, two enum values must be the same variant of the same `Enum` type. Enum variants are not ordered, so they cannot be compared (i.e., using `>`, `>=`, `<`, `<=`).
+Two enum values can be tested for equality (i.e., using `==` or `!=`). To be equal, two enum values must be the same variant of the same `enum` type. Enum variants are not ordered, so they cannot be compared (i.e., using `>`, `>=`, `<`, `<=`).
 
-An `Enum` cannot be coerced to or from any other type. However, an enum value can be [serialized to/deserialized from a `String`]() or to/from [JSON]().
+An `enum` cannot be coerced to or from any other type. However, an enum value can be [serialized to/deserialized from a `String`]() or to/from [JSON]().
 
 ```wdl
 version 1.2
@@ -3675,7 +3719,7 @@ struct Patient {
 
 ### Importing and Aliasing Enums
 
-Enums are [imported in the same way as `Struct`s](#struct-namespacing) and have the same namespacing rules, namely that Enums exist in the document's global scope, and importing an `Enum` copies its definition into the global scope of the importing document (potentially using an alias).
+Enums are [imported in the same way as `Struct`s](#struct-namespacing) and have the same namespacing rules, namely that Enums exist in the document's global scope, and importing an `enum` copies its definition into the global scope of the importing document (potentially using an alias).
 
 ```wdl
 version 1.2
--- a/SPEC.md
+++ b/SPEC.md
@@ -39,6 +39,7 @@ Revisions to this specification are made periodically in order to correct errors
         - [Map\[P, Y\]](#mapp-y)
         - [🗑 Object](#-object)
         - [Custom Types (Structs)](#custom-types-structs)
+        - [Enumeration Types (Enums)](#enumeration-types-enums)
       - [Hidden and Scoped Types](#hidden-and-scoped-types)
         - [`Union` (Hidden Type)](#union-hidden-type)
         - [`hints`, `input`, and `output` (Scoped Types)](#hints-input-and-output-scoped-types)
@@ -73,9 +74,12 @@ Revisions to this specification are made periodically in order to correct errors
   - [WDL Documents](#wdl-documents)
   - [Versioning](#versioning)
   - [Struct Definition](#struct-definition)
+  - [Enum Definition](#enum-definition)
+    - [Enum Usage](#enum-usage)
   - [Import Statements](#import-statements)
     - [Import URIs](#import-uris)
     - [Importing and Aliasing Structs](#importing-and-aliasing-structs)
+    - [Importing and Aliasing Enums](#importing-and-aliasing-enums)
   - [Task Definition](#task-definition)
     - [Task Inputs](#task-inputs)
       - [Task Input Localization](#task-input-localization)
@@ -1357,7 +1361,7 @@ Due to the lack of explicitness in the typing of `Object` being at odds with the
 
 ##### Custom Types (Structs)
 
-WDL provides the ability to define custom compound types called [structs](#struct-definition). `Struct` types are defined directly in the WDL document and are usable like any other type. A struct is defined using the `struct` keyword, followed by a unique name, followed by member declarations within braces. A struct definition contains any number of declarations of any types, including other `Struct`s.
+WDL provides the ability to define custom compound types called [structs](#struct-definition). `Struct` types are defined at the top-level of the WDL document and are usable like any other type. A struct is defined using the `struct` keyword, followed by a unique name, followed by member declarations within braces. A struct definition contains any number of declarations of any types, including other `Struct`s.
 
 A declaration with a custom type can be initialized with a struct literal, which begins with the `Struct` type name followed by a comma-separated list of name-value pairs in braces (`{}`), where name-value pairs are delimited by `:`. The member names in a struct literal are not quoted. A struct literal must provide values for all of the struct's non-optional members, and may provide values for any of the optional members. The members of a struct literal are validated against the struct's definition at the time of creation. Members do not need to be in any specific order. Once a struct literal is created, it is immutable like any other WDL value.
 
@@ -1484,6 +1488,45 @@ Example output:
 * Any `Object`/`Map` member that does not correspond to a member of the struct is ignored.
 
 Note that the ability to assign values to `Struct` declarations other than struct literals is deprecated and will be removed in WDL 2.0.
+
+##### Enumeration Types (Enums)
+
+An enumeration (or "enum") is a closed set of alternatives that are considered semantically valid in a specific context. An enum is [defined]() at the top-level of the WDL document and can be used as a declaration type anywhere in the document.
+
+An enum is defined using the `enum` keyword, followed by a globally unique name, followed by a comma-delimited list of identifiers in braces. The value for an enum cannot be created; rather, it must be selected from the list of available values using the `<name>.<value>` syntax.
+
+```wdl
+enum FileKind {
+  FASTQ,
+  BAM
+}
+task process_file {
+  input {
+    File infile
+    FileKind kind = FileKind.FASTQ
+  }
+  
+  command <<<
+  echo "Processing ~{kind} file"
+  ...
+  >>>
+}
+workflow process_files {
+  input {
+    Array[File] files
+    FileKind kind
+  }
+  scatter (file in files) {
+    call process_file {
+      input:
+        infile = file,
+        kind = kind
+    }
+  }
+}
+```
+
+As an example, consider a workflow that processes different types of NGS files and has a `file_kind` input parameter that is expected to be either "fastq" or "bam". Using `String` as the type of `file_kind` is not ideal - if the user specifies an invalid value, the error will not be caught until runtime, perhaps after the workflow has already run for several hours. Alternatively, using an `enum` type for `file_kind` restricts the allowed values such that the execution engine can validate the input prior to executing the workflow.
 
 #### Hidden and Scoped Types
 
@@ -3352,6 +3395,94 @@ struct Invalid {
 }
 ```
 
+## Enum Definition
+
+An `Enum` is an enumerated type. Enums enable the creation of types that represent closed sets of alternatives (called "variants") that are semantically valid in a specific context. Once defined, an `Enum` type can be used as the type of a declaration like any other type. However, new variants of an `Enum` cannot be created. Instead, a declaration having an `Enum` type must be assigned one of the variants created as part of the `Enum`'s definition.
+
+An enum definition is a top-level WDL element, meaning it is defined at the same level as tasks, workflows, and structs, and it cannot be defined within a task or workflow body. An enum is defined using the `enum` keyword, followed by a name that is unique within the WDL document, and a body containing a comma-delimited list of varaints in braces (`{}`).
+
+```wdl
+enum Color {
+  RED,
+  BLUE,
+  GREEN
+}
+```
+
+An enum can be thought of as two different constructs sharing the same name:
+
+* A `Struct` with one `String`-type member, `name`, whose value is the "stringified" version of the variant's identifier. For example, the name of `Color.RED` is `"RED"`. Unlike structs, it is not possible to create new instances of an `Enum` outside of the enum's definition.
+* A global variable of type `Object` whose members have names equal to the enum variant names, and whose values are instances of the `Enum`.
+
+For example, the above definition of `Color` can be thought of as shorthand for:
+
+```wdl
+version 1.2
+struct Color {
+  String name
+}
+Object Color = object {
+  RED: Color { name: "RED" },
+  BLUE: Color { name: "BLUE" },
+  GREEN: Color { name: "GREEN" }
+}
+workflow test {
+  input {
+    # The first usage of `Color` on the left is a reference to the struct type.
+    # The second usage of `Color` on the right is a reference to the object declaration.
+    Color color = Color.RED
+  }
+  # Error! Creating additional instances of `Color` is not allowed.
+  Color pink = Color { name: "PINK" }
+}
+```
+
+Keep in mind that the above example is illustrative - it is not actually possible to have a top-level declaration, nor is it possible to create a struct that is not able to be instantiated. The example just shows the closest semantic equivalent of Enums using existing WDL syntax.
+
+
+### Enum Usage
+
+An `Enum`'s variants are [accessed](#member-access) using a `.` to separate the variant name from the `Enum`'s identifier. Each variant has single member, `String name`, that can be accessed through a `.` separator.
+
+A declaration with an `Enum` type can only be initialized by referencing a variant directly or by assigning it to the value of another declaration of the same `Enum` type.
+
+Two enum values can be tested for equality (i.e., using `==` or `!=`). To be equal, two enum values must be the same variant of the same `Enum` type. Enum variants are not ordered, so they cannot be compared (i.e., using `>`, `>=`, `<`, `<=`).
+
+An `Enum` cannot be coerced to or from any other type. However, an enum value can be [serialized to/deserialized from a `String`]() or to/from [JSON]().
+
+```wdl
+version 1.2
+enum Pet {
+  CAT,
+  DOG,
+  RAT
+}
+enum HotFood {
+  DOG,
+  POTATO,
+  TAMALE
+}
+task pet_pet {
+  input {
+    Pet? pet
+  }
+  Pet my_pet = select_first([pet, Pet.DOG])
+  Array[String] all_pet_names = [
+    Pet.CAT.name,
+    Pet.DOG.name,
+    Pet.RAT.name
+  ]
+  command <<<
+  echo "There are ~{length(all_pet_names)} kinds of pet: ~{sep(", ", all_pet_names)}"
+  echo "I have a pet ~{my_pet.name}"
+  echo "I am petting my ~{my_pet}"
+  >>>
+  output {
+    Boolean is_false = Pet.DOG == HotFood.DOG
+  }
+}
+```
+
 ## Import Statements
 
 Although a WDL workflow and the task(s) it calls may be defined completely within a single WDL document, splitting it into multiple documents can be beneficial in terms of modularity and code resuse. Furthermore, complex workflows that consist of multiple subworkflows must be defined in multiple documents because each document is only allowed to contain at most one workflow.
@@ -3539,6 +3670,21 @@ struct Patient {
   Int age
   PatientIncome? income
   Map[String, Array[File]] assay_data
+}
+```
+
+### Importing and Aliasing Enums
+
+Enums are [imported in the same way as `Struct`s](#struct-namespacing) and have the same namespacing rules, namely that Enums exist in the document's global scope, and importing an `Enum` copies its definition into the global scope of the importing document (potentially using an alias).
+
+```wdl
+version 1.2
+import "color.wdl" alias Color as Hue
+workflow another_wf {
+  input {
+    Hue hue = Hue.BLUE
+  }
+  ...
 }
 ```
 
--- a/SPEC.md
+++ b/SPEC.md
@@ -76,6 +76,9 @@ Revisions to this specification are made periodically in order to correct errors
   - [Struct Definition](#struct-definition)
   - [Enum Definition](#enum-definition)
     - [Enum Usage](#enum-usage)
+    - [Enum Serialization and Deserialization](#enum-serialization-and-deserialization)
+      - [JSON Input and Output](#json-input-and-output)
+      - [Command Section Serialization](#command-section-serialization)
   - [Import Statements](#import-statements)
     - [Import URIs](#import-uris)
     - [Importing and Aliasing Structs](#importing-and-aliasing-structs)
@@ -199,6 +202,10 @@ Revisions to this specification are made periodically in order to correct errors
     - [`contains_key`](#-contains_key)
     - [`values`](#-values)
     - [`collect_by_key`](#collect_by_key)
+  - ✨ [Enum Functions](#enum-functions)
+    - ✨ [`name`](#name)
+    - ✨ [`names`](#names)
+    - ✨ [`variants`](#variants)
   - [Other Functions](#other-functions)
     - [`defined`](#defined)
     - [`length`](#length)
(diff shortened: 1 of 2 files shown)


claymcleod:
I think this is ready for a preliminary review—I haven't added any tests yet, but that's because I want to make sure the idea, as I have codified it here, has support from a good number of people before investing the time to do that.

markjschreiber (commented):
I like the proposal. I think you should add that Enum values cannot be mutated or reassigned (they are immutable) and that Enums are closed once declared (you can't extend them to add new values).

markjschreiber (commented):
Open question, can the values of an enum be scattered? In most programming languages Enums are iterable.

SPEC.md
```diff
@@ -3352,6 +3397,115 @@ struct Invalid {
...
+  ThreePointOh = 3,
+  FourPointOh = 4.0
+}
+
+# ERROR: because the enum is implicitly typed, the type cannot be unambiguously
+# resolved, which results in an error.
+enum FavoriteNumber {
+  ThreePointOh = 3,
+  FourPointOh = 4.0
+}
```
peterhuene: Given other places where we deduce a common type in WDL (e.g. `[1, 2.0]` -> `Array[Float]`), I would expect this to unambiguously resolve to `Float` as all of the assigned types coerce to it, rather than be considered an error.

I would expect it to error if the first assignment was to `"3"`, for example.

peterhuene: Additionally, given that the type parameter to the enum declaration is required to be primitive and that there aren't that many primitive type coercions, I wonder if we can do without supporting an explicit syntax, at least initially.

The only use I can foresee with explicitly specifying the type would be if one wanted to have an enum of `File` or `Directory`, which makes little sense to me (an enum of `String` would have its values implicitly coerced to `File` and `Directory` where needed anyway).

peterhuene: Although, reading it again, there doesn't seem to be an explicit requirement that a type in the _explicit syntax_ be primitive. Can we do something like:

```wdl
enum Foo[Array[String]] {
   Foo = ["foo", "bar", "baz"],
   Bar = ["qux", "quux", "quuux"],
}
```
?

If this is desired, then we probably should support the explicit syntax. If not, I personally think we can eliminate it.

Additionally, we should call out that the RHS expression in the assignment must only contain literal expressions and that string interpolation isn't supported in this context as these are global declarations.

claymcleod: > Given other places where we deduce a common type in WDL (e.g. `[1, 2.0]` -> `Array[Float]`), I would expect this to unambiguously resolve to `Float` as all of the assigned types coerce to it, rather than be considered an error.
> 
> 
> 
> I would expect it to error if the first assignment was to `"3"`, for example.

Ah yes, you are correct. I will update this example.

claymcleod: > Although, reading it again, there doesn't seem to be an explicit requirement that a type in the _explicit syntax_ be primitive. Can we do something like:
> 
> 
> 
> ```wdl
> 
> enum Foo[Array[String]] {
> 
>    Foo = ["foo", "bar", "baz"],
> 
>    Bar = ["qux", "quux", "quuux"],
> 
> }
> 
> ```
> 
> ?
> 
> 
> 
> If this is desired, then we probably should support the explicit syntax. If not, I personally think we can eliminate it.
> 
> 
> 
> Additionally, we should call out that the RHS expression in the assignment must only contain literal expressions and that string interpolation isn't supported in this context as these are global declarations.

Exactly. I can see a map being useful: for example, perhaps you have an enum that can help you scatter over _all_ contigs versus just a defined list of canonical contigs.

That being said, perhaps it's not required to explicitly define non-primitive types as well. For instance, the above is unequivocally resolvable to the `Array[String]`—would it be fair to just say the type must be unambiguously resolved?

I'll add the clarification on string interpolation.

claymcleod:
> Open question, can the values of an enum be scattered? In most programming languages Enums are iterable. 

Great point. In keeping with the spirit of what was done for similar methods, like getting the [`keys`](https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#keys) of an object, I will add a `variants` standard library function.

markjschreiber:
Could the function just be called `enumerate`

SPEC.md
peterhuene: It should be unambiguous to deduce a type or otherwise error; the only catch is `File` and `Directory` since they don't have a literal type (but otherwise coerce from `String` just fine).

claymcleod:
> Could the function just be called `enumerate`

Personally, I think it makes me sense to call it `variants`. This is to be consistent with the choice of the function name matching the concept name for enumeration of those concepts within `Object`s. For example, `keys` defines _what_ is being enumerated, not the literal name of the enumeration action (`enumerate`, `enumerate_keys`, `enumerate_values`). Since each member of an enum is called a "variant", the name `variants` would align with the pre-existing convention. Furthermore, you might want to enumerate more than just the variants, so it has the added benefit of being co-introduced with an overload to the `values` function for an enum.

SPEC.md
claymcleod: In the latest commit, I've:

- Updated the example to be something that wouldn't resolve.
- Clarified what should happen if a `String` that looks like a `File` or `Directory` is passed in to an implicitly typed enum.

claymcleod:
I've incorporated all of the feedback above, and this should be ready for a re-review. That being said, I don't like what I have on the page right now because I realized something while writing the example (well, maybe two things).

- To my knowledge, the `variants` method would be a new paradigm being added to WDL, as it's a function that takes a custom type _itself_ rather than an instance of the type. For example, you can iterate on the `keys` in a `Map`, but that's called on an _instance_ of `Map`. There are, for example, no methods that iterate the fields in a `Struct`.
- During enum assignment, you're really assigning the value to one of the variants _within_ the enum, not the enum itself. So it's not strictly true to call this variable "an instance of the enum". It is probably more correct to say "an instance of one of the enum's variants".

Here's an example to make sure it's clear why this distinction is important.

```wdl
enum FileType {
  FASTQ,
  BAM
}

FileType foo = FileType.FASTQ

# You wouldn't call `variants` on `foo`—why would that be necessary? And does it even really make sense? 
# `foo` represents a variant _within_ the enum rather than the `enum` itself, so enumerating the variants
# within the enum itself seems strange.
variants(foo)

# You'd more likely want to, say, scatter across the variants from the enum type directly.
# But this introduces sort of a strange-looking function call.
variants(FileType)
```

I will continue to chew on this but curious if others have thoughts on what should be done here. My gut reaction is that perhaps the definition custom types (enums and structs) need to introduce a namespace within the same name as the type (or alias) within which standard library functions can be constructed automatically. This would be similar to the `::` operator in Rust, e.g.,

```wdl
scatter(type in FileType::variants()) {
  # ...
}
```

patmagee:
> I've incorporated all of the feedback above, and this should be ready for a re-review. That being said, I don't like what I have on the page right now because I realized something while writing the example (well, maybe two things).
> 
> * To my knowledge, the `variants` method would be a new paradigm being added to WDL, as it's a function that takes a custom type _itself_ rather than an instance of the type. For example, you can iterate on the `keys` in a `Map`, but that's called on an _instance_ of `Map`. There are, for example, no methods that iterate the fields in a `Struct`.
> * During enum assignment, you're really assigning the value to one of the variants _within_ the enum, not the enum itself. So it's not strictly true to call this variable "an instance of the enum". It is probably more correct to say "an instance of one of the enum's variants".
> 
> Here's an example to make sure it's clear why this distinction is important.
> 
> ```wdl
> enum FileType {
>   FASTQ,
>   BAM
> }
> 
> FileType foo = FileType.FASTQ
> 
> # You wouldn't call `variants` on `foo`—why would that be necessary? And does it even really make sense? 
> # `foo` represents a variant _within_ the enum rather than the `enum` itself, so enumerating the variants
> # within the enum itself seems strange.
> variants(foo)
> 
> # You'd more likely want to, say, scatter across the variants from the enum type directly.
> # But this introduces sort of a strange-looking function call.
> variants(FileType)
> ```
> 
> I will continue to chew on this but curious if others have thoughts on what should be done here. My gut reaction is that perhaps the definition custom types (enums and structs) need to introduce a namespace within the same name as the type (or alias) within which standard library functions can be constructed automatically. This would be similar to the `::` operator in Rust, e.g.,
> 
> ```wdl
> scatter(type in FileType::variants()) {
>   # ...
> }
> ```

@claymcleod What about a new paradigm but one that I think will be easier to roll out: Adding class level methods

``` wdl
enum FileType {
  FASTQ,
  BAM
}

FileType.variants()
```

jdidion:
What if we think of an `enum` as a special kind of `Map` whose keys are strings and values are of a hidden type.

So

```
enum Foo {
  BAR = "bar"
  BAZ = "baz"
}
```

would be equivalent to the following

```
# this is a hidden type
struct Foo {
  String value
}

# this is a singleton instance - hand-wave that the instance
# can have the same name as the type
Map[String, Foo] Foo = {
  "BAR": Foo { value = "bar" },
  "BAZ": Foo { value = "baz" }
}

```

If we wanted to, we could also extend existing methods to handle an argument of type `enum`, e.g.

* `keys` would return the names of the variants.
* Indexing (`[]`) would enable getting a variant by name

```
Foo bar1 = Foo.BAR
Foo bar2 = Foo["BAR"]
```

There's also the question of whether an `enum` variant is automatically coerced to it's value type, or if an explicit call to `.value` is required:

```
# this
String bar_string1 = bar1
# or this?
String bar_string2 = bar2.value
# or both?
```

claymcleod:
> @claymcleod What about a new paradigm but one that I think will be easier to roll out: Adding class level methods
> 
> ```wdl
> enum FileType {
>   FASTQ,
>   BAM
> }
> 
> FileType.variants()
> ```

Based on my understanding of what you're saying, I think both suggestions are essentially equivalent. The only real question is "which operator should we chose"? I can only think of one relatively good argument, and it's in favor of `::` (or, more strictly, anything other than `.`).

If we introduce these static methods with a `.` operator, the next logical question is "why can I not do something like `<variable>.keys()` instead of `keys(<variable>)`? After all, the use of `.` to access methods directly attached to an instance of a class is going to be something that users are used to (and perhaps even expect) from their experience using more general programming languages. In truth, I'm not sure if there is a good reason that something like that doesn't work other than simplicity. So, in this case, it might be quite discordant to find that the less obvious case of using `.` (accessing static methods) works but the more obvious case of using `.` (accessing methods on a class instance) doesn't work.

In the case of an operator like `::` (or any operator other than `.`), I don't believe these expectations (and, thus, the perceived discordance) are introduced—because `::` is a more uncommon operator, it will likely be more natural to think of this as much different than what a method accessor would do. In other words, it doesn't follow that the addition of `::` to access namespaced methods will introduce frustration about the lack of a method accessor.

Just to be clear, I don't think this _has_ to be `::` specifically. If we don't like the way it looks or we don't want to lean too hard into what Rust is doing (I'm obviously biased here). I just think it should probably not be `.` based on the argument above. To that end, I don't have a better suggestion for what it _should_ be, and I'd be open to hearing any.

claymcleod:
> What if we think of an `enum` as a special kind of `Map` whose keys are strings and values are of a hidden type.
> 
> So
> 
> ```
> enum Foo {
>   BAR = "bar"
>   BAZ = "baz"
> }
> ```
> 
> would be equivalent to the following
> 
> ```
> # this is a hidden type
> struct Foo {
>   String value
> }
> 
> # this is a singleton instance - hand-wave that the instance
> # can have the same name as the type
> Map[String, Foo] Foo = {
>   "BAR": Foo { value = "bar" },
>   "BAZ": Foo { value = "baz" }
> }
> ```
> 
> If we wanted to, we could also extend existing methods to handle an argument of type `enum`, e.g.
> 
>     * `keys` would return the names of the variants.
> 
>     * Indexing (`[]`) would enable getting a variant by name
> 
> 
> ```
> Foo bar1 = Foo.BAR
> Foo bar2 = Foo["BAR"]
> ```
> 
> There's also the question of whether an `enum` variant is automatically coerced to it's value type, or if an explicit call to `.value` is required:
> 
> ```
> # this
> String bar_string1 = bar1
> # or this?
> String bar_string2 = bar2.value
> # or both?
> ```

I think I'm not connecting what the proposed advantages of this design are over what I've got in the PR, so I'm not sure on my overall feeling on this. If you could share a bit more from your perspective, that'd be great. To address some smaller points:

**Accessors**

```
Foo bar1 = Foo.BAR
Foo bar2 = Foo["BAR"]
```

On this, I'm not sure I like enabled the second line here. The core value of introducing enums is that there is a way to define a closed set of values that can be checked at compile time. When you start accessing things as if they were a map, you run into the issue that the variant might not exist, causing a runtime exception. Beyond that, I just don't see a lot of value to being able to access variants this way: to me, enums seem much stronger if we keep them as a strongly typed and statically checked part of the language with limited scope (in other words, if you need a `Map`, then use a `Map`).

**Coercion**

> ```
> # this
> String bar_string1 = bar1
> # or this?
> String bar_string2 = bar2.value
> # or both?
> ```


Leaning a bit on my philosophy I laid out above regarding the core value of enums, there's two contexts you can work with an enum in this proposal:

- Within the WDL type system, wherein enumerations should be strictly typed and statically verifiable. So, if you are referring to an enum variant as a _type_, then it should be a unique, closed type that only exactly matches the enum variant itself—the value within is never even considered.
- Within an evaluation context (e.g., expressions or the command section), there is a one-way conversion of the enum variant to its inner value. This decision is made because it is asserted that you never want to resolve the name of an enum variant type _directly_ (and, if you do, the shorthand version where the value within is a stringified version of the variant name helps you).

DavyCats:
I think adding a new operator like `::` would make the language look more intimidating to non programmers (and maybe even more casual programmers not familiar with Rust or R). This would go against one of the central premises behind the design of WDL. I'm wondering whether this is even a necessary feature to add (obtaining a list of variants that is; enums would be very useful). Is there an actual use case for it?

If you do want to add it, would it be an idea to simply add a reserved `variants` member variable to the enum type? Essentially, enums would simply have an additional implicitly added variant which equals a list of all other variants.

eg.
```wdl
enum Animal {
    CAT,
    DOG
}

scatter (animal  in Animal.variants) {} # = [Animal.CAT, Animal.DOG]
```

Regarding coercion, one might be using a task which takes some integer as input. Perhaps some filtering threshold. The task and underlying tool support any integer value. However, in this particular workflow, only a handful of values might actually be supported:

```wdl
# task.wdl (Some third party developed task)
task filter_variants {
    input {
        Int threshold
        File vcf
    }
[...]

# workflow.wdl
import "https://example.com/task.wdl" as filter

enum FilterValue[Int] {
  STRICT = 10,
  NORMAL = 5,
  WHO_CARES = 1
}

workflow somatic {
    input {
        FilterValue stictness
        File tumor_fastq
        File normal_fastq
    }
[...]
    call filter.filter_variants {
        input:
            threshold = strictness, # coercion from FilterValue to Int
            vcf = vcf
    }
[...]
```

Would/should this be allowed?

claymcleod:
> I think adding a new operator like `::` would make the language look more intimidating to non programmers (and maybe even more casual programmers not familiar with Rust or R). This would go against one of the central premises behind the design of WDL. I'm wondering whether this is even a necessary feature to add (obtaining a list of variants that is; enums would be very useful). Is there an actual use case for it?
> 
> If you do want to add it, would it be an idea to simply add a reserved `variants` member variable to the enum type? Essentially, enums would simply have an additional implicitly added variant which equals a list of all other variants.
> 
> eg.
> 
> ```wdl
> enum Animal {
>     CAT,
>     DOG
> }
> 
> scatter (animal  in Animal.variants) {} # = [Animal.CAT, Animal.DOG]
> ```

I think this is a _great_ point about not being approachable. So to me, that kills `::` specifically. I do think there is going to be a use case for scattering over enum variants, but I'm also happy to make a case for it if we feel like we need to. In the meantime, I think there are three viable approaches to getting variants:

- `variants(FileType)`. I think this looks weird, but maybe it's okay.
- `FileType.variants()`. We don't call functions/methods in this way historically, and perhaps it even runs afoul of the point you made above regarding being non-approachable (because it complicates the ways in which you can call a function on something—now I have to know what context I'm in and call it the right way).
- `FileType.variants`. It feels strange to me to rather arbitrarily dictate that there is a field named `variants` that is reserved. It's also a bit magical to automatically get this _field_ assigned to the enum.

Overall, I think I'm leaning towards coming back to the first one—it looks a bit weird, but it's consistent with the way we've called functions in the past, and I think it's the least magical. Perhaps once we put in the right syntax highlighting, this wouldn't seem as weird as I'm envisioning.

> Regarding coercion, one might be using a task which takes some integer as input. Perhaps some filtering threshold. The task and underlying tool support any integer value. However, in this particular workflow, only a handful of values might actually be supported:
> 
> ```wdl
> # task.wdl (Some third party developed task)
> task filter_variants {
>     input {
>         Int threshold
>         File vcf
>     }
> [...]
> 
> # workflow.wdl
> import "https://example.com/task.wdl" as filter
> 
> enum FilterValue[Int] {
>   STRICT = 10,
>   NORMAL = 5,
>   WHO_CARES = 1
> }
> 
> workflow somatic {
>     input {
>         FilterValue stictness
>         File tumor_fastq
>         File normal_fastq
>     }
> [...]
>     call filter.filter_variants {
>         input:
>             threshold = strictness, # coercion from FilterValue to Int
>             vcf = vcf
>     }
> [...]
> ```
> 
> Would/should this be allowed?

In this case, as proposed, I think the automatic coercion would _not_ work out of the box—doing the coercion implicitly within the type system is probably not a good idea. There might be a case for a method called `value(<variant>)` that gets the value out of a variant to satisfy the condition you're talking about here.

peterhuene:
My preference is also for the first alternative (e.g. `variants(FileType)`), simply for consistency with the rest of WDL and to avoid adding new/overloaded syntax, but thinking about it from a programming language perspective, to treat it as a function call to `variants`, it would need to accept a _value_ in that position as functions take values. I would greatly prefer not to go down the road of having "special" functions that aren't of the production form `(value*) -> value`.

In this instance, `FileType` would be a name reference expression and normally we'd try to resolve that against local names (there's nothing special about _custom type names_ in WDL). For this to work, we'd have to also check the reference against custom type names and also prohibit a local name from conflicting with a custom type name; I'm not sure if the spec _already_ requires that, but right now `miniwdl` warns about the name conflict and `sprocket` does not emit a diagnostic.  The expression would then need to produce a "type reference" value which can be passed to the function.

Type references (i.e. the ability to reference a type as a runtime value) are not currently represented in WDL's value system.

The other syntaxes would not require modifying the value system to work as they could be accomplished by special-casing the access expression to see if we're operating on a name reference to a custom type and skip evaluation of the name reference expression (i.e. the expression `FileType` would therefore not need to produce a value).

It also means that the expression `(FileType).variants()` would not be equivalent as we'd treat that as a name reference to `FileType`, which wouldn't exist as a local name, assuming the conflicting name would be prohibited. This distinction can be seen in languages that use a single access operator for both static and instance member access (for example Java and C#).

geoffjentry:
One quick thought based on a skim: Given the primary domain in which WDL is used, `.variants()` maybe a bit too overloaded of a term

patmagee:
Given the contention that the `variants` function has, I wonder if we need to release the enum feature with it, or if we can release and gauge interest in that feature. If we see Workflows in the wilds that are doing something like: 

```wdl
scatter(type in [FileType.FASTQ,FileType.BAM]) {

}
```

Then we have a pretty good signal people actually want it and we can come back to it

claymcleod:
Yeah I think that's reasonable Patrick. So, I say we scratch the notion of enumerating variants for now and see whether people ask for it. As you show, it will be possible to enumerate over values explicitly like that, it's just annoying and not really scalable.

This also limits the impact of Geoff's observation, as the only time users will need to wrangle with them being called variants is in the documentation (I think variants is still the proper choice here, but I do agree that conflating the topics of variants in these two contexts on a day-to-day level are not helpful).

After that decision, I'm quite sure that all of the feedback has been addressed and now this is just ready for a review again.

claymcleod:
Thanks for the feedback! After a few months of thinking on this, I've come back with a fresh mind and made the following decisions.

@peterhuene, I've addressed your your concerns in the following ways:

1. Changed the `FavoriteNumber` example to correctly coerce to `Float` and added a new error example showing truly incompatible types.
2. Explicitly documented that enum values can be any WDL type, including compound types like `Array[String]` and `Map[String, Int]` (see example updates).
3. Kept the explicit type syntax enum `Foo[T]` to as you discussed if compound types were allowed.
4. Made clear that enum values must be literal expressions since enums are global.

Other improvements:

- Replaced the `.name` member with a `name()` standard library function to get the stringified variant name. This was needed after some thought regarding how a struct might have a `.name` member.
- Added `names()` and `variants()` functions as well.

patmagee (commented):
@claymcleod could you add a section on serialization and deserialization of enums? What value gets set? Is it always the name? Can we use the value for typed enums instead of the name? When we are writing TOO json, what is the behaviour?

claymcleod:
Apologies, I forgot to push my changes on the last comment.

@patmagee, I've added a section on serialization/deserialization. In brief, I think serializing things in "user land" (i.e., the JSON inputs and outputs) should be using the key while serializing enums in the command block should be using the inner value.

SPEC.md
```diff
@@ -3352,6 +3469,162 @@ struct Invalid {
...
+{
+  "example.result": "Red"
+}
+```
+
+The execution engine validates that the provided string matches one of the enum's variant names. If an invalid variant name is provided, the execution engine must raise an error during input validation.
+
+#### Command Section Serialization
+
+When an enum value is serialized in a command section using string interpolation, it is serialized to its **inner value** (not the variant name). This allows enum variants to have meaningful values for use in commands.
```
patmagee: You should probably say that it uses the typed typed value first, or the name of the variant if it is not a typed value.

Although, this type of am iguity leads me to wonder if explicit access to the variant is clearer here

SPEC.md
```diff
@@ -3352,6 +3469,162 @@ struct Invalid {
...
+    Boolean different_types = Pet.Mouse != ComputerDevice.Mouse
+  }
+}
+```
+
+### Enum Serialization and Deserialization
+
+Enum values are serialized and deserialized differently depending on the context.
+
+#### JSON Input and Output
```
patmagee: I also am not 100% sold on the asymmetrical approach here. It feels like command is the exception and not the rule. Which means someone will make a mistake.

It feels like there should be consistency to the usage and behaviour of the enum.


Another question then,in a task definition, how does one deserialize an enum?

In other languages there are things like `valueOf`. Also what is the result of `read_json()`?

SPEC.md
claymcleod: Technically in the current proposal, variants with no value are implicitly typed with `String` values of the same name as the variant. So it's not so much that there's ambiguity here as there is magic at work. But, that being said, perhaps explicit selection of the name or value is good. 

The open question then is really whether we want to do `variant.name`/`variant.value` or `name(variant)`/`value(variant)`. On my first look at it, I'm thinking the former to match `.left` and `.right` of pair, but curious what you think.

SPEC.md
claymcleod: I've responded to the part on the asymmetrical approach above.

WRT deserializing an enum in a task definition, I'm not sure what you mean by that. Can you give me an example of what you'd be trying to do?

SPEC.md
claymcleod: In fact, I think the most logical thing might actually be to default to serializing/deserializing the name and just allowing access to the value with `value(variant)`.

SPEC.md
```diff
@@ -3352,6 +3469,162 @@ struct Invalid {
...
+  Array[String] all_pet_names = names(Pet)
+
+  command <<<
+    echo "There are ~{length(all_pet_names)} kinds of pet: ~{sep(", ", all_pet_names)}"
+    echo "I have a pet ~{name(my_pet)}"
+    echo "My pet is a ~{my_pet}"
+  >>>
+
+  output {
+    Boolean different_types = Pet.Mouse != ComputerDevice.Mouse
```
patmagee, line 3573: @claymcleod what are the rules around equality? Would this always evaluate to `true` or Is this effectively equivalent to `value(Pet.Mouse) != value(ComputerDevice.Mouse)` which would be false?

patmagee (commented):
I left one more comment that leads to the question around clarifying equality. You mention type coercion but how does an Enum behave within an equality statement?

SPEC.md
claymcleod: In my mind, it is not equivalent to `value(Pet.Mouse) != value(ComputerDevice.Mouse)`, because the values within each variant do not constitute the entire identity of the value (though you can get the inner values out of variants and compare equality that way, if that's what you desire).

I can illustrate the core idea this way using existing valid WDL: consider two structs, each with a single member `foo: String` with a value `"bar"`; though the constituent data within the structs are identical, the fact that they are typed different necessitates the inequality because the type is part of the identity.

jdidion:
* I think it makes sense to refer to an enum as a kind of custom type. Technically, it would be most like a type alias with a restricted set of values.
* I am against introducing more implicit typing on principal, as I think we've seen some of these issues this can introduce and have been moving towards making WDL more explicit. I am fine with enum variants being able to take values that are defined using literals of different types, as long as those values are immediately coerced to the enum's underlying type. I think it should be required to always specify the type of an enum when there is any ambiguity.
    ```
    # eliding the type is ok as it's unambiguously `String`
    enum Simple { A, B }
    # eliding the type is not ok as variants are defined using literals of different types
    enum Ambig[Float] { X = 1, Y = 2.0 }
    # the type is unambiguously `Array[Integer]` so we could elide the type, but
    # I would still be in favor of requiring the type to be explicit for both clarity and
    # ease of implementation
    enum Shrug[Array[Integer]] { M = [1, 2, 3], N = [4, 5, 6] }
    ```
* Do we support coercion of an enum variant to it's underlying type? I think the answer should be no. I like the idea of an explicit `value` function. That way, different enum variants are unequal (thus answering your question about equality) whereas their values can be the same.
    ```
    enum Foo[String] { Val1 = "Hello", Val2 = "Hello" }
    
    workflow test {
        if (Foo.Val1 == Foo.Val2) { # evaluates to false }
        if (value(Foo.Val1) == value(Foo.Val2)) { # evaluate to true }
    }
    ```
* The `names()` and `variants()` functions are unlike any other in the standard library - I think it might be useful to introduce new notation, e.g. `Array[String] names(T)` where `T` is a custom type (`Struct` or `Enum`).
* I also don't think these functions are strictly necessary. We could instead use python-inspired notation:
    ```
    enum Foo {
        Bar = 1,
        Baz = 2
    }

    workflow test {
        # scatter over variants of Foo
        scatter (foo in Foo) {
            # get variant of Foo using dot notation
            if (foo == Foo.Bar) { ... }
            # get variant of Foo by name
            if (foo == Foo["Bar"]) { ... }
            # get name of Foo variant
            if (name(foo) == "Bar") { ... }
            # get value of Foo variant
            if (value(foo) == 1) { ... }
        }
    }
    ```

claymcleod:
> * I think it makes sense to refer to an enum as a kind of custom type. Technically, it would be most like a type alias with a restricted set of values.
> 
>     * I am against introducing more implicit typing on principal, as I think we've seen some of these issues this can introduce and have been moving towards making WDL more explicit. I am fine with enum variants being able to take values that are defined using literals of different types, as long as those values are immediately coerced to the enum's underlying type. I think it should be required to always specify the type of an enum when there is any ambiguity.
>       ```
>       # eliding the type is ok as it's unambiguously `String`
>       enum Simple { A, B }
>       # eliding the type is not ok as variants are defined using literals of different types
>       enum Ambig[Float] { X = 1, Y = 2.0 }
>       # the type is unambiguously `Array[Integer]` so we could elide the type, but
>       # I would still be in favor of requiring the type to be explicit for both clarity and
>       # ease of implementation
>       enum Shrug[Array[Integer]] { M = [1, 2, 3], N = [4, 5, 6] }
>       ```
> 
>     * Do we support coercion of an enum variant to it's underlying type? I think the answer should be no. I like the idea of an explicit `value` function. That way, different enum variants are unequal (thus answering your question about equality) whereas their values can be the same.
>       ```
>       enum Foo[String] { Val1 = "Hello", Val2 = "Hello" }
>       
>       workflow test {
>           if (Foo.Val1 == Foo.Val2) { # evaluates to false }
>           if (value(Foo.Val1) == value(Foo.Val2)) { # evaluate to true }
>       }
>       ```
> 
>     * The `names()` and `variants()` functions are unlike any other in the standard library - I think it might be useful to introduce new notation, e.g. `Array[String] names(T)` where `T` is a custom type (`Struct` or `Enum`).
> 
>     * I also don't think these functions are strictly necessary. We could instead use python-inspired notation:
>       ```
>       enum Foo {
>           Bar = 1,
>           Baz = 2
>       }
>       
>       workflow test {
>           # scatter over variants of Foo
>           scatter (foo in Foo) {
>               # get variant of Foo using dot notation
>               if (foo == Foo.Bar) { ... }
>               # get variant of Foo by name
>               if (foo == Foo["Bar"]) { ... }
>               # get name of Foo variant
>               if (name(foo) == "Bar") { ... }
>               # get value of Foo variant
>               if (value(foo) == 1) { ... }
>           }
>       }
>       ```

Everything you said here either (a) makes sense to me or (b) is already how the proposal lays things out.

claymcleod:
After some discussion with Peter, I've decided to walk back introducing the `names()` and `variants()` functions in WDL v1.3. Changing the language to support types themselves being passed into standard library functions seems like it might be a bit much to roll into this release, and these methods aren't strictly needed to get most of the value. We'll still have the `value()` method, since that _is_ needed and operates on a value instead of a type.

claymcleod:
This is now fully implemented here: https://github.com/stjude-rust-labs/sprocket/pull/445. I believe all of the feedback has been addressed.

SPEC.md
a-frantz: Given the current behavior around comparisons of different types, where Sprocket fails to validate `==` operations with different types, why would the above line be valid WDL? Although `Pet` and `ComputerDevice` are both `enum` they are different types, which should make their comparison invalid.

Example of the current spec behavior:
```console
$  cat tmp.wdl
version 1.2

workflow equality {
    input {
        Int five = 5
        String foo = "foo"
    }

    output {
        Boolean equals = five == foo
    }
}
$  sprocket check tmp.wdl
error: type mismatch: operator `==` cannot compare type `Int` to type `String`
   ┌─ tmp.wdl:10:26
   │
10 │         Boolean equals = five == foo
   │                          ^^^^^^^^^^^
   │                          │       │
   │                          │       this is type `String`
   │                          this is type `Int`

error: failing due to 1 error
```

----------

I also want to make sure I'm fully understanding this PR/comment thread: How does `value(Pet.Mouse) != value(ComputerDevice.Mouse)` evaluate? To `true` or to `false`? I expect the LHS and RHS to both evaluate to `"Mouse"`, making the expression valid WDL evaluating to `false` due to the `!=` operator.

a-frantz (commented):
Raised a question about one of the examples and what expressions with `enum`s will be allowed with this change, but overall this looks great to me!

SPEC.md
```diff
@@ -1485,6 +1495,112 @@ Example output:
 
 Note that the ability to assign values to `Struct` declarations other than struct literals is deprecated and will be removed in WDL 2.0.
 
+##### Enumeration Types (Enums)
```
peterhuene, line 1499: Some place in this section I think we need a paragraph on what constitutes a _compatible definition_ of an enum for the purpose of resolving name conflicts in imported types.

I would like the order of the variant declarations to be relevant for determining compatible definitions in case we add the ability to reflect on an enum's type (e.g. a hypothetical `values` function that takes the enum type and returns `Array[T]`) in the future.

SPEC.md
```diff
@@ -3352,6 +3468,149 @@ struct Invalid {
...
+}
+```
+
+An enum can be thought of as a closed type with a fixed set of instances. The `enum` keyword creates both a type (that can be used in declarations) and a global namespace containing the enum's variants. For example, `Color.Red` refers to a specific instance of the `Color` enum type.
+
+Unlike structs, it is not possible to create new instances of an `enum` outside of the enum's definition. An enum value can only be one of the variants defined in the enum's declaration.
+
+### Enum Usage
+
+An `enum`'s variants are [accessed](#member-access) using a `.` to separate the variant name from the `enum`'s identifier.
```
peterhuene: So this is tricky to implement properly without either further expansion of the _value space_ of WDL or specifically defining the access (dot) expression grammar to support enumerations (however, the latter is _ambiguous_, I'll explain why in a sec).

In the current grammar of WDL, the access expression is effectively:

```ebnf
access-expr := expr '.' ident => value
```

Given that the LHS of the expression is itself an expression, the expression must produce a _value_.

The expression could be a _name reference_ expression (e.g. `foo`), which produces the _value_ of the name from the current scope.

A type name (e.g. a `Color` enum) doesn't have a corresponding _value_ in WDL. We would have to extend the WDL value space by adding a "type reference" value to be able to access it with the _current_ WDL grammar rules.

One might attempt to _special case_ the access expression by amending the rule to be:

```ebnf
access-expr := (ident | expr) '.' ident => value
```

However, that is ambiguous because `expr` can be `ident` in the case of a name reference expression so it is unclear from the grammar alone whether it is a special-case access on a type name or an access on a named _value_.

We _could_ define the access expression in terms of "if the LHS `expr` happens to _directly_ be a name reference expression that names an enum type, access it; otherwise, if it's any other expression, evaluate it and access the produced _value_" but that seems brittle to me (or, at the very least, should be specifically discussed for the semantics of the access expression).

Another alternative would to _not_ use the dot operator here, but instead some other novel syntax specifically for accessing a type (e.g. `Color::Red`) where the grammar rule could be:

```ebnf
variant-ref-expr := ident '::' ident => value (specifically an enum value)
```

peterhuene: Related to this issue, currently one may have a declaration name that _shadows_ a custom type name, as the naming rules for declarations is specified as:

> Each declaration must be unique within its scope, and may not collide with a [reserved WDL keyword](https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#reserved-keywords) (e.g., workflow, or input).

Currently Sprocket accepts this (although `miniwdl` does not):

```wdl
version 1.3

struct Foo {
    Int x
}

workflow example {
    input {
        # `Foo` is unique within its scope and isn't a reserved keyword 
        String Foo
    }
}
```

Unless I missed somewhere else in the spec that says a declaration name cannot shadow a custom type name, then we _definitely_ need to add such language to the spec to properly support using an access expression on an enum, otherwise:

```wdl
version 1.3

enum Foo {
    Foo,
    Bar,
    Baz,
}

struct Bar {
    Int x
}

workflow example {
    input {
        Bar Foo = Bar { x: 1 }
    }

    Int x = Foo.x # ???
}
```

SPEC.md
peterhuene: I am also of the opinion that we should treat the above comparison as a _static_ type mismatch and thus _not evaluable_, similar to what Ari said and also:

```
version 1.1

struct Foo {
    Int x
}

struct Bar {
    Int x
}

workflow test {
    Foo a = Foo { x: 1 }
    Bar b = Bar { x: 1 }

    output {
        Boolean eq = a == b
    }
}
```

SPEC.md
```diff
@@ -3352,6 +3468,149 @@ struct Invalid {
...
+
+A declaration with an `enum` type can only be initialized by referencing a variant directly or by assigning it to the value of another declaration of the same `enum` type.
+
+Two enum values can be tested for equality (i.e., using `==` or `!=`). To be equal, two enum values must be the same variant of the same `enum` type. For example, `Color.Red == Color.Red` evaluates to `true`, while `Color.Red == Color.Blue` evaluates to `false`. Even if two different enum variants have the same inner value, they are not equal (though their inner values extracted via `value()` may be equal). Enum variants are not ordered, so they cannot be compared (i.e., using `>`, `>=`, `<`, `<=`).
```
peterhuene:
```suggestion
Two enum values can be tested for equality (i.e., using `==` or `!=`). To be equal, two enum values must be the same variant of the same `enum` type. For example, `Color.Red == Color.Red` evaluates to `true`, while `Color.Red == Color.Blue` evaluates to `false`. A comparison of two enum values of different `enum` types is considered a type mismatch error. Enum values are not ordered, so they cannot be compared with ordinal operators (i.e., using `>`, `>=`, `<`, `<=`).
```

SPEC.md
peterhuene: I'd also suggest adding a line to this paragraph to the effect of:

> The access expression produces an _enum value_ specific to the named variant and not the value associated with the variant (see the [`value()`](#-value) standard library function for retrieving a variant's associated value).

I think that might help clear up some confusion.

claymcleod: One thing to consider was this comment from @DavyCats:

> I think adding a new operator like `::` would make the language look more intimidating to non programmers (and maybe even more casual programmers not familiar with Rust or R).

I tend to agree with that, but I don't know what a better operator would be.

peterhuene: I'm okay with using the access expression, but we need to better codify how it's expected to work and that it's a bit of a hack around the grammar (granted, we're not terribly formal with the WDL grammar definition to begin with).

What we could do to make it less of a hack is to formalize the name reference expression and extend the WDL value space with a "type reference value". If the name refers to a name in scope, the expression produces the value of that name. If the name refers to a visible _type name_, then it produces a _type reference value_. Otherwise, it's an unknown name error.

The access expression would then work exactly like it does today for accessing a field of a value, except the addition of enumerations means one can access a type reference value of an enum by variant name to produce an _enum value_.

This would also allow us to define functions such as `variants(T) -> Array[String] where T: EnumTypeRef` (i.e. any type reference value referring to an enum; an example call to the function would be `variants(Color)`).

peterhuene: After some discussion with Clay, I think the best approach for the initial feature is:

* Use `.` for accessing an enum to produce an _enum value_ (we already use `.` for "accessing" namespaces in call statements, for example).
* Explicitly call out in this section that the LHS expression of the access expression must:
  * Be _exactly_ a name reference expression to a name that is _not_ in scope but is a visible enum type name.
  * Currently prohibit an expression like `(Color).Red` due to the above requirement; one would expect the inner name reference expression to error if `Color` isn't a name _in scope_ (for now at least).
 * Do not yet introduce a "type reference" value into the WDL value system in 1.3.

Like with how we punted on a `variants` stdlib function, I think the above would allow us to later introduce a "type reference" value in a backwards compatible way; if suddenly the expression `Color` produced such a value, then the expression `Color.Red` would behave like the special-cased name reference access defined in this proposal (additionally, `(Color).Red` would then be a semantically identical expression).

For future consideration, an alternative to introducing a "type reference value" into the value system would be to support methods (i.e. stdlib functions scoped to a type, e.g. `Color::variants()` or some such).

SPEC.md
claymcleod: Okay, I suppose I'm in favor of enforcing the order of variants when considering whether two definitions are equal. I'll add this language in.

SPEC.md
claymcleod: Yeah, that makes sense—I must have missed coming back to this and updating after my Sprocket implementation (where this does not compile), so I'll update this.

And yes, Ari, your understanding of the evaluation using `value` is correct.

SPEC.md
patmagee: I like this as a compromise, especially as it allows us to ship enums without closing out the future to the improvements you are suggesting

SPEC.md
patmagee: do we specify anywhere the expected order order of import resolution? I would imagine in order for this to be true, we would need some sort of engine agnostic and deterministic heuristic for evaluating a WDL

patmagee (approved):
Incredible work! I think this is going to be a game changer to the WDL community!!

SPEC.md
```diff
@@ -1485,6 +1495,112 @@ Example output:
 
 Note that the ability to assign values to `Struct` declarations other than struct literals is deprecated and will be removed in WDL 2.0.
 
+##### Enumeration Types (Enums)
+
+An enumeration (or "enum") is a closed set of enumerated values (known as "variants") that are considered semantically valid in a specific context. An enum is defined at the top-level of the WDL document and can be used as a declaration type anywhere in the document.
```
geoffjentry: re `variant` - couple of thoughts here:

1) Given the domain we're in, this feels like it could get confusing due to terminology overload
2) It feels very Rust-y. Traditionally the model has been to use Python as a guide for terminology and syntax

vsmalladi: I agree I would change to not be variants. Maybe `options` or `allowed values`

SPEC.md
vsmalladi: Ya can we get some clarification.

SPEC.md
claymcleod: Agree. I'll find something less Rust-y.

SPEC.md
claymcleod: In the implementation in Sprocket, we ended up just going the whole 9 yards and (a) formalizing type name references and (b) allowing these to be evaluated as part of expressions. This was largely because hacking in access expressions for these seemed weird and was ultimately probably going to be more difficult (i.e., internal accounting) than just doing it the "proper way". So I'm going to amend the RFC to formalize that concept and make clear that type name references are a new part of the type system.

I should note that I also surveyed the code for miniwdl and Cromwell before making this decision—both of them have code similar to Sprocket's for type such that I imagined this implementation would be easier than hacking it in. So there was due diligence done there.

SPEC.md
claymcleod: I've been thinking about it, and the two best choices (no pun intended) are:

- "Member", which is what Python uses. Member does conflict logically with members of a struct though which operate differently.
- "Choice", which is a bit informal but clearly defines the idea while using a word not already claimed in the spec.

For now, I've updated this to "choice" but I'm open to feedback on this (or if there are any other names people think are better).

claymcleod:
Whoops, it looks like a glitch in the GitHub UI meant @patmagee was re-requested on accident. I'm going to go ahead and count his approval here.

SPEC.md
peterhuene: I don't have a strong preference, but if I had to rank those choices, I think my ranking would be:

1. `variant`
2. `choice`
3. `member`

I can acknowledge that `variant` seems overly formal, but I also think that `member` isn't quite right either. `choice` seems like an okay compromise that conveys it being specifically related to an enum.

Issue #697, opened 2025-01-20
`Else` statements


claymcleod:
I recently discovered that WDL does not have `else` statements. To achieve the same effect, it's suggested [by the spec](https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#conditional-statement) to use two `if` statements: one with the conditional you desire, and then one with the negated conditional.

Searching through the issues, I didn't immediately find an explanation as to why we can't have `else` statements. Is there a good reason for this that I just can't find?

geoffjentry:
I actually remember this one. 

It's because there were a lot of strong motivating use cases (in terms of scientific workflows people were actually trying to run at the time) for "optionally run a task" but very few for deeper branching logic. Because that impact prioritization plus how it would complicate both WDL & implementation and that "design a full programming language" was an explicit non-goal we left it at just the `if`. 

Most of these boil down to some form of there not being enough of motivating use cases to be worth the extra effort. Sometimes this was IMO for the better (e.g. the lack of loops: 10 years later and still doesn't come up hardly at all) and some for the worse (e.g. `Pair` which only exists as a half-ass on richer object types due to key use cases at the time only needing pairs of things)

claymcleod:
Yeah that totally makes sense, thank you for the context and the history here.

I don't know if you've been following the work I'm doing [on the `wdl-docs` repository](https://github.com/openwdl/wdl-docs/pull/33), but I definitely think that this minimalism is a strength (in particular, [this section](https://claymcleod.github.io/wdl-docs/overview.html#antivalues)). I totally agree on loops that they are probably not needed.

Though I 100% agree that WDL should remain simple and should not often "purchase" flexibility at the cost of more complexity, to my eye, `else` would add a relatively very small amount of complexity while enabling a moderately useful design pattern in a more straightforward way. For example, doing things like in the [the second example here](https://claymcleod.github.io/wdl-docs/language-guide/workflows.html#conditional-statement) are probably going to be much more straightforward with an `else` clause.

patmagee:
I think an else statement is a complimentary pattern to an if statement and it is really awkward to have one and not the other. If we are going to allow branching, we should not shortchange the end user IMO.

Without it, I will often see things like:

```wdl
if ( condition x){

}

if (not condition x){

}

```

geoffjentry:
A question I'd have is: if one were to survey the world's breadth of public WDLs, how often do you see the above?

[...]

PR #699, opened 2025-01-26
feat: adds `else if` and `else` conditional clauses

### LINKED ISSUE #268: else?
I noticed 1) works but not 2). Can we have else for conditional blocks (not just one-liners)?
### 1)
```
workflow merge {
    Array[File] smallfiles = []
    if(length(smallfiles)>2) {
        call paste {input: files = smallfiles}
    }
    if(length(smallfiles)<=2) {
        call cat {input: files = smallfiles}
    }
}
```

### 2)
```
workflow merge {
    Array[File] smallfiles = []
    if(length(smallfiles)>2) {
        call paste {input: files = smallfiles}
    }
    else {
        call cat {input: files = smallfiles}
    }
}
```
```
ERROR: Unexpected symbol (line 6, col 5) when parsing '_gen10'.
Expected rbrace, got else.
```

claymcleod:
This change proposes extending conditional statements to include `else if` and `else` clauses.

These types of clauses have been proposed many times on Slack and more formally in #268 and #697. Overall, the feedback seems positive, though concerns have been raised as to whether the cost of added complexity is worth it (see #697 for more details). I have made my case at the end of #697 that I feel it's actually _removing_ mental burden/complexity to add these clauses into the language.

TODO:

- [ ] Clarify that braces can be used in ternary expressions.
- [ ] Covers both expressions and workflow statements.
- [ ] Update logic to derive non-optional and optionals based on @cjllanwarne's comment.

Closes #268.

- [x] You have updated the `README.md` or other documentation to account for these changes (when appropriate).

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -7146,11 +7146,11 @@ Example output:
 
 ### Conditional Statement
 
-A conditional statement consists of the `if` keyword, followed by a `Boolean` expression and a body of (potentially nested) statements. The conditional body is only evaluated if the conditional expression evaluates to `true`.
+A conditional statement consists of one or more conditional clauses. Each conditional clause is comprised of an expression that evaluates to a `Boolean` and an associated clause body. When a conditional statement is executed, each of the conditional clauses is evaluated sequentially: (a) the expression for that clause is evaluated and, if the returned value is `true`, the body of that clause is executed and the entire conditional statement suspends further execution. The simplest conditional statement contains a single "if" clause, which is the `if` keyword, followed by the clause's boolean expression, and, finally, the clause body of (potentially nested) statements wrapped in curly brackets.
 
 After evaluation of the conditional has completed, each declaration or call output in the conditional body is exposed in the enclosing context as an optional declaration. In other words, for a declaration or call output `T <name>` within a conditional body, a declaration `T? <name>` is implicitly available outside of the conditional body. If the expression evaluated to `true`, and thus the body of the conditional was evaluated, then the value of each exposed declaration is the same as its original value inside the conditional body. If the expression evaluated to `false` and thus the body of the conditional was not evaluated, then the value of each exposed declaration is `None`.
 
-The scoping rules for conditionals are similar to those for scatters - declarations or call outputs inside a conditional body are accessible within that conditional and any nested statements.
+The scoping rules for conditionals are similar to those for scatters—declarations or call outputs inside a conditional body are accessible within that conditional and any nested statements.
 
 In the example below, `Int j` is accessible anywhere in the conditional body, and `Int? j` is an optional that is accessible outside of the conditional anywhere in `workflow test_conditional`.
 
@@ -7231,7 +7231,9 @@ Example output:
 </p>
 </details>
 
-WDL has no `else` keyword. To mimic an `if-else` statement, you would simply use two conditionals with inverted boolean expressions. A common idiom is to use `select_first` to select a value from either the `if` or the `if not` body, whichever one is defined.
+After the initial `if` clause, conditional statements may have any number of `else if` clauses and a single, final `else` clause. As described in the paragraph above, `else if` clauses are only evaluated if all prior clauses in the statement have evaluated to `false`. If present, the final `else` clause executes only if all prior clauses evaluated to `false`.
+
+When gathering results from conditionals, a common idiom is to use `select_first` to select the defined value from the `if`, `else if`, or `else` body that was evaluated.
 
 <details>
 <summary>
@@ -7246,7 +7248,7 @@ task greet {
   }
 
   command <<<
-  printf "Good ~{time} buddy!"
+    printf "Good ~{time} buddy!"
   >>>
 
   output {
@@ -7259,13 +7261,11 @@ workflow if_else {
     Boolean is_morning = false
   }
   
-  # the body *is not* evaluated since 'b' is false
   if (is_morning) {
+    # The body *is not* evaluated since `is_morning` is `false`.
     call greet as morning { time = "morning" }
-  }
-
-  # the body *is* evaluated since !b is true
-  if (!is_morning) {
+  } else {
+    # The body *is* evaluated since the clause above did not trigger.
     call greet as afternoon { time = "afternoon" }
   }
 
--- a/SPEC.md
+++ b/SPEC.md
@@ -7158,7 +7158,14 @@ A conditional statement consists of one or more conditional clauses, each having
 
 When a conditional statement is evaluated, each conditional clause is evaluated sequentially; for each `if` and `else if` clause, the expression is evaluated—if the result of the evaluation is `true`, the body of that clause is evaluated and the entire conditional statement suspends further evaluation. If none of the `if` or `else if` clauses execute and we reach the final `else` clause, the `else` clause is executed and the conditional suspends further evaluation.
 
-The declarations promoted to the parent scope depend on a union of the scopes for each conditional statement clause using the following algorithm:
+The declarations and call outputs promoted to the parent scope depend on a union of the scopes for each conditional statement clause:
+
+- Declarations and call outputs that are made available under every condition, including the exhaustive `else` clause, are promoted to the parent scope as their declared type.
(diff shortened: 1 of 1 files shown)


a-frantz:
> I feel it's actually _removing_ mental burden/complexity to add these clauses into the language.

I agree with this, though I'll admit my programming background bias. I find it more complicated to create two unconnected `if` clauses with negated expressions than it would be to simply have an `else` (or `else if`) clause.

@geoffjentry said:
> A question I'd have is: if one were to survey the world's breadth of public WDLs, how often do you see the above?

And I can point to at least one instance of this pattern in our `workflows` repository - https://github.com/stjudecloud/workflows/blob/main/workflows/qc/quality-check-standard.wdl#L351-L386

SPEC.md
```diff
@@ -7146,11 +7146,11 @@ Example output:
...
 
 After evaluation of the conditional has completed, each declaration or call output in the conditional body is exposed in the enclosing context as an optional declaration. In other words, for a declaration or call output `T <name>` within a conditional body, a declaration `T? <name>` is implicitly available outside of the conditional body. If the expression evaluated to `true`, and thus the body of the conditional was evaluated, then the value of each exposed declaration is the same as its original value inside the conditional body. If the expression evaluated to `false` and thus the body of the conditional was not evaluated, then the value of each exposed declaration is `None`.
```
peterhuene: I feel like this paragraph probably needs to be expanded upon.

It should probably read that the declarations within the `if`, `else if`, or `else` body that executes (depending on the evaluation of the conditional expressions) will be assigned to the values of their declarations; the declarations in the bodies that did not execute will be `None`.

SPEC.md
```diff
@@ -7231,7 +7231,9 @@ Example output:
 </p>
 </details>
 
-WDL has no `else` keyword. To mimic an `if-else` statement, you would simply use two conditionals with inverted boolean expressions. A common idiom is to use `select_first` to select a value from either the `if` or the `if not` body, whichever one is defined.
+After the initial `if` clause, conditional statements may have any number of `else if` clauses and a single, final `else` clause. As described in the paragraph above, `else if` clauses are only evaluated if all prior clauses in the statement have evaluated to `false`. If present, the final `else` clause executes only if all prior clauses evaluated to `false`.
+
+When gathering results from conditionals, a common idiom is to use `select_first` to select the defined value from the `if`, `else if`, or `else` body that was evaluated.
```
cjllanwarne: IIRC, this clunkiness is the reason I shied away from `else` originally. It's... fine. But I always felt like committing forever to a clunky system like this would be disappointing. So... I punted.

Now that I don't actually have to implement what I'm suggesting, I feel more free to say that the ideal situation - IMO - would be more like:

1. Values (vals and call outputs) which are sure to be produced in every case are automatically gathered for you
2. Values which are sometimes produced are returned as optionals (or, it could be a syntax error, but I think returned as optional is nicer)

eg
```wdl
if (some_bool) {
  Int a = 5
  Int b = 6
  call x { a = a, b = b}
} else {
  Int a = 100
  call x { a = a, b = 6}
}

# Out here, a is an Int and b is an Optional[Int]. And any x.out would also be non-optional
```

claymcleod: Yeah, that makes sense. I think I'll go with this instead.

SPEC.md
claymcleod: I've added these details in.

SPEC.md
claymcleod: In my latest draft, I've included this behavior.

claymcleod:
@peterhuene and @cjllanwarne, there still may be some polishing needed on the text, but let me know how this reads generally.

claymcleod:
I've now prototyped the behavior indicated in @cjllanwarne's comment in this PR: https://github.com/stjude-rust-labs/sprocket/pull/411. If you'd like to try it out, you can by installing Sprocket from the associated branch.

```bash
cargo install --git https://github.com/stjude-rust-labs/sprocket.git --branch cmcleod/if-else sprocket
```

This does require (a) the WDL version to be `version 1.3` and (b) a `sprocket.toml` with the following line:

```toml
run.experimental_features_enabled = true
```

SPEC.md
```diff
@@ -7146,9 +7146,56 @@ Example output:
...
+    * If the name _isn't_ already in the map, insert the name into the map and assign the type seen.
+    * If the name _is_ already in the map, update the mapped type to a common type between the current declaration's type and the type stored in the map. If there is no common type, emit an error.
+2. Perform a second pass through each clause in the conditional statement. For each name in the map created in step 1, if that name is _not_ seen in the current clause's scope, mark that type as optional.
+3. If there is no `else` clause, mark every type in the map as optional.
+
+The result is a set of declarations available in the parent scope that concretely represent the union of all scopes of the conditional statement. Any declaration in the union map that does not evaluate in a conditional statement clause's body is set to `None`. Further, when finding common types across scopes, the type declared in the earliest conditional statement clause is used as the base type. If a declaration that _would_ be promoted to a parent scope conflicts with an existing name in the parent scope, an error should be returned.
+
+For example,
```
cjllanwarne: Could be worth showing a task call in the example too to make it clear that this applies to calls as well as values?

SPEC.md
```diff
@@ -7146,9 +7146,56 @@ Example output:
...
+  evaluates to a `Boolean`. If present, `else if` clauses must follow the `if`
+  clause and be before the optional `else` clause.
+* At most, one `else` clause with no associated expression. The `else` clause
+  must be last in the conditional expression.
+
+When a conditional statement is evaluated, each conditional clause is evaluated sequentially; for each `if` and `else if` clause, the expression is evaluated—if the result of the evaluation is `true`, the body of that clause is evaluated and the entire conditional statement suspends further evaluation. If none of the `if` or `else if` clauses execute and we reach the final `else` clause, the `else` clause is executed and the conditional suspends further evaluation.
+
+The declarations promoted to the parent scope depend on a union of the scopes for each conditional statement clause using the following algorithm:
+
+1. Create a new map of declaration names to types. Traverse all clauses in the conditional statement, gathering the declarations in the scope into a mapping of declaration names to types. For each clause:
```
cjllanwarne, line 7170: This section is a little more in-the-weeds procedural for my taste, rather than stating what the outcome should be. Which is fine as a style choice but might be worth mentioning that. Eg we could say that this is one way to make things work, and is included to get the point across, but an engine that implements this algorithm a different way is obviously not "incorrect" per the spec as long as the outcome is the same.

cjllanwarne (commented):
Looks good to me, just a couple of minor style nits. I do think mentioning that it works for call results as well as declared values is worth making explicit with an example though.

claymcleod:
@cjllanwarne what do you think about these revisions?

peterhuene (approved):

SPEC.md
```diff
@@ -7146,11 +7146,69 @@ Example output:
...
+When a conditional statement is evaluated, each conditional clause is evaluated sequentially; for each `if` and `else if` clause, the expression is evaluated—if the result of the evaluation is `true`, the body of that clause is evaluated and the entire conditional statement suspends further evaluation. If none of the `if` or `else if` clauses execute and we reach the final `else` clause, the `else` clause is executed and the conditional suspends further evaluation.
+
+The declarations and call outputs promoted to the parent scope depend on a union of the scopes for each conditional statement clause:
+
+- Declarations and call outputs that are made available under every condition, including the exhaustive `else` clause, are promoted to the parent scope as their declared type.
+- Declarations and call outputs that are missing from one or more clauses, are declared as optional is one more clauses, or are missing from the exhaustive `else` clause are promoted to the parent scope as optional versions of their declared type.
```
a-frantz:
```suggestion
- Declarations and call outputs that are missing from one or more clauses, are declared as optional in one more clauses, or are missing from the exhaustive `else` clause are promoted to the parent scope as optional versions of their declared type.
```

a-frantz (changes requested):
found a typo, otherwise looks awesome 👍

patmagee (approved):

SPEC.md
tefirman:
```suggestion
- Declarations and call outputs that are missing from one or more clauses, are declared as optional in one or more clauses, or are missing from the exhaustive `else` clause are promoted to the parent scope as optional versions of their declared type.
```

tefirman (approved):
Excited about this addition! 🙌

PR #728, opened 2025-10-15
feat: adds the `preemptible` task hint

claymcleod:
This PR proposes the formalization of the `preemptible` hint, which is already supported in both Sprocket and [Cromwell](https://cromwell.readthedocs.io/en/stable/RuntimeAttributes/#preemptible) (albiet as a runtime key).

Briefly, `preemptible` aims to indicate the number of times a preemptible instance _may_ be tried before using a non-preemptible instance. Failures other than preemption are proposed to count against the `max_retries` requirement instead.

      changes (when appropriate).
- [ ] You have added or updated relevant example WDL tests to the specification.


SPEC.md
```diff
@@ -5562,6 +5563,25 @@ Reserved input-specific attributes:
...
+A hint to the execution engine that a task _may_ be tried using a preemptible instance the specified number of times. Engines that are not configured to use or do not support preemptible instances may ignore this parameter and the implied number of retries completely.
+
+Engines should only count failed tasks against the `preemptible` count if the reason for the failure was preemption. Other failures, such as unexpected non-zero exit codes, should be counted against the [`max_retries`](#max_retries) requirement instead.
+
+A value of `0` means the task should not be tried with a preemptible instance.
+
+```wdl
+task preemptible_example {
+  hints {
+    preemptible: 3
```
patmagee, line 6178: So I am interested in opening the can of worms here and wonder if we should move to a model of

- `preemptible` is boolean
- `max_retries` counts as the number of attempts (even from preemptible failures)
- If the task cannot complete on preemptible instances it fails UNLESS an explicit flag is set in the hints to toggle this behaviour and have the task complete on a non preemptible instance instead. This of course ONLY applies if the targeted engine supports preemptible instances

Rational:

The typical behaviour has been beneficial for sure.. but it kind of represents the worst case scenario. You have to pay for up to N preemptions PLUS the cost of the non preemptible instance. If people have no problem with this and we just want to leave the recommended behaviour as is then that is cool.

claymcleod:
Note that this was already informally implemented in Sprocket, as I believe we just copied how Cromwell handles this key. Thus, there is no associated PR—you can just grab the latest published version of Sprocket and try it out. It should work for any TES server that implements pre-emptible instances at a minimum.

SPEC.md
claymcleod: I see where you're coming from, and idealistic me would like to consider it, but I'm hesitant to pull into the spec a _conflicting_ definition of how Cromwell thinks about preemptible instances (even though it is technically proper with these being hint). I feel this way, if for no other reason, just because of the unanticipated headaches most will encounter when porting their 1.0 WDL scripts to 1.2. 

Perhaps there is a middle ground model where:

- `preemptible` is the number of preemptible retries
- `max_retries` is the number of non-preemptible retries
- there's a new hint that indicates whether _any_ non-preemptible instances should occur (effectively turns off `max_retries`).

Honestly, if we're going to go all in on a better hint, I might rather call it something altogether (indicating a more sharp deviation from the Cromwell `preemptible` key)... something like

```wdl
hints {
  execution_plan: ["preemptible:3", "non-preemptible:1"]
}
```

patmagee (approved):

SPEC.md
```diff
@@ -5562,6 +5563,25 @@ Reserved input-specific attributes:
...
 Provides output-specific hints. Each key must refer to a parameter defined in the task's [`output`](#task-outputs) section. A key may also use dotted notation to refer to a specific member of a struct output.
 
+##### ✨ `preemptible`
+
+* Accepted types: `Int`
+* Default value: `0`
+
+A hint to the execution engine that a task _may_ be tried using a preemptible instance the specified number of times. Engines that are not configured to use or do not support preemptible instances may ignore this parameter and the implied number of retries completely.
+
+Engines should only count failed tasks against the `preemptible` count if the reason for the failure was preemption. Other failures, such as unexpected non-zero exit codes, should be counted against the [`max_retries`](#max_retries) requirement instead.
```
geoffjentry: `Engines should only count`

I think `should` is technically correct here, but given the follow on context I believe it is really a `must`. `should` implies recommended behavior, but the full sentence is explaining the only reason an engine would not do this.

SPEC.md
```diff
@@ -5562,6 +5563,25 @@ Reserved input-specific attributes:
...
+
+A value of `0` means the task should not be tried with a preemptible instance.
```
geoffjentry: `must not be tried` or `shall not be tried`

SPEC.md
geoffjentry: Speaking generally about bfx workflows, the most common scenario I see people try to set up looks like:

1) Try the preemptible instance up to N times
2) If that still fails, fall back to full price instance
3) However, under no circumstances run more than M times total

I think that's fully achievable with just `preemtible` and `max_retries`

vsmalladi: I agree that is achievable. The assumption is the following.

`non-preemptible` = `max_retries` - `preemptible`

Where @geoffjentry  lays out the logic above

I wonder if we should be more flexible. Since one might not want to start `preemptible` nodes first if costs more than x amount.

geoffjentry: `Since one might not want to start preemptible nodes first if costs more than x amount.`

@vsmalladi Can you describe that further? I'm parsing this a few different ways. What use case(s) are you picturing here and how are you thinking it'd work?

vsmalladi: I think there is an argument to be made that a user might have an lower cost for dedicated nodes they want to try to use, but if those are full or the preemptible  also are not meeting a particular cost relative to dedicated to default to dedicated. So we might want engines to be able to read these variables to determine how they want to proceed for the user rather than default with logic preemptible first then dedicated.

geoffjentry: @vsmalladi I **think** you're suggesting what I was curious about. Ignore me if this isn't what you're saying. I don't think it's a good idea to have the WDL spec require understanding of underlying cost & apply logic in that regard. I do think enough vendors have something that maps to `preemptible ` such that it's worth the flag, but I don't agree w/ going down the road of baking higher level cost smarts into the spec.

To me that's something that if an engine **really** wanted to do it it could be a hint.

vsmalladi: @geoffjentry  I agree. Which is why having the rules on `preemptible` running before other `dedicated` types baked into the spec feels off to me. I just want both flags and engines can determine how to run those orders.

claymcleod:
There is considerable debate about exactly how this should be implemented. Given this, I think it's best to exclude from the WDL 1.3 release. At this point, I'm not even positive I feel this feature should be stabilized in this exact form—the problem begs for a more formal specification IMO.

vsmalladi:
I agree @claymcleod  This should be moved to a 1.4 maybe

claymcleod:
I've been thinking about this more, and I agree with Geoff's general direction. I also agree with Venkat that we want more flexibility in the future, but it's not a regression to introduce this approach first before a more complex solution—we need time to figure out what that should look like.

Given that, and the fact that we can formalize what both Sprocket and Cromwell do today, I think it's okay to move forward. A future `execution_plan` hint (or similar) could explicitly override these defaults when we're ready for that.

claymcleod:
And, to be clear, I'm still in favor of just bumping this to 1.4.

claymcleod:
If this isn't also clear from the answer above, I do think there is a world where we don't need a more complex way to outline an execution plan. We'll have to find that out as we go through the design process.

SPEC.md
claymcleod: I've updated the spec to implement Geoff's proposed flow: (1) try preemptible up to N times, (2) fall back to non-preemptible if all preemptible attempts fail due to preemption, (3) `max_retries` serves as the total cap on all attempts. This formalizes what Sprocket and Cromwell already do today.

We can always propose a more complex hint (e.g., `execution_plan`) in the future that would supersede this behavior if we find the need for more flexibility.

SPEC.md
```diff
@@ -6151,6 +6152,34 @@ Reserved input-specific attributes:
 
 Provides output-specific hints. Each key must refer to a parameter defined in the task's [`output`](#task-outputs) section. A key may also use dotted notation to refer to a specific member of a struct output.
 
+##### ✨ `preemptible`
+
+* Accepted types: `Int`
+* Default value: `0`
+
+A hint to the execution engine that a task _may_ be tried using a preemptible instance up to the specified number of times. If all preemptible attempts fail due to preemption, the engine must fall back to executing on a non-preemptible instance (up to the limit imposed by `max_retries`). Engines that are not configured to use or do not support preemptible instances may ignore this hint entirely.
```
patmagee, line 6160: I feel like we should be clear here what a "preemptible" instance is.. at one point it went without saying since the clouds defined that behaviour for us, but now I think it would be worthwhile specifying it.

SPEC.md
```diff
@@ -6151,6 +6152,34 @@ Reserved input-specific attributes:
 
 Provides output-specific hints. Each key must refer to a parameter defined in the task's [`output`](#task-outputs) section. A key may also use dotted notation to refer to a specific member of a struct output.
 
+##### ✨ `preemptible`
+
+* Accepted types: `Int`
+* Default value: `0`
```
patmagee, line 6158: should we say something about the absence of preemptible indicates is the same as the default value `preemptible: 0`

claymcleod:
@patmagee @geoffjentry @vsmalladi I have redesigned this proposal for WDL 1.4 by separating the preferred number of preemptible attempts from the requirement to use preemptible instances exclusively.

The new Boolean `preemptible_only` requirement defaults to `false`. When it is `true`, the engine must honor a positive `preemptible` value and must never run the task on a dedicated instance. Using `preemptible_only: true` without a positive `preemptible` value is a validation error. An engine that cannot provide a preemptible instance must fail the task.

Preemptions now count only against `preemptible`, while retries caused by all other failures count only against `max_retries`. When `preemptible_only` is `false`, starting a dedicated attempt after exhausting the preemptible attempts consumes one retry, so `max_retries: 0` does not permit that transition. When `preemptible_only` is `true`, exhausting the preemptible attempts fails the task regardless of the remaining `max_retries`.

I chose two flat attributes to express the full range of options without introducing new type machinery. WDL cannot cleanly express `preemptible: { max_attempts: 3, only: true }`: a bare `{ ... }` is a map literal whose values must be homogeneous, while `object { ... }` uses the deprecated `Object` type and a named struct would require reserving a new type.

The spec also defines “preemptible instance” and explicitly states that absence of `preemptible` is equivalent to `0`. I would appreciate another review of this design.

SPEC.md
```diff
@@ -5802,6 +5804,38 @@ task max_retries_test {
...
+    max_retries: 2
+  }
+
+  hints {
+    preemptible: 5
+  }
+}
+```
+
+Up to five attempts may run on a preemptible instance. If all five end in preemption, the task fails, and the engine must not run it on a non-preemptible instance. Independently, up to two attempts that fail for reasons other than preemption may be retried on preemptible instances.
```
vsmalladi: @claymcleod  I would clarify this last sentence to be:

One retry remains, allowing the engine an additional attempt for reasons other than preemption on non-preemptible instance.

vsmalladi (changes requested):
@claymcleod My major concern is the split between `preemptible_only` and `preemptible` in `requirments` and `hints`

If the specification states, "When preemptible_only is true, the engine must honor this hint and must not run the task on a non-preemptible instance," then classifying preemptible_only as a hint breaks the fundamental definition of what a "hint" is supposed to be.

A `hint`  can safely ignore by an engine. But the way its written now. if an engine is forced to fail the task because a setting is true, that setting is no longer a hint. It becomes a hard requirement.

I would suggest move `preemtible` to the `requirement` section to be cleanest.

claymcleod:
`preemptible_only` is already a requirement. “This hint” refers to `preemptible`, but I agree that requiring the engine to honor it contradicts the general semantics of hints.

The intended distinction is that `preemptible_only: true` independently prohibits non-preemptible execution, while `preemptible` remains an advisory attempt count that the engine may safely ignore. I will remove the language and validation rule that make `preemptible_only` depend on a positive `preemptible` value. Moving `preemptible` into requirements would remove the intentionally optional behavior.

vsmalladi (approved):

claymcleod:
This vote has passed with 3 for (Venkat Malladi, John Didion [via thumbs up on main PR], Clay McLeod) and 0 against.

Issue #553, opened 2023-04-04
`split` function to split a string on a delimiter


jdidion:
`Array[String] split(String, String)`

Splits the first argument (`input`) each time it encounters the second argument (`pattern`), where `pattern` is a POSIX ERE. This is the inverse of the `sep` function for a fixed pattern.

One impetus for this is to support deserialization of a delimited string to a string array.

claymcleod:
Thgis was implemented in #729.

PR #729, opened 2025-10-15
feat: introduces the `split` standard library function

claymcleod:
This PR adds the `split` standard library function as proposed in [#553](https://github.com/openwdl/wdl/issues/553).

      changes (when appropriate).

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -7798,7 +7798,7 @@ Regular expressions are written using regular WDL strings, so backslash characte
 Array[String] split(String, String)
 ```
 
-Given the two `String` parameters `input` and `delimeter`, this function splits the input string on the provided delimiter and stores the results in a `Array[String]`. `delimiter` is a [regular expression](https://en.wikipedia.org/wiki/Regular_expression) and is evaluated as a [POSIX Extended Regular Expression (ERE)](https://en.wikipedia.org/wiki/Regular_expression#POSIX_basic_and_extended).
+Given the two `String` parameters `input` and `delimiter`, this function splits the input string on the provided delimiter and stores the results in a `Array[String]`. `delimiter` is a [regular expression](https://en.wikipedia.org/wiki/Regular_expression) and is evaluated as a [POSIX Extended Regular Expression (ERE)](https://en.wikipedia.org/wiki/Regular_expression#POSIX_basic_and_extended).
 Regular expressions are written using regular WDL strings, so backslash characters need to be double-escaped (e.g., `"\\t"`).
 
 **Parameters**:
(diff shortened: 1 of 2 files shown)


claymcleod:
I've now prototyped this in Sprocket here: https://github.com/stjude-rust-labs/sprocket/pull/424.

SPEC.md
```diff
@@ -7791,84 +7792,36 @@ Regular expressions are written using regular WDL strings, so backslash characte
...
-```
-</p>
-</details>
+**Parameters**:
+
+1. `String`: the input string.
+2. `String`: the delimiter to split on as a regular expression.
 
-Any arguments are allowed so long as they can be coerced to `String`s. For example, this can be useful to swap the extension of a filename:
+**Returns**: the parts of the input string split by the delimiter. If the input delimiter does not match anything in the input string, an array containing a single entry of the input string is returned.
```
patmagee, line 7919: Will the method return an empty string if the split occurs at the start / end of the string?

claymcleod: Yes, `split(" hello ", " ")` will return `["", "hello", ""]`. I tested the default implementation in Python, Javascript, and Rust—all of them have this functionality. So I don't necessarily think it's worth calling out specifically, but open to thoughts on that.

patmagee (approved):
I feel like this will be a welcomed inclusion to the specification. Simple string manipulation seems like a bad reason to spin up a completely new task

a-frantz:
> I feel like this will be a welcomed inclusion to the specification. Simple string manipulation seems like a bad reason to spin up a completely new task

agree! I feel awful every time I have to call our WDL task which implements this - https://stjudecloud.github.io/workflows/tools/util/split_string-task.html

SPEC.md
```diff
@@ -7791,84 +7792,36 @@ Regular expressions are written using regular WDL strings, so backslash characte
...
-
-  output {
-    String chocolove = sub(chocolike, "like", "love") # I love chocolate when\nit's late
-    String chocoearly = sub(chocolike, "late", "early") # I like chocoearly when\nit's early
-    String chocolate = sub(chocolike, "late$", "early") # I like chocolate when\nit's early
-    String chocoearlylate = sub(chocolike, "[^ ]late", "early") # I like chocearly when\nit's late
-    String choco4 = sub(chocolike, " [:alpha:]{4} ", " 4444 ") # I 4444 chocolate 4444\nit's late
-    String no_newline = sub(chocolike, "\\n", " ") # "I like chocolate when it's late"
-  }
-}
```
tefirman, line 7820: Looks like the `sub` example is getting deleted here, not sure if that was intentional, but feels beneficial to leave in.

tefirman (approved):
Great addition, just a small note regarding inclusion of the `sub` example.

SPEC.md
vsmalladi: Ya i think this should be left in.

vsmalladi (changes requested):
I think looks good. But still need the sub example

claymcleod:
Thanks for the catch!

Issue #727, opened 2025-09-25
`wdl-1.2.1` `string_to_file.wdl` coercion example contains a `String`-to-`File` comparison


adamnovak:
The `wdl-1.2.1` branch for #702 has a comparison here between a `String` and a `File`.
https://github.com/openwdl/wdl/blame/7840869df42c14fbed4db5472127329e9c827da3/SPEC.md#L1607

In the `wdl-1.1` branch this was a comparison between a `File` and a `File`, and it was changed in 70947bf0 by @jdidion.

MiniWDL rejects this and says that `String` and `File` can't be compared with `==`. Reading the coercion rules, I'm inclined to agree with it: they say you can coerce from `String` to `File`, and that comparisons should try and coerce to `String`, but you don't seem to be able to coerce from `File` to `String`, and comparison isn't supposed to try and coerce to `File`.

Should the example be changed back to a `File`-to-`File` comparison? Or should the coercion rules be adjusted to make it clear that a `String`-to-`File` comparison is allowed (and, in that case, what is being coerced to what)?

claymcleod:
This was fixed in [#702](https://github.com/openwdl/wdl/pull/702).

PR #706, opened 2025-02-20
Fix tests 1.1

vsmalladi:
Fix changes to test inputs for wdl 1.1 spec

Close #707 
Close #708 
Close #701 

Before submitting this PR, please make sure:

- [x] You have added a few sentences describing the PR here.
- [x] You have added yourself or the appropriate individual as the assignee.
- [x] You have updated the `README.md` or other documentation to account for these 
      changes (when appropriate).
- [x] You have updated the `CHANGELOG.md` describing the change and linking back to your pull request.
- [x] You have read and agree to the [`CONTRIBUTING.md`](https://github.com/openwdl/wdl/blob/wdl-1.2/CONTRIBUTING.md) document.
- [x] You have added or updated relevant example WDL tests to the specification.
  - See the [guide](https://github.com/openwdl/wdl-tests/blob/main/docs/MarkdownTests.md) for more details.

PR #731, opened 2025-10-16
Fix comparison and output names

adamnovak:
This fixes a String-to-File comparison and some incorrect expected output JSON keys.

This is against the `fix-tests` branch that is going to be merged in #706.

I need at least the output name fix to fix how Toil's CI system parses the WDL spec tests, and the comparison change is needed for that workflow to typecheck under MiniWDL. The existing approach of comparing a `File` and a `String` doesn't seem to be allowed by the spec text (see https://github.com/openwdl/wdl/issues/727).

- (N/A: can't do) You have added yourself or the appropriate individual as the assignee.
- [X] You have updated the `README.md` or other documentation to account for these 
      changes (when appropriate).
- (N/A: file missing) You have read and agree to the [`CONTRIBUTING.md`](https://github.com/openwdl/wdl/blob/wdl-1.2/CONTRIBUTING.md) document.
- [X] You have added or updated relevant example WDL tests to the specification.


adamnovak:
I think the real changelog entry will be for when #706 merges, so I'm not including one here.

vsmalladi:
@adamnovak  let me know when you are ready for me to review and I can merge.

claymcleod:
Thanks for pointing out the `CONTRIBUTING.md` link, I've fixed it now on the `wdl-1.2` branch.

SPEC.md
```diff
@@ -7762,8 +7762,8 @@ Example output:
 
 ```json
 {
-  "test_find.match1": "ello",
-  "test_matches.is_read1": null
+  "find_string.match1": "ello",
```
vsmalladi: @adamnovak  Not sure why this is changed. The Output is from the workflow `test_find`

vsmalladi (changes requested):

SPEC.md
adamnovak: I'm not sure either; I think I worked off an old version and then merged and it changed in the workflow?

Anyway, I fixed it now.

vsmalladi (approved):
Looks great. Will merge into my branch.

PR #733, opened 2025-10-20
feat: adds `max_retries` to the `task` type

### LINKED ISSUE #732: The `task` type should include `max_retries`
I noticed recently during implementation of https://github.com/stjude-rust-labs/sprocket/pull/432 that the `task` type was missing the `max_retries` member (see this comment for more information on where exactly I ran into it: https://github.com/stjude-rust-labs/sprocket/pull/432#discussion_r2441325673). You can see it missing from [this section](https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#runtime-access-to-requirements-hints-and-metadata) of the spec.

There are basically two arguments for including it.

1. **Completeness.** The `task` type includes a concept for all of the requirements fields _except_ `max_retries`. Given that there is no obvious reason to exclude this field, this makes me think that it was simply omitted erroneously.
2. **Real use-cases exist.** Though we tend to not try to contrive use-cases on the WDL project, this is a member that I think would have a material use-case when combined with #696. In particular, I can envision someone wanting to crank up the resource requirements when `task.attempt == task.max_retries`.

If anyone is aware of a reason why this should _not_ be included other than an oversight, please let me know.

claymcleod:
The max retries (`max_retries`) appears to have been omitted from the `task` type on accident. This PR amends that by adding it to the type. It also makes explicit a few of the types that otherwise had to be read fully to understand the type.

Closes #732.

      changes (when appropriate).


vsmalladi:
@claymcleod  is this total retries or retries not `preemtible`

claymcleod:
Correct.

patmagee:
@claymcleod which part of @vsmalladi's question is correct :)?

claymcleod:
> @claymcleod which part of @vsmalladi's question is correct :)?

Sorry, it's the former. It represents the same thing that `max_retries` represents as part of `requirements`.

> The `max_retries` attribute specifies the maximum number of times a task should be retried in the event of failure. The execution engine must retry the task at least once and up to (but not exceeding) the specified number of attempts.

vsmalladi:
This is good. As long as its clear to implementers I am good.

PR #734, opened 2025-10-20
feat: adds pre-evaluation task availability alongside task.previous in post-evaluation contexts

### LINKED ISSUE #696: Context-aware retries using previous resource requirements.
It would be really useful to reference parameters from previous runs upon retry. For example, if you'd like to keep retrying a job up until a particular limit until it succeeds, it would could theoretically be expressed like this.

```wdl
task foo {
  command <<<
    echo "do some work"
  >>>

  requirements {
    memory: if task.iteration > 0 {
      task.previous.memory * 2
    } else {
      # 2 GB
      2000000000
    }
  }
}
```

In other words, if this is the first iteration of the task, start with 2 GB. Otherwise, take the previous task requirement and double it. I'm curious if this is something that might be generally useful when building robust task implementations.

## Open Questions

- Should only the last task requirement be kept as `task.previous` (or similar), or should there be, say, an array of task requirements from previous attempts?

claymcleod:
This PR implements (a) the `task` variable being available in pre-evaluation contexts (with limited fields) and (b) the `previous` field on the `task` type.

In pre-evaluation contexts, the following fields are available on the newly introduced "pre-evaluation context" `task`:

1. `task.name`, `String` - The name of the task
2. `task.id`, `String` - The unique identifier for the task execution
3. `task.attempt`, `Int` - The current attempt number (0-indexed)
4. `task.meta`, `Object` - the task's metadata
5. `task.parameter_meta`, `Object` - the task's parameter metadata
6. `task.ext`, `Object` - the task's extension metadata
5. `task.previous`, `task.previous` hidden type - Contains the previous attempt's requirements (see below).

In post-evaluation contexts, the fields remain the same, except WDL 1.3 now adds the `task.previous` member.

The `task.previous` field contains a new `task.previous` hidden type that contains the previously computed requirements for the last task attempt. It has has these optional members:

- `task.previous.memory`, `Int?` - Previous value of `task.memory`
- `task.previous.cpu`, `Float?` - Previous value of `task.cpu`
- `task.previous.container`, `String?` - Previous value of `task.container`
- `task.previous.gpu`, `Array[String]?` - Previous value of `task.gpu`
- `task.previous.fpga`, `Array[String]?` - Previous value of `task.fpga`
- `task.previous.disks`, `Map[String, Int]?` - Previous value of `task.disks`
- `task.previous.max_retries`, `Int?` - Previous value of `task.max_retries`

This allows for dynamic resource allocation based on the current task attempt (e.g., doubling memory each time the task runs).

Closes #696.

      changes (when appropriate).


claymcleod:
This is now implemented in https://github.com/stjude-rust-labs/sprocket/pull/432.

patmagee (approved):
Nice clarification and addition

vsmalladi:
@claymcleod  is this just looking at the previous attempt requested resources or is the expectation that this will provide the actual resources that were used in the task for the next task?

claymcleod:
I actually just changed this in my latest commit because I realized I wanted it to provide the latter (the actual resources used). In this way, all of the `task.previous.*` fields match the value of `task.*` in the last task attempt.

vsmalladi:
@claymcleod  thanks for the clarification. Now I have a follow up question. How do these play with the hints like `max_cpu`, `max_memory`? Is the expectation that if `max_cpu` is 4 then if we are on the 4th retry that the `cpu` would request 5 but the max reality is that we could only get a machine with 4. Is that correct?

```
requirements {
    # Only name, id, attempt, previous, meta, parameter_meta, and ext are available in pre-evaluation
    cpu: task.attempt + 1
    memory: "~{256 * (2 ** task.attempt)} MB"
    container: "ubuntu:latest"
    max_retries: 4
  }
```

claymcleod:
> @claymcleod thanks for the clarification. Now I have a follow up question. How do these play with the hints like `max_cpu`, `max_memory`? Is the expectation that if `max_cpu` is 4 then if we are on the 4th retry that the `cpu` would request 5 but the max reality is that we could only get a machine with 4. Is that correct?
> 
> ```
> requirements {
>     # Only name, id, attempt, previous, meta, parameter_meta, and ext are available in pre-evaluation
>     cpu: task.attempt + 1
>     memory: "~{256 * (2 ** task.attempt)} MB"
>     container: "ubuntu:latest"
>     max_retries: 4
>   }
> ```

Good question. First, I will point out that this is a bit orthogonal to this PR enabling dynamic values—the situation where I specify a `cpu` higher than the `max_cpu` is possible today without needing to use this feature. So I don't think this thread is the best long-term place for this discussion.

That being said, I'll take an initial swing at it.

For `max_cpu`, the spec states:

> the task expects to use no more than the specified number of CPUs

This doesn't read to me as a hard cap but, instead, some softer indication to the engine of what might be used. I did some digging, and there is some discussion on #595 between @markjschreiber and @jdidion talking about this field (you can see it in the resolved comments). In brief, Mark asks about this exact thing to which John responds

> In this case I think the hint is really just meant as a potential optimization. I.e. it's communicating to the engine that the task will use no more than maxCpu cpus, so it is free to limit the provisioned resources. But maybe others have a different view.

They eventually agree on a softer language with this communicated expectation for engines

> The engine is free to make any optimization it deems beneficial based on this information

So I think the answer to your question given the current state of the spec would be "the engine is free to limit the resources based on this hint, but there is no requirement to do so".

tefirman (approved):
Really like the potential of being able to iteratively increase resource requests after each retry. Especially with WGS alignment tasks, that would be super useful. Great addition, thanks @claymcleod !

vsmalladi (approved):
@claymcleod  This is great.

PR #735, opened 2025-10-21
revise: clarifies behavior around non-output relative paths

### LINKED ISSUE #676: Disallow relative path literals in declaration expressions except in output section
This is an abridged version of the same issue from [here](https://github.com/openwdl/wdl/issues/673#issuecomment-2248828116)

An computation that depends on nullable/optional files can vary depending on where in the task/workflow it is placed:
```wdl
version 1.1
task t {
  input {}
  command <<<>>>
  Array[File?] file_arr = ["example1.txt"]
  output {
    Int len = length(select_all(file_arr)
  }
}
```
```wdl
version 1.1
task t {
  input {}
  command <<<>>>
  output {
    Array[File?] file_arr = ["example1.txt"]
    Int len = length(select_all(file_arr))
  }
}
```
According to the spec:
https://github.com/openwdl/wdl/blob/caff59db192636d9f93f3f5659eb5939f51ff877/SPEC.md?plain=1#L3880
> All file *outputs* are required to exist, otherwise the *task* will fail... the value will be undefined if the file does not exist

Since only in the task output are the values allowed to be undefined, the first workflow will output `len = 1` while the second workflow will output `len = 0`

claymcleod:
This PR provides further clarification around issue #676 by clarifying how relative paths in `File` and `Directory` declarations are resolved depending on their context within a WDL document.

As stated in #676, the same code `Array[File?] file_arr = ["example.txt"]` where `example.txt` doesn't exist produces different results depending on location in WDL v1.1:

  - In private declarations, the length of this array is 1 because the 1.1 spec does not explicitly state that missing files are considered `None`.
  - In output sections, the length will evaluate to 0 because the spec _does_ explicitly state that missing files are coerced to `None`.

This confusion stemmed from unclear semantics about where relative paths are resolved and when file existence is checked.

John advocates in #676 that we should deprecate relative path literals in bound declarations outside the output section, forcing users to use `String` instead. However, the issue was partially resolved in a clarification to the WDL v1.2 spec, which now states:

> Within a WDL file, literal values for files may only be (relative or absolute) paths that are local to the execution environment. **If the specified path does not exist, it is an error unless the declaration is optional.**

As such, WDL v1.2 now makes it clear that `File?`s and `Directory?`s should _always_ evaluate to `None` if the file doesn't exist.

That being said, the current specification still falls short on explicitly defining what relative paths are relative _to_ in inputs and private declarations. This PR aims to clarify that behavior and also clarify a few places where supporting language exists throughout the specification.

My plan is to first get these changes merged into WDL v1.3 and then backport them as needed to WDL v1.2 and WDL v1.1 after we settle on the final desired behavior.

Closes #676.

      changes (when appropriate).

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -4654,7 +4654,7 @@ Test config:
 </p>
 </details>
 
-Relative paths are interpreted relative to the execution directory, whereas absolute paths are interpreted in a container-dependent way. Absolute paths that reference locations outside the task's execution directory (e.g., system files like `/etc/hosts`) may not be supported by all execution engines, particularly in containerized or cloud environments where access to the host filesystem is restricted.
+Relative paths are interpreted relative to the execution directory, whereas absolute paths are interpreted in a container-dependent way. Absolute paths that reference locations outside the task's execution directory may not be supported by all execution engines, particularly in containerized or cloud environments where access to the host filesystem is restricted.
 
 <details>
 <summary>
--- a/SPEC.md
+++ b/SPEC.md
@@ -891,10 +891,8 @@ Example: relative_paths_context.wdl
 version 1.3
 
 task relative_paths_context {
-  input {
-    # This relative path is resolved relative to the WDL document's parent directory.
-    File input_file = "hello.txt"
-  }
+  # This relative path is resolved relative to the WDL document's parent directory.
+  File input_file = "hello.txt"
 
   command <<<
     cat ~{input_file} > output.txt
--- a/SPEC.md
+++ b/SPEC.md
@@ -868,6 +868,69 @@ task literals_paths {
 }
 ```
 
+The interpretation of relative paths (paths that do not start with `/`) depends on the context in which they appear:
+
+* *Outside the `output` section (e.g., in `input` or private declarations)*, relative paths are interpreted relative to the parent directory of the WDL document itself, similar to how [import](#import-statements) paths are resolved. This allows a WDL document to reference data files that are co-located with it on the host filesystem.
+* *Inside the `output` section*, relative paths are interpreted relative to the task's execution directory. This is where task commands create their output files. See [Task Outputs](#task-outputs) for details.
+
+Absolute paths (paths starting with `/`) refer to specific locations on the host filesystem when used outside the `output` section. Within the `output` section, absolute paths may be interpreted in a container-dependent way—see [Task Outputs](#task-outputs) for details.
+
+<details>
+<summary>
+Example: relative_paths_context.wdl
+
+```wdl
+version 1.3
+
+task relative_paths_context {
+  input {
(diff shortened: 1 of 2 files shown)


SPEC.md
```diff
@@ -865,9 +865,80 @@ task literals_paths {
     # does not exist, then `f2` is set to `None`.
     File? f2 = "/foo/bar.txt"
   }
+
+  # If baz.txt does not exist, this is an error.
```
claymcleod, line 871: This is just making explicit the assertion above that bound declarations behave this way as well.

SPEC.md
```diff
@@ -4583,7 +4654,7 @@ Test config:
 </p>
 </details>
 
-Relative paths are interpreted relative to the execution directory, whereas absolute paths are interpreted in a container-dependent way.
+Relative paths are interpreted relative to the execution directory, whereas absolute paths are interpreted in a container-dependent way. Absolute paths that reference locations outside the task's execution directory may not be supported by all execution engines, particularly in containerized or cloud environments where access to the host filesystem is restricted.
```
peterhuene:
```suggestion
Relative paths are interpreted relative to the execution directory, whereas absolute paths are interpreted in a container-dependent way. Absolute paths that reference locations outside the task's execution directory may not be supported by all execution engines, particularly in environments where access to the container's filesystem is restricted.
```

SPEC.md
```diff
@@ -865,9 +865,80 @@ task literals_paths {
...
+  }
+
+  command <<<
+    cat ~{input_file} > output.txt
+  >>>
+
+  output {
+    # This relative path is resolved relative to the execution directory.
+    File result = "output.txt"
+    String content = read_string(result)
```
vsmalladi, line 906: @claymcleod can we add example of relative paths outside the output directory `/` and what the expected results of also missing files. I feel like this test doesn't really cover the update definition

vsmalladi (changes requested):

SPEC.md
claymcleod: This test is showing the correct behavior and is meant to be an integration tests that succeeds. It covers the updated definition by ensuring that `hello.txt` must resolve relative to the WDL document's parent directory and printing the correct results. Though the semantics of a `File` in input blocks and in private declarations in the same, I can change it to a private declaration to make it a bit more clear.

claymcleod: Also, line 869 above shows the error cases. This should cover the concerns above.

peterhuene (approved):

claymcleod:
This was backported to WDL 1.2 in https://github.com/openwdl/wdl/commit/c8fa1816ab798e169d90f24115469138cb7a8e2d. I did not end up backporting to WDL 1.1 because the structure of the files and directories documentation is significantly different. I determined it would be more effort than it's worth to try to backport it there.

PR #742, opened 2025-11-02
revise: moves `task.return_code` to be available in the output sectio…

### LINKED ISSUE #726: Clarification of type for `return_code` in output section
Currently the spec reads:

`return_code`: An `Int?` whose value is initially `None` and is set to the value of the `command`'s return code. The value is only guaranteed to be defined in the `output` section.

Does this mean that its expected to be `Int?` in the output section or if it can just be an `Int`.

claymcleod:
This PR updates the WDL 1.2 spec by making the `task.return_code` only be available in the `output` section as an `Int`. This seems more logical to me, and it's in accordance with what the existing example `test_runtime_info_task.wdl` seems to suggest the intention was. Technically, this is a breaking change—however, I think that the risk is low that this will affect anyone's code, and we can quickly implement the fix in Sprocket.

Closes #726.

- [ ] You have added yourself or the appropriate individual as the assignee.
      changes (when appropriate).


claymcleod:
cc: @peterhuene, what do you think about this?

peterhuene:
This makes sense to me. For 1.3 where we want to make `task` available to `hints` and `requirements`, I assume we'll keep it on `previous` but also as `Int` instead of `Int?`?

Let's get some more sign-off from other committee members on this though as it is technically a breaking change.

claymcleod:
> I assume we'll keep it on `previous` but also as `Int` instead of `Int?`?

That was my thinking as well.

PR #749, opened 2025-11-10
revise: clarifies the semantics of the replacement string in `sub()`

### LINKED ISSUE #705: `sub()` STDLIB function: formalize the syntax for using capturing groups in the `replace` string
The SPEC (https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#sub) is explicit that the `pattern` string must be POSIX ERE compliant, but doesn't say anything about what's permitted in the `replace` string. This leaves it up to the engine whether capturing groups from the RE can be used in the `replace` string, and there isn't any uniformity as to the syntax for that, leading to non-portable WDL.

e.g. `sprocket` uses the `regex` crate which supports `$<ref>` syntax for capturing groups in the `replace` string. I _think_ that `miniwdl` uses `\<ref>` syntax (or that's the syntax for the Python sub() func). No idea what cromwell does.

One standard should be enforced instead of leaving it up to the engine. I don't think the specific standard (`$ref` or `\ref`) matters, so long as it's portable across engines.

claymcleod:
This PR addresses #705 by following the POSIX ERE inspiration and deciding on the `\1`-`\9` route for capture group replacement present in [sed](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/sed.html).

Closes #705.

For OpenWDL team members:


claymcleod:
This needs to be backported to WDL 1.1

patmagee (approved):
LGTM

PR #752, opened 2026-01-01
Updates the 1.2.1 branch to use `spectool`

claymcleod:
This PR updates the compliance tests to use [`spectool`](https://github.com/openwdl/spectool). It also fixes quite a few of the compliance tests. Last, it defines what compliance with the specification actually means (i.e., it quantifies how many compliance tests you have to pass to be considered "compliant"). 

For OpenWDL team members:


SPEC.md
```diff
@@ -4817,13 +4830,16 @@ version 1.2
...
-  mkdir -p my/path/to
-  printf "something" > my/path/to/something.txt
+    mkdir -p my/path/to
+    printf "something" > my/path/to/something.txt
   >>>
 
   output {
     String something = read_string("my/path/to/something.txt")
-    File bashrc = "/root/.bashrc"
+    # The following may or may not work depending on what the execution engine
```
claymcleod, line 4839: Importantly, the text says right above that an execution engine may not support absolute paths outside the container's working directory. I think then it is unwise to have this as part of the conformance tests. For example, Sprocket fails this task emitting an error, but I don't think that indicates lack of conformance to the spec.

SPEC.md
```diff
@@ -6547,19 +6555,12 @@ Example input:
...
-}
-```
-
 Test config:
 
 ```json
 {
-  "dependencies": "allow_nested_inputs"
+  "capabilities": ["allow_nested_inputs"],
+  "fail": true
```
claymcleod, line 6555: As the paragraph above states, settings `allow_nested_inputs` to `false` sets it to `false` for all subworkflows, so this actually should fail.

SPEC.md
```diff
@@ -10996,7 +11009,7 @@ For example, if the first argument is a `Map[String, Map[String, Int]]` and the
 
   struct Person {
     String name
-    Map[String, String]? details
+    Map[String, String] details
```
claymcleod, line 10994: The function signature above requires that this not be optional. That being said, we can just omit the details for `Agent X` and get the desired result here.

claymcleod:
Also pertinent to the discussion: I've set a threshold of 95% pass rate or higher to be considered "compliant" with WDL 1.2. Curious what people's thoughts are on this threshold—I think a 100% compliance rating is likely going to be too stringent.

SPEC.md
```diff
@@ -8064,30 +8056,20 @@ task join_paths {
...
   }
 
   # these are all equivalent to '/usr/bin/echo'
   String bin1 = join_paths(abs_dir, [rel_dir_str, rel_file])
   String bin2 = join_paths(abs_str, [rel_dir_str, rel_file])
   String bin3 = join_paths([abs_str, rel_dir_str, rel_file])
 
-  # the default behavior is that this resolves to
-  # '<working dir>/mydir/mydata.txt'
-  String data = join_paths(rel_dir, rel_str)
```
claymcleod, line 8069: Given the semantics of how paths are joined, this joins the host path for `rel_dir` with the file, which is not accessible inside the contain. As such, this is not a solid demonstration of the `join_paths` function.

PR #765, opened 2026-04-23
feat: add WDL module specification and symbolic imports

### LINKED ISSUE #226: Versioned import statements
NB this is *not* about WDL 1.0 vs draft-2 etc. It's about versioning of WDL files within a repo.

As I understand things, TRS repos (amongst others) will allow people to have multiple versions available for the same workflow and task and responsible WDL authors will presumably want to be able to choose which version to import (to avoid things changing underneath them when a new update is added, for example)

To support this - and to avoid having to work out a complex or repo-specific URL structure, I'd like to suggest we allow embedding the version to import into the `import` statement. 

Maybe something like:
```wdl
version ...
import "tools/published_tool" @ "v1.0"
import "tools/new_tool_version" @ "v11.3"
import "structs/alpha_struct_defs" @ "v0.5"
```

### LINKED ISSUE #758: Implement package specification and ecosystem
## Summary

Implement formal package specification and ecosystem support for WDL, enabling symbolic imports, manifest files, and versioned dependencies.

See #700 for the full RFC and details.

## Key Goals

- Symbolic imports (e.g., `import fastp from biowdl`)
- Manifest file (`wdl-package.json`) defining package metadata and dependencies
- Support for local paths, URLs, and git repositories as dependency sources
- Semantic versioning requirements for packages

## Related

- RFC: #700
- Relaxed version imports: #698

claymcleod:
This introduces a module system to WDL in two parts: a new peer specification at `modules/SPEC.md` covering the ecosystem (manifest, resolution, lockfile, hashing, signing, credentials), and language-level additions in `SPEC.md` (the `from` keyword and two new import forms).

The module specification defines:

- `module.json`, the manifest format: core metadata fields, upstream `tools` provenance (with CURIE-based `ids` for identifier registries like DOI and bio.tools), and a `dependencies` object supporting `version`, `tag`, `branch`, `commit`, and local `path` selectors.
- Tag-based versioning following Go modules: a module's version is defined entirely by its Git tags (`v<semver>`, or `<path>/v<semver>` in multi-module repositories). The manifest carries no version field.
- `module-lock.json`, a lockfile that pins the fully resolved tree with commit SHAs and content checksums for Git sources; local path dependencies are read as-is.
- A resolution algorithm describing how symbolic paths (`<dep-name>[/<sub-path>]`) map to modules, including hyphen/underscore normalization, semver requirement syntax with partial versions, and cycle detection by source coordinates.
- A content hashing scheme: a deterministic SHA-256 algorithm that serves both lockfile integrity and signature verification.
- Module signing via optional Ed25519 signatures in `module.sig`, with trust-on-first-use recorded in the lockfile, covering key changes and signature-removal downgrades.
- Credential management that delegates to Git credential helpers so private repositories work without a new credential store.

The language changes in `SPEC.md`:

- Reserve `from` as a keyword in WDL 1.4 documents. Earlier `version` declarations continue to parse `from` as an ordinary identifier.
- Add two new import forms alongside the existing one, for three forms total:
  1. `import <source> [as <alias>] (alias <Old> as <New>)*`, the existing namespaced form.
  2. `import * from <source>`, which brings every task, workflow, and user-defined type from `<source>` into the importing document's scope, with no namespace.
  3. `import { <member> [as <Name>] (, <member> [as <Name>])* [,] } from <source>`, which brings only the listed items, with an optional per-member `as <Name>` rename, also with no namespace.
- Define scope propagation and collision rules: names merged by forms 2 and 3 become members of the document's namespace and are visible to downstream importers; form 1 namespaces do not propagate; duplicate names are errors unless they denote the same declaration.
- Unify the two source styles: a `<source>` is either a quoted URI (resolved per [Import URIs](https://github.com/openwdl/wdl/blob/feat/modules/SPEC.md#import-uris)) or an unquoted symbolic module path (resolved through the consuming module's `module.json`). Once resolved, the two styles produce identical scoping in every form.

JSON Schemas for both `module.json` and `module-lock.json` ship under `modules/schemas/` and validate the in-spec examples.

Full context: RFC discussion #700.

For OpenWDL team members:

Closes #226.
Closes #758.


modules/SPEC.md
```diff
@@ -0,0 +1,552 @@
...
+  index.wdl
+  csvcut.wdl
+```
+
+Where `index.wdl` contains:
+
+```wdl
+version 1.4
+
+import "csvcut.wdl"
```
claymcleod: Here, make it clear that it's according to the rules of importing with quotes. Also add an example of using the new import syntax and stick with that for the rest of the guide (as it's more canonical for modules to use that syntax).

modules/SPEC.md
```diff
@@ -0,0 +1,552 @@
...
+```
+
+The contents of `csvcut.wdl`—its tasks, workflows, and user-defined types—become available under the `csvcut` namespace to consumers.
```
claymcleod:
```suggestion
The contents of `csvcut.wdl`—its tasks, workflows, and user-defined types—become available under the `csvcut` namespace to consumers using the associated rules outlined in [`SPEC.md`](../SPEC.md).
```

modules/SPEC.md
```diff
@@ -0,0 +1,552 @@
...
+
+These steps describe the logical behavior that compliant engines must produce. Implementation mechanics—caching strategies, scan ordering, eager vs. lazy fetching—are left to the engine.
+
+### Version Discovery
+
+How versions are discovered depends on the source type.
+
+For **Git-based dependencies**, the resolver lists the repository's Git tags and parses each as a semver version, stripping a leading `v` if present (e.g., tag `v1.2.0` → version `1.2.0`). Tags that do not parse as valid semver are ignored. The resulting set is matched against the `version` requirement. Publishing a new version therefore consists of tagging a commit; there is no separate publication step, no upload, no registry submission.
+
+**Tag-to-manifest consistency.** The `version` field in `module.json` at the tagged commit must match the version encoded in the tag (after stripping `v` and any path prefix). A tag `v1.2.0` pointing to a commit whose `module.json` declares `"version": "1.3.0"` is a validation error. Engines must reject such mismatches during resolution and surface a clear error message.
```
claymcleod: This should also indicate that the error message should suggest to file an issue with the upstream.

modules/SPEC.md
```diff
@@ -0,0 +1,552 @@
...
+
+### Version Resolution and Conflicts
+
+When multiple modules in the dependency tree require the same dependency with compatible version constraints (e.g., `^1.2.0` and `^1.5.0`), the resolver should attempt to find a single version satisfying all constraints. This avoids unnecessary duplication.
+
+When the constraints are incompatible (e.g., `^1.0.0` and `^2.0.0`), both versions are fetched and used independently. No deduplication is attempted and no warning is emitted. WDL modules are text files, and the tasks they define execute in isolated containers with no shared runtime state; no conflict can arise from duplicate versions coexisting. The cost of duplication is a few kilobytes of WDL source per duplicate.
+
+## Lockfile (`module-lock.json`)
+
+Every module must have a `module-lock.json` file at its root. The lockfile pins the fully resolved dependency tree—every module in the tree, the exact commit each was resolved to, and a content checksum that detects tampering. The lockfile must be committed to version control; it is what makes builds reproducible regardless of upstream changes.
```
claymcleod: I don't think every module must have this. It's required if you want to pin versions, but not if you explicitly don't want to do that (say, for libraries of WDL files). We should also consider somewhere in the spec mentioning the responsibility of engines to uphold the `module-lock.json` invariants for dependencies.

modules/SPEC.md
```diff
@@ -0,0 +1,552 @@
...
+
+Cached module sources (the local directories where resolved modules are downloaded or cloned) must **not** be committed. These are ephemeral and can be reconstructed from the lockfile. Engines should store cached modules in a location outside the project directory (e.g., a user-level cache) or in a directory conventionally covered by `.gitignore`.
```
claymcleod: I think we need to make this a little more general. We do no prescribe what folder the modules need to go in (though, if no other name is preferred, an engine can use simply the `modules/` folder as the default). This, of course, does not mean we should remove the part about the global cache directory.

modules/SPEC.md
```diff
@@ -0,0 +1,552 @@
...
+csvcut/
+  module.json
+  module.sig
+  index.wdl
+  csvcut.wdl
+```
+
+Ed25519 was chosen because it is fast, produces small signatures (64 bytes) and small keys (32 bytes), and has mature implementations in every major language. Engines can verify signatures in-process without shelling out to external tools or depending on a system keychain.
+
+### Why Out-of-Band Rather than Git-Native Signing
```
claymcleod: Delete this section.

modules/SPEC.md
```diff
@@ -0,0 +1,552 @@
...
+Engines are encouraged to:
+
+- Surface a notice when resolving unsigned modules, nudging authors toward signing.
+- Provide a configuration option (e.g., `require_signed = true`) that rejects unsigned modules entirely. Whether this defaults to on or off is left to the engine implementor.
+
+## Credential Management
+
+Engines must rely on the user's configured Git credential helpers for authentication to private repositories. If the user's `git` command can clone a repository, module resolution from that repository must succeed without additional configuration. Engines must not introduce a new credential store or configuration format specifically for module resolution.
+
+This approach works with SSH keys, personal access tokens, and platform-specific credential managers (macOS Keychain, Windows Credential Manager, `git-credential-libsecret`, etc.) without modification. Most developers and CI systems already have Git credentials configured, and any such configuration applies to module resolution automatically.
```
claymcleod: Remove `git-credential-libsecret`.

modules/SPEC.md
```diff
@@ -0,0 +1,552 @@
...
+3. **Signature verification.** When a `module.sig` is present, verify the signature against the module's content hash, compare the signer's public key against the lockfile entry, and enforce the trust-on-first-use contract.
+4. **Structural validation.** Report clear errors for malformed `module.json` files, missing required fields, invalid SPDX license expressions, invalid semver versions, tag-to-manifest version mismatches, and unrecognized selectors.
+
+Engines may additionally provide higher-level commands for authoring convenience (e.g., scaffolding, validation, upgrade). This specification does not prescribe the command surface; engine authors are free to design their own CLI.
+
+Semantic annotation of task inputs and outputs (e.g., file format, ontology terms from EDAM) is a complementary concern and out of scope for this specification. Standardized `parameter_meta` conventions for declaring input/output formats would help tooling match compatible tasks and are worth pursuing as a separate effort.
```
claymcleod: This can be removed.

modules/SPEC.md
```diff
@@ -0,0 +1,552 @@
...
+
+## Known Open Questions
```
claymcleod: Remove this section.

modules/SPEC.md
```diff
@@ -0,0 +1,552 @@
...
+
+**Separate tool versioning.** The module version and the upstream tool version are tracked in distinct fields, but they are not independent. The module version contract is that unchanged versions must produce unchanged expected output; a tool update that alters output requires a module version bump. The `tools` array exists for provenance and license tracking, not as a substitute for proper semver on the module itself.
+
+**Display name, not resolution name.** The `name` field exists for human consumption. It is not used for dependency resolution; the importer names each dependency locally. This eliminates global namespace management, the squatting problem, and the need for a naming authority.
+
+**Lockfile as a specification requirement.** Making `module-lock.json` optional would undermine reproducibility. Requiring it ensures every module is reproducible by default.
```
claymcleod: This needs careful consideration: do we really want to make this required?

modules/SPEC.md
```diff
@@ -0,0 +1,552 @@
...
+
+**Trust on first use over a certificate authority.** TOFU has known downsides: the first resolution is unverified, key rotation requires manual acceptance, and there is no revocation mechanism. A PKI would address these but would require infrastructure and governance that a small open-source community cannot realistically sustain. TOFU protects against the most common attack—a repository compromised after adoption—and accepts the tradeoff that it cannot protect against pre-existing compromise.
+
+**Optional signing with encouraged adoption.** Requiring signatures would be more secure but would raise the authoring barrier. Making it optional, with an engine-level policy that enforces it for security-conscious environments, preserves adoption while enabling strong policy where needed.
+
+**Soft deprecation of remote URL imports.** A hard removal would strand existing workflows. Warnings give the ecosystem time to migrate while making the direction clear.
```
claymcleod, line 633: Remove this.

claymcleod:
@peterhuene and I have agreed to land a simpler model than the original eight-form proposal. Symbolic imports are interchangeable with quoted imports — the build system resolves a symbolic path to the equivalent of a quoted import. All three forms accept either source type with identical semantics.

The three forms:

1. `import <source> [as <alias>] (alias <Old> as <New>)*` — the existing import. UDTs (structs, enums) enter the importing document's scope; tasks and workflows are accessible through a pseudo-namespace but are not exported. `as <alias>` renames the pseudo-namespace; `alias <Old> as <New>` renames a UDT. `alias` cannot rename tasks or workflows.
2. `import * from <source>` — every task, workflow, and UDT enters scope.
3. `import { <member> [as <Name>], ... } from <source>` — only the listed items enter scope.

`<source>` is either a quoted URI or a symbolic module path. Forms 2 and 3 don't accept a trailing `as <alias>` or `alias` clause.

What changed from the prior eight-form model:

- No exported namespaces. The pseudo-namespace in form 1 still exists for accessing tasks and workflows, but it isn't part of the importing document's exported scope.
- `from` works with quoted URIs too. `import { foo } from "bar.wdl"` parses, since symbolic and quoted sources have identical semantics.
- Two imports both bringing `baz` into scope must be renamed explicitly:
  ```wdl
  import "foo.wdl" alias baz as foo_baz
  import { baz as bar_baz } from "bar.wdl"
  ```

This model fits the existing WDL mental model (no new namespace concept to teach), is easier to implement, and parses with no lookahead beyond `peek`. We'll try implementing some packages in this mode and see whether it feels constraining; if it does, we'll come back and formalize exported namespaces.

claymcleod:
After _much_ discussion, @peterhuene and I have agreed that this is the approach we're going to take.

- We're going to implement a simpler version of modules that is closer to what WDL is today (see the ruleset below) in Sprocket and see how it goes. This should be easier for WDL users to understand (follows the existing rules without introducing concepts like namespaces) but may introduce strain on more complicated packages.
- We'll try implementing some of our packages in this mode and see if it feels constraining. We'll probably also discuss with some other future package maintainers to see what their thought is. If necessary, we'll extend this PR to include formalizing the concept of namespaces (which exist today in WDL but only within a document—i.e., namespaces are not exported as part of the document's scope).

If we decide later on that namespaces are needed, then the directives to encapsulate members within a namespace (`as`/`within` on forms 2 and 3) will be introduced, and namespaces will be more fully built out in the proposal.

-----

The rules for the simpler version we're going to try first are below.

1. `import "foo.wdl"` follows the behavior that exists today.
  a. User-defined types (UDTs) like structs and enums are brought into the importing document's scope.
  b. Tasks and workflows are accessible in the document via a pseudo-namespace but that namespace is not actually brought in to the document's scope/exported.
2. `import * from "foo.wdl"` imports all tasks, workflows, and UDTs into the document's scope (no namespace) and makes them exportable.
3. `import { bar, baz } from "foo.wdl"` only imports the items `bar` and `baz` into the document's scope and also makes them exportable.

Other notables:

- You can use `as` in the first form to rename the pseudo-namespace that items are available under (exists today).
- You can use `alias` in the first form to rename members that get imported into the top-level document scope (the UDTs). Alias cannot be used to rename tasks or workflows (exists today).
- You can use `import { bar as baz } from "foo.wdl"` to rename individual items that are pulled in using the new syntax. The is semantically equivalent to the as in the first form, as it renames the thing that immediately precedes the keyword (even though the subject is a pseudo-namespace in form 1 and an item in form 3).
- This means that, if there is a conflict between two modules `foo` and `bar` that both import an item `baz`, a package maintainer's only choice is to alias imports in form 1/use `as` in form 3 to rename the items, e.g.
  - `import "foo.wdl" alias baz as foo_baz`, or
  - `import { baz as bar_baz } from "bar.wdl"
- Symbolic imports are largely going to be interchangable with the existing quoted imports, with the symbolic part simply resolving to the equivalent of a quoted import through the build system. So all forms accept both quoted and symbolic imports.

SPEC.md
```diff
@@ -3929,9 +3927,35 @@ workflow wf {
 }
 ```
 
+### ✨ Import Forms
+
+An import statement takes one of three forms. Every form accepts the same two source styles, and the source style affects only how the build system locates the imported document.
+
+1. `import <source> [as <alias>] (alias <Old> as <New>)*`. User-defined types (structs and enums) from `<source>` are copied into the importing document's scope. Tasks and workflows from `<source>` are accessible only through the import's namespace, which defaults to the filename minus the `.wdl` extension for a quoted URI or to the last component of the path for a symbolic module path. `as <alias>` overrides the default namespace; `alias <Old> as <New>` renames a struct or enum as it is copied. `alias` cannot rename tasks or workflows. See [Fully Qualified Names & Namespaced Identifiers](#fully-qualified-names--namespaced-identifiers) for how the namespace is used.
+2. `import * from <source>`. Every task, workflow, and user-defined type from `<source>` enters the importing document's scope. No namespace is introduced.
```
DavyCats, line 3935: Personally, I'm not a huge fan of these kinds of imports. They obfuscate where the imported task/workflow is actually defined. I'm wondering how others feel about this.

SPEC.md
```diff
@@ -3929,9 +3927,35 @@ workflow wf {
...
+3. `import { <member> [as <Name>], ... } from <source>`. Only the listed items enter the importing document's scope. A per-member `as <Name>` renames the selected item locally. A trailing comma after the last member is permitted. No namespace is introduced.
```
DavyCats: For consistency with how form 1 is explained, I guess `...` should be `*`?

```suggestion
3. `import { (<member> [as <Name>],)* } from <source>`. Only the listed items enter the importing document's scope. A per-member `as <Name>` renames the selected item locally. A trailing comma after the last member is permitted. No namespace is introduced.
```

Or maybe `+` if at least one needs to be present.

Or is `...` supposed to have a separate functionality, like "also import everything else but don't rename it".

SPEC.md
```diff
@@ -3929,9 +3927,35 @@ workflow wf {
...
+```wdl
+version 1.4
+
+import "csvkit.wdl"                                              # tasks via `csvkit` namespace; structs/enums in scope
+import "csvkit.wdl" as csv                                       # tasks via `csv` namespace; structs/enums in scope
+import * from "csvkit.wdl"                                       # tasks, workflows, structs, enums all in scope
+import { CsvSort } from "csvkit.wdl"                             # only `CsvSort` in scope
+import { CsvSort as MySort, CsvSortStable } from "csvkit.wdl"    # `MySort` and `CsvSortStable` in scope
+import openwdl/csvkit                                            # tasks via `csvkit` namespace; structs/enums in scope
+import { CsvSort } from openwdl/csvkit                           # only `CsvSort` in scope
```
DavyCats, line 3960: This is missing as example of form 1's alias old as new syntax.

DavyCats (commented):
Nice! Thanks for writing this up!

SPEC.md
hexylena: I used to feel this way when I was resolving them manually / by eye. But LSPs have mostly solved this issue for me, now I just click 'jump to definition' (also in e.g. python where they're also argued against for the same reason.)

claymcleod: Yeah and it's a little tricky too because of the design decisions that were made prior in WDL (i.e., https://github.com/openwdl/wdl/pull/160). It's actually a split situation today where `import "foo.wdl"` means:

- Import all of the user-defined types (structs and enums) into the current document's scope, effectively losing the tether to where they came from, and
- Make all of the tasks and workflows available at `foo.<task>`/`foo.<workflow>` but _don't_ export those as part of the scope.

This asymmetry poses real challenges to making any sort of module system usable under the current syntax.

Additionally, WDL, as it stands today, has no real concept of a formal namespace (there is something referred to as a namespace in the spec, but it's not allowed to be passed between document's using their scope).

Short of introducing namespaces more formally (which has a whole other set of language implications), we had to find a new syntax that made sense to actually pull the tasks and workflows into the document's scope alongside the UDTs while still feeling WDL-like. This was the best we could come up with, but I'm open to other ways to express this.

SPEC.md
claymcleod: Yeah sounds good, I'll clarify this.

modules/schemas/module-lock.schema.json
```diff
@@ -0,0 +1,62 @@
...
+    },
+    "source": {
+      "oneOf": [
+        {
+          "type": "object",
+          "additionalProperties": false,
+          "required": ["git", "commit"],
+          "properties": {
+            "git": { "type": "string", "format": "uri" },
+            "commit": { "type": "string", "pattern": "^[0-9a-f]{40}$" },
```
patmagee: Should this be `sha` to make the it explicit what the pattern is enforcing?

SPEC.md
patmagee: Agree that this can obfuscate what actually gets into the document, but It is a common pattern in most languages so I think a lot of people will expect this to exist.

modules/SPEC.md
```diff
@@ -0,0 +1,587 @@
...
+- The deterministic content-hashing algorithm used by both the lockfile and module signatures.
+- The Ed25519 signing scheme for module provenance.
+- Credential management for private repositories.
+- The minimum functionality compliant engines must provide.
+
+The language-level grammar for symbolic imports, and the scoping rules that govern the names brought into scope by those imports, are defined in [`SPEC.md`](../SPEC.md).
+
+## Module Directory Layout
+
+A module is a directory containing a `module.json` manifest at its root and one or more `.wdl` files. There is no additional organizational construct—no workspace type, no grouping file, no hierarchy requirement.
```
patmagee: For the time being this format will work, but I am sure at some point the question of module distribution and packaging is going to come up. This will be needed to solve for issues like offline dependency resolution, caching and maybe the future central dependency managment layer

- Python has wheels (`.whl`)
- Java has jars (`.jar`)
- Rust has crates (`.crate`)

these are all either zipped or gzipped folders that contain the package manifest and all other files.

I would propose a similar format and build a wdl module: `.wdlm` that is essentially just a zip file of the Module directory

modules/SPEC.md
```diff
@@ -0,0 +1,587 @@
...
+
+Each dependency declared by a consumer points to exactly one module folder: the directory containing that module's `module.json`. A repository may host a single module at its root or several modules in distinct subdirectories, but a single dependency entry never resolves to more than one module. To consume multiple modules from the same repository, declare each as its own dependency, distinguished by the `path` field (see [Path within a Repository](#path-within-a-repository)).
+
+A single-module repository:
+
+```
+csvcut-wdl/
+  module.json
```
patmagee, line 65: does a module ship with it's own module-lock.json? If there are conflicts how are these resolved?

modules/SPEC.md
```diff
@@ -0,0 +1,587 @@
...
+
+## Module File Tree Constraints
+
+The structural rules below are normative.
+
+1. Each file's relative path (computed per [Content Hashing](#content-hashing)) is treated as a sequence of `/`-separated components. The path must contain no `.` or `..` components, must not be absolute (no leading `/`, no Windows-style drive letter), and must not contain a null byte.
+2. Symbolic links inside a module are permitted only if their targets, after full resolution, fall inside the module root. A symlink whose target escapes the module root makes the module invalid.
+3. The names `module.json`, `module-lock.json`, and `module.sig` are reserved for files at the module root. A file with any of these names appearing at any other path within the module tree is a validation error.
+4. A module that, when materialized on disk, would produce any path failing these rules is invalid; engines must refuse to load it and surface a clear error identifying the offending entry.
```
patmagee: If a file in a module references a relative path outside the module root directly (not in the dependencies) would the module be considered invalid?

More generally what do import statements within a module look like?

modules/SPEC.md
```diff
@@ -0,0 +1,587 @@
...
+- **`license`** (string, required). An [SPDX license expression](https://spdx.github.io/spdx-spec/v2.3/SPDX-license-expressions/) (e.g., `"MIT"`, `"Apache-2.0"`, `"MIT OR Apache-2.0"`, `"MIT AND (Apache-2.0 WITH LLVM-exception)"`).
+- **`authors`** (array of strings, optional). Author descriptions. The convention for individual authors is `"First Last <first.last@example.com>"`, but this is not enforced.
+- **`description`** (string, optional). A brief description of what the module does.
+- **`repository`** (string, optional). The canonical Git URL for the module's source repository.
+- **`homepage`** (string, optional). A URL for the module's documentation or landing page, if distinct from the repository.
+- **`entrypoint`** (string, optional). Path to the module's entrypoint WDL file. The path must be relative and, after resolving any `.` or `..` components, must point to a location under the module root; absolute paths (leading `/` or a Windows-style drive letter) and paths that resolve outside the module root are not permitted. The path separator is `/`. Defaults to `index.wdl` if omitted.
```
patmagee, line 115: nit: wouldn't `main.wdl` be the canonical WDL way of defining the entrypoint?

modules/SPEC.md
```diff
@@ -0,0 +1,587 @@
...
+- **`readme`** (string, optional). Path to a markdown file. The path must be relative and, after resolving any `.` or `..` components, must point to a location under the module root; absolute paths (leading `/` or a Windows-style drive letter) and paths that resolve outside the module root are not permitted. The path separator is `/`. If omitted, engines and tooling look for `README.md` in the module directory. If set to `false`, no readme is associated with the module.
```
patmagee: should this be documentation? What if we want to point to an external URL instead of a README?

modules/SPEC.md
```diff
@@ -0,0 +1,587 @@
...
+- **`exclude`** (array of strings, optional). A list of gitignore-style glob patterns identifying files within the module that consumers may not reach via symbolic import. Each pattern must be a relative path that, after resolving any `.` or `..` components, points to a location under the module root. Absolute paths (leading `/` or a Windows-style drive letter) and patterns that resolve outside the module root are not permitted. The path separator is `/`. Plain directory names exclude the directory and everything beneath it; `*` matches any sequence of non-separator characters; `**` matches any sequence including separators. The patterns govern the public import surface only and have no effect on content hashing, signing, validation, or quoted imports within the module itself. See [Symbolic Module Paths](#symbolic-module-paths) for the resolution behavior. Defaults to the empty list.
```
patmagee: what is the use case for exclude and its semantics. Are these files referencable internally within the module and not externally or they should just be ignored in general?

If we were to package the module, would these be in the package or omitted entirely

modules/SPEC.md
```diff
@@ -0,0 +1,587 @@
...
+
+### Tools
+
+The `tools` field is an array of objects that tracks the upstream software wrapped by the module. Each entry records:
+
+- **`name`** (string, required). The tool name.
+- **`version`** (string, required). The tool version.
+- **`license`** (string, required). The tool's SPDX license identifier.
+- **`homepage`** (string, optional). URL for the tool's homepage or repository.
```
patmagee: maybe just url?

modules/SPEC.md
```diff
@@ -0,0 +1,587 @@
...
+- **`doi`** (string, optional). DOI for the tool's publication.
+- **`biotools`** (string, optional). [bio.tools](https://bio.tools/) registry identifier.
```
patmagee: instead of encoding this one, what about an additional `links` section that takes `Map(string, string)` that a user can add any number of additional links to

modules/SPEC.md
```diff
@@ -0,0 +1,587 @@
...
+}
+```
+
+#### Alternative Version Selectors
+
+For cases where semver requirements do not suffice (pre-release testing, pinning to a specific commit, tracking a development branch), the following alternative selectors are available:
+
+- **`tag`** — a specific Git tag name (e.g., `"v1.2.0-rc1"`). Not subject to semver resolution.
+- **`branch`** — a Git branch name. The resolved commit varies over time; the lockfile pins the exact commit at resolution time.
+- **`commit`** — a Git commit SHA. Any prefix that uniquely identifies a commit in the source repository is accepted; the resolver expands the prefix to the full 40-character SHA at lock time and records the full SHA in `module-lock.json`. The most precise and immutable selector.
```
patmagee: maybe change to sha?

modules/SPEC.md
```diff
@@ -0,0 +1,587 @@
...
+```
+
+#### Local Path Dependencies
+
+A dependency with a **`path`** key points to a local filesystem directory. No version selector is required; the module is used as-is from the local path.
+
+```json
+{
+  "dependencies": {
+    "local_utils": { "path": "../../shared/utils" }
```
patmagee, line 189: the implication of this is that it breaks out of the module root. To that extent it feels like the module spec is doing double duty:

1. defining a boundary for a reusable workflow
2. defining the dependencies for a workflow that may or may not be a module

If we allow relative paths pointing outside of the current module root we will satisfy 2, but not 1. If we restrict to not allow that then we will only  satisfy 1 and not 2

modules/SPEC.md
```diff
@@ -0,0 +1,587 @@
...
+  "homepage": "https://csvcut-wdl.someone.example.com",
+  "tools": [
+    {
+      "name": "csvcut",
+      "version": "2.0.1",
+      "license": "MIT",
+      "homepage": "https://csvkit.readthedocs.io/"
+    }
+  ],
+  "dependencies": {}
```
patmagee, line 237: can we expand the dependencies section to include an example?

modules/SPEC.md
```diff
@@ -0,0 +1,587 @@
...
+A consumer with a dependency named `csvcut` then chooses how to bring those names into scope using the import forms defined in [`SPEC.md`](../SPEC.md#-import-forms):
+
+```wdl
+version 1.4
+
+import csvcut              # tasks/workflows reachable as `csvcut.*`; UDTs in scope unqualified
+import * from csvcut       # tasks, workflows, and UDTs all in scope unqualified, no namespace
+```
+
+A module with multiple files curates its default surface by importing them from the entrypoint:
```
patmagee: can an entrypoint wdl consist of ONLY imports and no workflow/task/enum etc?

modules/SPEC.md
```diff
@@ -0,0 +1,587 @@
...
+
+Sub-path resolution is a direct file lookup. Intermediate directories along the sub-path do not need to contain `module.json` files, and any nested `module.json` files that happen to be present along the way are ignored.
+
+If the manifest's `exclude` field matches the path that resolution would otherwise read—either the entrypoint file for a root module import or `<sub-path>.wdl` for a sub-path import—the engine must treat the import as unresolvable and surface a missing-file error that names the path. The file's presence on disk is irrelevant; an excluded path is not part of the module's public import surface. Excluding the entrypoint therefore makes root module imports of that dependency fail while leaving non-excluded files reachable by sub-path. Quoted imports inside the module (e.g., the entrypoint's own `import "helper.wdl"`) are file-relative and are unaffected by `exclude`.
```
patmagee: this feels like a weakly enforceable security feature. if the file exists they can just address it directly instead of resolving via the module

patmagee (commented):
Looking amazing! Round 1 of comments done.. I still have a lot to go :D

SPEC.md
claymcleod: Since this thread started, the spec now defines the scoping rules for these forms precisely: names merged by forms 2 and 3 become members of the document's namespace, duplicate names are errors unless they denote the same declaration, and form 1 namespaces don't propagate (a5f2ecd). Doesn't change the taste question raised here, but the semantics are no longer implicit.

modules/SPEC.md
```diff
@@ -52,7 +52,7 @@ The language-level grammar for symbolic imports, and the scoping rules that gove
 
 ## Module Directory Layout
 
-A module is a directory containing a `module.json` manifest at its root and one or more `.wdl` files. There is no additional organizational construct—no workspace type, no grouping file, no hierarchy requirement.
+A module is a directory containing a `module.json` manifest at its root and one or more `.wdl` files. There is no additional organizational construct—no workspace type, no grouping file, no hierarchy requirement. This specification defines the logical module tree that engines consume; archive formats for distributing that tree, such as zipped module bundles, are intentionally left to future tooling as long as they materialize the same files and metadata.
```
claymcleod, line 57: I'm willing to revisit this before merging with the archive format, as suggested by @patmagee, but I'm just adding this here as it reflects the current state.

modules/SPEC.md
```diff
@@ -93,9 +93,10 @@ Both layouts are valid. The resolution logic is identical for both. Authors who
 The structural rules below are normative.
 
 1. Each file's relative path (computed per [Content Hashing](#content-hashing)) is treated as a sequence of `/`-separated components. The path must contain no `.` or `..` components, must not be absolute (no leading `/`, no Windows-style drive letter), and must not contain a null byte.
-2. Symbolic links inside a module are permitted only if their targets, after full resolution, fall inside the module root. A symlink whose target escapes the module root makes the module invalid.
+2. Symbolic links are not permitted anywhere in a module tree. A module containing a symbolic link is invalid.
```
claymcleod, line 98: This is a major change. I am just coming around the the perspective that the number of problems introduced by symlinks outweight any potential benefits.

modules/SPEC.md
claymcleod: Agree that this will come up. If it's alright with you, I'd like to defer just a bit on this. At this moment, it's unclear to me whether we will need to determine this before this PR merges or not. At a minimum, I've addressed this by making the spec clear that this is omitted for now.

modules/SPEC.md
claymcleod: Yes, a module may ship one (and should, if it wants reproducible resolution for its own development), but lockfiles apply only to the module they sit in: the consumer's resolver owns the full tree and upstream lockfiles are never consulted, so there's no conflict to resolve. This is now stated explicitly in the Lockfile section (a5f2ecd).

modules/SPEC.md
claymcleod: Yes, that's now invalid: file tree constraint 4 requires every quoted import inside a module to resolve within the module root, even if the target file exists on disk (a5f2ecd). As for what imports look like inside a module: quoted relative imports between the module's own files, and symbolic imports for anything declared in its `dependencies`. The Module Entrypoint section shows the former; transitive dependencies resolve through the latter. To be clear, a module can consist of only quoted imports if it has no dependencies and, thereby, can be entirely comprised of pre-WDL 1.4 documents.

modules/schemas/module-lock.schema.json
claymcleod: Sure. I've changed it. I kept it as `commit` in the `module.json`, as that can be a sha prefix.

modules/SPEC.md
claymcleod: The spec now explains the choice: `index.wdl` follows package index conventions (it names the file that represents the directory as a whole, where `main` typically names a program entry). A module that prefers `main.wdl` can set it via the `entrypoint` field without changing resolution semantics (a5f2ecd).

modules/SPEC.md
claymcleod: `homepage` covers the external case: its description now directs external documentation there, while `readme` is only for markdown files that ship inside the module tree (a5f2ecd). Keeping them separate lets tooling render an in-tree readme directly while still linking out to hosted docs.

modules/SPEC.md
claymcleod: The description now spells this out: `exclude` governs only the public symbolic-import surface. Excluded files remain referencable internally via quoted imports, stay part of the module tree, and have no effect on hashing, signing, validation, or packaging (a5f2ecd). The use case is marking helper files as internal so consumers don't grow dependencies on them.

modules/SPEC.md
claymcleod: I went a slightly different direction after thinking about rich rendering: `doi` and `biotools` are identifiers with canonical resolvers rather than URLs, so tool entries now carry an `ids` array of CURIEs (`doi:10.21105/joss.04704`, `biotools:csvkit`) using identifiers.org/Bioregistry prefixes, alongside the `url` field (e9ec7a1). Any registered prefix works without a spec change, and tooling gets the bare identifier for API lookups instead of parsing it back out of a URL.

modules/SPEC.md
claymcleod: I kept the selector named `commit` deliberately: it names what you're selecting by (alongside `tag` and `branch`), and its value may be any unique prefix rather than a full hash. The lockfile field is now `sha` per your other comment, since that value is always the fully expanded 40-character hash. The selector is the input; `sha` is the resolved output.

modules/SPEC.md
claymcleod: This is now explicit in the spec: a `path` dependency doesn't extend the consuming module's boundary. It names a separate module, exactly as a `git` dependency does; the target must be a module root with its own `module.json`, validated independently (e8e680b). The file-tree constraints (including the new rule that quoted imports may not escape the module root) apply to each module's own tree. So both halves hold: the boundary is preserved per module, and dependencies (local or remote) are the only way to reach anything outside it.

modules/SPEC.md
claymcleod: Yes. An entrypoint may contain only `version` and `import` statements; the imported declarations form the root module surface.

modules/SPEC.md
claymcleod: Agreed, and the spec now frames it that way: `exclude` is an import-surface policy, not a filesystem security boundary (a5f2ecd). It exists so authors can mark files as internal and keep consumers from growing accidental dependencies on them through symbolic imports; it makes no attempt to hide bytes from someone who has the module source.

SPEC.md
claymcleod: For my own preference, I'm inclined to agree with @patmagee and @hexylena on this one. If ones prefers using the namespaced syntax to know where these references came from, nothing prevents them from using the existing form 1 and doing that. Further, I view the benefit gained from not having hierarchical namespacing (which itself would be a nightmare) to be a good tradeoff.

claymcleod:
I've updated the PR with a litany of updates based on my implementation experiences in Sprocket. Additionally, I believe I've resolved (or at least addressed) every concern now. Please rereview and let me know what you think.

modules/SPEC.md
patmagee: makes sense

patmagee (approved):
This is a behemoth amount of work, but I think it lands somewhere really great. Amazing work @claymcleod on all of this!

SPEC.md
a-frantz: I agree with what everyone's said here. I don't like `*` imports either, but I think that's a matter of preference I don't see the need to enforce via spec. I would potentially even argue WDL linters should warn against this form of import, but that's something that can be hashed out later down the road.

modules/SPEC.md
```diff
@@ -0,0 +1,622 @@
...
+4. A quoted import inside a module must resolve to a file inside the same module root. An import such as `import "../shared.wdl"` that escapes the module root makes the module invalid, even if the target file exists on disk.
+5. A module that, when materialized on disk, would produce any path failing these rules is invalid; engines must refuse to load it and surface a clear error identifying the offending entry.
+
+## Manifest File (`module.json`)
+
+The manifest file for a module is always located at the module's root directory with the name `module.json`. It is a JSON document containing the fields described in the subsections below. A companion JSON Schema is available at [`schemas/module.schema.json`](schemas/module.schema.json).
+
+### Core Fields
+
+- **`name`** (string, required). A human-readable display name for the module (e.g., `"csvcut"`, `"csvkit-sort"`). Used by tooling for display. Not used for dependency resolution.
```
a-frantz, line 109: I'd include an example with a space and capital letters for extra clarity that these names don't need to fit any particular format
```suggestion
- **`name`** (string, required). A human-readable display name for the module (e.g., `"csvcut"`, `"csvkit-sort"`, `"BAM Sort"`). Used by tooling for display. Not used for dependency resolution.
```

modules/SPEC.md
```diff
@@ -0,0 +1,622 @@
...
+  module.json
+  index.wdl
+```
+
+A consumer with a dependency named `csvcut` then chooses how to bring those names into scope using the import forms defined in [`SPEC.md`](../SPEC.md#-import-forms):
+
+```wdl
+version 1.4
+
+import csvcut              # tasks/workflows reachable as `csvcut.*`; UDTs in scope unqualified
```
a-frantz, line 274: nit: I don't think "UDT" is a common enough abbreviation to include here

modules/SPEC.md
```diff
@@ -0,0 +1,622 @@
...
+
+The default name `index.wdl` mirrors package index conventions: it names the file that represents the directory as a whole. A module that prefers `main.wdl` or another project-specific name can use the `entrypoint` field without changing resolution semantics.
+
+The entrypoint provides the module's **default surface**: when a consumer writes a **root module import** of the dependency (`import samtools`, with no sub-path), the engine resolves the import to the entrypoint file. The entrypoint file uses ordinary quoted imports to pull in its sibling files, exactly as defined in [`SPEC.md`](../SPEC.md#-import-forms), and the names brought into its scope become the surface that root module consumers see. Tasks and workflows enter that scope only through the scope-merging forms (`import * from`, `import { ... } from`); a form 1 (namespaced) import contributes the user-defined types it copies into scope, while its tasks and workflows remain referenceable only by namespace within the entrypoint and are not part of the module's surface.
```
a-frantz, line 255: > The entrypoint file uses ordinary quoted imports to pull in its sibling files

"quoted imports"? Is this a typo or can an `index.wdl` not use symbolic imports?

modules/SPEC.md
```diff
@@ -0,0 +1,622 @@
...
+
+A sub-path may not escape the dependency's module folder. The identifier-only component rule already forbids `..`, so a path such as `samtools/../another_file` is rejected at parse time rather than being resolved against the surrounding repository. This guarantee is normative: engines may rely on it to fetch only the module folder—via sparse Git checkout, partial clone, or any other mechanism that omits the rest of the repository—without risk that a symbolic import will later require content outside that folder.
+
+Sub-path resolution is a direct file lookup. Intermediate directories along the sub-path do not need to contain `module.json` files, and any nested `module.json` files that happen to be present along the way are ignored.
```
a-frantz, line 331: I've been waiting for this explanation of module structure during my whole read so far. This should be explained earlier in the document IMO.

a-frantz (commented):
really just some clarifications and nits. But I'm stoked about the overall shape and all the awesome use cases this is going to enable 🚀

modules/SPEC.md
a-frantz: I see now, this is not a typo but a consequence of symbolic imports only working for _imported modules_, and as the `index.wdl` (or other entrypoint WDL) is "initiating" or defining a _new_ module, it must use quoted imports.

This strikes me as awkward. My confusion stems from going into this proposal expecting quoted imports to be deprecated along with the introduction of symbolic imports.

My ideal vision for WDL involves a deprecation/removal of quoted imports, but that doesn't need to be shoehorned into this set of changes. I would not send us back to the drawing board over this, so consider this thread resolved 👍

a-frantz:
I've started implementing a few modules using the [stjudecloud/workflows](https://github.com/stjudecloud/workflows) repo as a jumping off point: I really liked this proposal at first, when I was just focused on a small handful of tools. But stepping back and looking at the entirety of our `tools/` directory has led me to request an addition: a `modules.json` (plural) that allows multiple modules to share a directory.

# Rationale

Each "tool" should be encapsulated by an individual module, as this will lead to the most stable API surfaces. However the current proposal necessitates a proliferation of directories and files that is burdensome. The `tools/` directory is flat (aside from the `test` dir) and contains a few dozen WDL files. Making each of those WDL files its own module turns each WDL file into a directory containing `index.wdl`, `module.json`, `module.sig` and `test/index.yaml`. **That's 2 levels of directory nesting with >=4 files per tool.**
```
$  tree tools/
tools/
├── arriba.wdl
├── bwa.wdl
├── CHANGELOG.md
├── deeptools.wdl
├── fastp.wdl
├── fastqc.wdl
├── fq.wdl
├── gatk4.wdl
├── htseq.wdl
├── kraken2.wdl
├── librarian.wdl
├── md5sum.wdl
├── mosdepth.wdl
├── multiqc.wdl
├── ngsderive.wdl
├── picard.wdl
├── qualimap.wdl
├── sambamba.wdl
├── samtools.wdl
├── star.wdl
├── test
│   ├── arriba.yaml
│   ├── bwa.yaml
│   ├── deeptools.yaml
│   ├── fastp.yaml
│   ├── fq.yaml
│   ├── gatk4.yaml
│   ├── htseq.yaml
│   ├── kraken2.yaml
│   ├── librarian.yaml
│   ├── md5sum.yaml
│   ├── mosdepth.yaml
│   ├── multiqc.yaml
│   ├── ngsderive.yaml
│   ├── picard.yaml
│   ├── qualimap.yaml
│   ├── sambamba.yaml
│   ├── samtools.yaml
│   ├── star.yaml
│   └── util.yaml
└── util.wdl

2 directories, 40 files
```
turns into:
```
  tree
.
├── arriba
│   ├── index.wdl
│   ├── module.json
│   ├── module.sig
│   └── test
│       └── index.yaml
├── bwa
│   ├── index.wdl
│   ├── module.json
│   ├── module.sig
│   └── test
│       └── index.yaml
├── CHANGELOG.md
├── deeptools
│   ├── index.wdl
│   ├── module.json
│   ├── module.sig
│   └── test
│       └── index.yaml
├── fastp
│   ├── index.wdl
│   ├── module.json
│   ├── module.sig
│   └── test
│       └── index.yaml
├── fastqc
│   ├── index.wdl
│   ├── module.json
│   ├── module.sig
│   └── test
│       └── index.yaml
├── fq
│   ├── index.wdl
│   ├── module.json
│   ├── module.sig
│   └── test
│       └── index.yaml
├── gatk4
│   ├── index.wdl
│   ├── module.json
│   ├── module.sig
│   └── test
│       └── index.yaml
├── htseq
│   ├── index.wdl
│   ├── module.json
│   ├── module.sig
│   └── test
│       └── index.yaml
├── kraken2
│   ├── index.wdl
│   ├── module.json
│   ├── module.sig
│   └── test
│       └── index.yaml
├── librarian
│   ├── index.wdl
│   ├── module.json
│   ├── module.sig
│   └── test
│       └── index.yaml
├── md5sum
│   ├── index.wdl
│   ├── module.json
│   ├── module.sig
│   └── test
│       └── index.yaml
├── mosdepth
│   ├── index.wdl
│   ├── module.json
│   ├── module.sig
│   └── test
│       └── index.yaml
├── multiqc
│   ├── index.wdl
│   ├── module.json
│   ├── module.sig
│   └── test
│       └── index.yaml
├── ngsderive
│   ├── index.wdl
│   ├── module.json
│   ├── module.sig
│   └── test
│       └── index.yaml
├── picard
│   ├── index.wdl
│   ├── module.json
│   ├── module.sig
│   └── test
│       └── index.yaml
├── qualimap
│   ├── index.wdl
│   ├── module.json
│   ├── module.sig
│   └── test
│       └── index.yaml
├── sambamba
│   ├── index.wdl
│   ├── module.json
│   ├── module.sig
│   └── test
│       └── index.yaml
├── samtools
│   ├── index.wdl
│   ├── module.json
│   ├── module.sig
│   └── test
│       └── index.yaml
├── star
│   ├── index.wdl
│   ├── module.json
│   ├── module.sig
│   └── test
│       └── index.yaml
└── util
    ├── index.wdl
    ├── module.json
    ├── module.sig
    └── test
        └── index.yaml

41 directories, 81 files
```
We are increasing the directory nesting (~2 dirs -> ~40 dirs) and file count (~40 files -> ~80 files) and I think the added file hierarchy complexity would be too burdensome for maintainers such as myself.

# Proposed solution

A `modules.json` that flattens the above structure by merging all individual `module.json` objects. The result would look like:
```
$  tree
.
├── arriba.wdl
├── bwa.wdl
├── CHANGELOG.md
├── deeptools.wdl
├── fastp.wdl
├── fastqc.wdl
├── fq.wdl
├── gatk4.wdl
├── htseq.wdl
├── kraken2.wdl
├── librarian.wdl
├── md5sum.wdl
├── modules.json
├── modules.sig
├── mosdepth.wdl
├── multiqc.wdl
├── ngsderive.wdl
├── picard.wdl
├── qualimap.wdl
├── sambamba.wdl
├── samtools.wdl
├── star.wdl
├── test
│   ├── arriba.yaml
│   ├── bwa.yaml
│   ├── deeptools.yaml
│   ├── fastp.yaml
│   ├── fq.yaml
│   ├── gatk4.yaml
│   ├── htseq.yaml
│   ├── kraken2.yaml
│   ├── librarian.yaml
│   ├── md5sum.yaml
│   ├── mosdepth.yaml
│   ├── multiqc.yaml
│   ├── ngsderive.yaml
│   ├── picard.yaml
│   ├── qualimap.yaml
│   ├── sambamba.yaml
│   ├── samtools.yaml
│   ├── star.yaml
│   └── util.yaml
└── util.wdl

2 directories, 42 files
```
Instead of restructuring ~20 WDL files and introducing messy directory nesting, we just add two new files (`modules.json` and `module.sig` (every module defined here would share the same sig))

claymcleod:
It's an interesting idea @a-frantz. There are a few things that I think make it intractable as an addition and instead indicate it would be a total redesign:

First, a module is essentially keyed by (a) the Git repository and (b) the path within that repository:

> For cycle detection, module identity is the module's source coordinates—the Git repository URL and path for Git sources, or the resolved directory for local path sources—irrespective of version or selector.

Changing that is a fairly consequential change that extends to entrypoints, versioning, hashing, etc.

Second, sharing one `module.sig` across these modules would couple their integrity boundaries. Under the current proposal, a signature covers the module’s content hash, and that hash includes every file beneath the module root. If `fastp.wdl` and `fastqc.wdl` shared a module root, changing either file would invalidate the shared signature and change the checksum recorded for every nominal module rooted there. So that's another thing that suggests this would require a major redesign to work.

Last,  we could preserve the flat layout by treating `tools/` as one module and exposing each WDL file through a sub-path. However (as you state), that makes the entire directory one release unit. A breaking change to one interface would require a major version bump for the entire module, preventing consumers on the previous major from receiving unrelated updates.

Unfortunately, that leaves us with a situation where there is no solution that addresses all concerns. With that understanding, I would advocate that we not complicate the mental model by adding `modules.json`; the current model provides the user the choice of prioritizing low file/folder count (pushing them to adopt one giant `tools` module and deal with the versioning problem) _or_ prioritizing module independence (pushing them to the more files/folders). So, given that the range of options is able to be expressed, though not perfect, I think it's the best option for this initial release.

Perhaps we can make tooling outside the language specification to make management of this many files and folder manageable?

PR #777, opened 2026-07-15
feat: adds postfix unwrap operator for optional values

### LINKED ISSUE #551: Question mark operator for unwrapping optional values
This is purely syntactic sugar for `select_first([x])` where `x` is a value that may be `None`.

```
Int? x = 5
Int x1 = x?
Int? y = None
Int y1 = y?  # error
```

 If we add this operator then we can get away from special-casing optional behavior within string interpolations. I.e. `"~{x + y}"` would no longer be allowed (i.e. deprecated in 1.2 and disallowed in 2.0) and instead you'd have to write `"~{x? + y?}"`, just as you would in an expression outside of an interpolation.

claymcleod:
Added the postfix `!` operator to unwrap a value of type `T?` as `T`, raising an error when the value is `None`. This provided a concise equivalent to `select_first([x])`, including in inline expressions and command placeholders. Closes #551.

Specified its typing, precedence, lexing interaction with `!=`, and placeholder behavior, with runnable success and failure examples. Within a placeholder, an unwrap failure is replaced by the empty string under the existing rule for errors caused by evaluating `None`; unrelated runtime errors and static-analysis errors remain errors. Flow-sensitive narrowing of optional values remained outside this change for separate consideration, and existing optional interpolation behavior remained unchanged.

For OpenWDL team members:


SPEC.md
```diff
@@ -2425,6 +2428,140 @@ In operations on mismatched numeric types (e.g., `Int` + `Float`), the `Int` is
...
+
+```json
+{
+  "fail": true
+}
+```
+</p>
+</details>
+
+Within [expression placeholders](#expression-placeholders-and-string-interpolation), an error raised by the unwrap operator is subject to the rules described in [Expression Placeholder Coercion](#expression-placeholder-coercion), i.e., the placeholder evaluates to the empty string.
```
claymcleod, line 2516: > If an expression within a placeholder evaluates to `None`, and either causes the entire placeholder to evaluate to `None` or causes an error, then the placeholder is replaced by the empty string.

claymcleod:
This vote has passed with 1 for (Clay McLeod) and 0 votes against.

PR #778, opened 2026-07-16
feat: adds input and output validation

### LINKED ISSUE #657: Support Input Validation for Workflow and Tasks
A common problem that consumers of WDL face is the need to perform input validation at both the tasks and the workflow level. The current status quo requires spinning up a task to perform any validation which can be both costly and time consuming. A fail fast mechanism would allow immediate feedback to the end user.

To accomplish this, I propose a new Syntax or function for performing validation within the input block of a task / workflow. This approach has the benefit of clearly showing that validation is part of the input blocks, however it does mean that it cannot happen anywhere else. An additional benefit is the fact the all validations can be evaluated and returned to the user instead of failing at the first one.

```wdl
workflow foo {
  input {
    String name
    Int age

    # New Validation Section
    validation {
      length(name) > 0: "Name must not be empty"
      age > = 0: "Age must be greater then 0"
      age < 110: "Really? Is someone actually that old?"
    }
  }
}
```

The above mechanism relies on new syntax to provide a clear and concise way to validate inputs. However we may want to make this more flexible with an engine function

```wdl
workflow foo {
  input {
    String name
    Int age
  }

  validate(length(name) > 0, "Name must not be empty")
  validate(age > 0, "Age must be greater then 0")
  validate(age < 110, "Really? Is someone actually that old")
}
```

## Validation Errors

When a workflow fails for a validation error, we would want to make a guarantee that the failure message is returned according to some standardized format. Since there is no format currently (see #611), a statement that the engine must return the relevant validation errors and their messages would be as much as we could provide

claymcleod:
Added nested `validation` blocks to task and workflow input and output sections so authors could express boundary conditions without running a helper task. Input validation aggregated failures before execution, while output validation checked completed values before exposing them. Closes #657.

Restricted input validation dependencies to input declarations to preserve DAG construction, required comma-separated clauses for unambiguous parsing, and reserved `validation` in WDL 1.4. Input failures were made non-retryable; task output failures followed `max_retries` because they occur after command execution.

For OpenWDL team members:

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -94,6 +94,7 @@ Revisions to this specification are made periodically in order to correct errors
         - [Special Case: Versioning Filesystem](#special-case-versioning-filesystem)
       - [Input Type Constraints](#input-type-constraints)
         - [Optional inputs with defaults](#optional-inputs-with-defaults)
+      - [✨ Input Validation](#-input-validation)
     - [Private Declarations](#private-declarations)
     - [Environment Variables](#environment-variables)
       - [String Escaping and Injection Prevention](#string-escaping-and-injection-prevention)
@@ -102,6 +103,7 @@ Revisions to this specification are made periodically in order to correct errors
       - [Stripping Leading Whitespace](#stripping-leading-whitespace)
     - [Task Outputs](#task-outputs)
       - [File, Directory, and Optional Outputs](#file-directory-and-optional-outputs)
+      - [✨ Output Validation](#-output-validation)
     - [Evaluation of Task Declarations](#evaluation-of-task-declarations)
     - [Requirements Section](#-requirements-section)
       - [Units of Storage](#units-of-storage)
@@ -548,6 +550,7 @@ struct
 task
 then
 true
+validation
 version
 workflow
 ```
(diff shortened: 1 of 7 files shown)


SPEC.md
```diff
@@ -5325,11 +5544,117 @@ task output_subset {
 }
 ```
 
+#### ✨ Output Validation
+
+An `output` section may end with a single, optional `validation` block. Output validation clauses use the same `<Boolean expression>: <String expression>` syntax and static types as [input validation](#-input-validation). A `validation` block must follow all declarations in its containing `output` section.
+
+Expressions in output validation clauses have the same visibility as output declaration expressions. The execution engine first evaluates every output declaration. If all output declarations are evaluated successfully, it evaluates every validation clause in source order using the same message aggregation and error handling rules as input validation. Message expressions are evaluated only for conditions that evaluate to `false`.
+
+Output validation occurs before the outputs are exposed to the caller or returned as final workflow outputs. If any clause fails, the task or workflow fails and none of its outputs are exposed. A task output validation failure occurs after the command has completed and is subject to the task's existing [`max_retries`](#max_retries) behavior.
```
a-frantz, line 5553: > A task output validation... is subject to the task's existing `max_retries` behavior.

This seems like wasted computation to me. What would the use case for this be? I'd prefer if output validation fails immediately returned, even if there are retries remaining. Maybe I'm missing a use case though?

PR #787, opened 2026-07-16
fix: make `read_tsv` return types statically resolvable

### LINKED ISSUE #690: Static type calculation for `read_tsv` isn't possible given the current overloads
As of 1.2, [`read_tsv`](https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#read_tsv) has the following overloads as covered by the "variants" section in the function's description:

* `Array[Array[String]] read_tsv(File, [false])`
* `Array[Object] read_tsv(File, true)`
* `Array[Object] read_tsv(File, Boolean, Array[String])`

The first two overloads are problematic as they're specified in terms of a _value_ and not a _type_.

An example of `read_tsv(file, header)` where `file` is of type `File` and `header` is type `Boolean` demonstrates the problem; we're unable to statically determine the resulting type of this expression (is it `Array[Array[String]]` or `Array[Object]`?). That is to say, called in this fashion, the function is overloaded in its _return type only_ which WDL has no facility to support.

Currently, the static analyzer used by `sprocket` defines `read_tsv` with these overloads instead:

* `Array[Array[String]] read_tsv(File)`
* `Array[Object] read_tsv(File, Boolean)` (if the second argument is `false`, it would be a runtime error)
* `Array[Object] read_tsv(File, Boolean, Array[String])`

Removing the optional argument to the first overload means that you can't call `read_tsv(file, false)` to get a resulting `Array[Array[String]]`, but as the optional (and I would argue, unnecessary) `false` argument was added in 1.2, I doubt anyone has done this yet.

Ideally I'd like to see these overloads split apart into two different functions: `read_tsv` and `read_tsv_with_header`.

`read_tsv` (doesn't read a header):

* `Array[Array[String]] read_tsv(File)`  (compatible with WDL 1.0+)
* `Array[Object] read_tsv(File, Array[String]))`

`read_tsv_with_header`:

* `Array[Object] read_tsv_with_header(File, [Array[String]])`

claymcleod:
Corrected `read_tsv` so overload resolution depended only on argument types: the one-argument form returned `Array[Array[String]]`, while the two- and three-argument forms returned `Array[Object]`. The two-argument form now raised an error when its Boolean evaluated to `false`, and conformance examples covered both a Boolean variable and the failure case. Closes #690.

This correction targeted `wdl-1.2`, where the header-aware overloads were introduced. The same correction needs to be applied separately to the `wdl-1.3` and `wdl-1.4` branches after this merges.

For OpenWDL team members:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ Keep the changelog pleasant to read in the text editor:
 version 1.2.2
 ---------------------------
 
-+ Corrected the `read_tsv` overloads so that return types can be determined statically. `read_tsv(File, false)` is now an error; callers that need an `Array[Array[String]]` must omit the second argument ([#690](https://github.com/openwdl/wdl/issues/690)).
++ Corrected the `read_tsv` overloads so that return types can be determined statically. `read_tsv(File, false)` is now an error; callers that need an `Array[Array[String]]` must omit the second argument ([#787](https://github.com/openwdl/wdl/pull/787)).
 
 + Clarified that `after` is a reserved keyword ([#766](https://github.com/openwdl/wdl/pull/766)).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ Keep the changelog pleasant to read in the text editor:
 version 1.2.2
 ---------------------------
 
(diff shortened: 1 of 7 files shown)


peterhuene (approved):

claymcleod:
This vote has been approved with 2 for (Clay McLeod, Peter Huene) and 0 against.

PR #788, opened 2026-07-16
fix: define workflow `File` evaluation context

### LINKED ISSUE #581: Coercion of String to File outside a task: where does it happen?
In WDL 1.1, `String` has an "obvious and unambiguous" coercion to `File`:

https://github.com/openwdl/wdl/blob/main/versions/1.1/SPEC.md#type-coercion

What is this conversion, exactly?

For example, a natural way to implement a scatter is to have each copy of the body execute in parallel on a different machine. So if you do something like:

```
workflow Example {

    input {
        String filename = "/etc/hostname"
    }

    File firstFile = filename
    String firstContent = read_string(firstFile)
    scatter(iteration in range(10000)) {
        File file = filename
        String content = read_string(file)
    }

    output {
        String first = firstContent
        Array[String] others = content
    }

}
```

What is supposed to be true about the `others` array? Are the values all the same? Are they all the same as `first`? For that matter, is `first`always the same thing you would see if you ran `cat /etc/hostname` instead of executing the workflow?

When working with functions that take either `File` or `String`, the spec says that `String`s are interpreted "relative to the current working directory of the task"; presumably that would apply to coercions as well. But here we're not in a task, and a workflow as a distributed thing doesn't seem to me to have a natural working directory.

In the Toil workflow runner's WDL interpreter, I really want to be allowed to move workflow code to any machine at any line, or even between lines, so I can ship fragments of it to the cluster when they're ready to run, just like the tasks. But that [seems to be breaking](https://github.com/DataBiosphere/toil/issues/4554) some of the [GATK workflows](https://github.com/gatk-workflows/seq-format-conversion/blob/6b69dde1877de997effedd3388aeaca929db04da/paired-fastq-to-unmapped-bam.wdl) which assume that, at least at top-level scope, a workflow can always read the same filesystem its caller sees. So I'm trying to work out how legitimate that assumption is, and whether there are guarantees that go further than that.

claymcleod:
Issue #581 asks what resource a `String` denotes when it is coerced to a `File` in workflow scope. The remaining ambiguity was whether top-level and scattered declarations resolve paths against one stable filesystem or against whichever machine evaluates each expression. That left three observable questions unanswered: whether scatter iterations may read different machine-local files for the same path, whether their values must match the equivalent top-level declaration, and whether an absolute path such as `/etc/hostname` necessarily refers to the file visible from the caller's shell.

This specified one logical filesystem context for each workflow invocation. Top-level, conditional, scattered, and subworkflow expressions all resolve paths in that context even when evaluated on different machines; an engine must preserve the denoted resource or fail rather than reinterpret the path locally. Thus the top-level and scattered values in #581 refer to the same resource. They match the caller's filesystem only when the caller shares the workflow's filesystem context. This targets WDL 1.1 as the earliest maintained affected version and should be forward-ported to `wdl-1.2`, `wdl-1.3`, `wdl-1.4`, and `wdl-2.0`. Closes #581.

For OpenWDL team members:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,10 @@ version development
 version 1.1.4
 ---------------------------
 
++ Specified that workflow-level `File` values use a stable logical file system
+  context, including when workflow expressions are evaluated on different
+  machines.
+
 + Clarified that `after` is a reserved keyword ([#766](https://github.com/openwdl/wdl/pull/766)).
 
 version 1.1.3
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,7 +47,7 @@ version 1.1.4
 
 + Specified that workflow-level `File` values use a stable logical file system
   context, including when workflow expressions are evaluated on different
-  machines.
+  machines ([#788](https://github.com/openwdl/wdl/pull/788)).
 
 + Clarified that `after` is a reserved keyword ([#766](https://github.com/openwdl/wdl/pull/766)).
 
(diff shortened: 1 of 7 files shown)


claymcleod:
This vote has passed with 1 for (Clay McLeod) and 0 against.

PR #789, opened 2026-07-16
fix: define truncating remainder semantics in WDL 1.0

### LINKED ISSUE #682: Spec: clarification needed for remainder operator on float operands
Currently the [remainder `%` operator (i.e. modulo)](https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#binary-operators-on-primitive-types) is accepting of `Int` and `Float` as operand types.

For two `Int` types, there is no question as to how the operation should be performed as it involves only integer division, e.g. `a - ((a / b) * b)`, where `a / b` naturally truncates.

However, for `Float` operands or `Int` operands implicitly coerced to `Float` because the other operand is `Float`, floating point division is involved and there are [several different methods](https://en.wikipedia.org/wiki/Modulo#Variants_of_the_definition) for performing the operation.

I would assume the intention is to define the operator in terms of _truncation_ to align with most programming languages, but some notable languages are _flooring_ instead (e.g. Python, Ruby), so an explicit clarification is probably required for the WDL specification.

claymcleod:
Defined WDL 1.0 `%` as the remainder produced by truncating the exact quotient toward zero, including its behavior for negative operands and zero divisors. This resolved the ambiguity reported in #682 at the earliest affected WDL minor version.

The equivalent clarification still needs to be applied to the `wdl-1.1`, `wdl-1.2`, `wdl-1.3`, and `wdl-1.4` branches. This PR therefore references rather than closes #682 so the issue remains open until the later versions are updated.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@ Keep the changelog pleasant to read in the text editor:
 version development
 ---------------------------
 
++ Clarified that the remainder operator truncates the quotient toward zero and
+  raises an error when the right-hand operand is zero
+  ([#682](https://github.com/openwdl/wdl/issues/682)).
+
 + Fixed description of ternary operator to say that the type, not the value,
   of the if-then-else is the same regardless of which side is evaluated.
   [PR 476](https://github.com/openwdl/wdl/pull/476) by @notestaff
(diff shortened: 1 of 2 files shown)


claymcleod:
This vote has passed with 1 for (Clay McLeod) and 0 against.

PR #790, opened 2026-07-16
fix: add explicit `String` path overloads

### LINKED ISSUE #683: `basename` and `size` should have explicit `String` overloads
## The `basename` function

Currently the [`basename`](https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#basename) function has two overloads defined in the spec:

```wdl
String basename(File, [String])
String basename(Directory, [String])
```

But it then goes on to say:

> Parameters
>
>   File|Directory: Path of the file or directory to read. If the argument is a String, it is assumed to be a local file path relative to the current working directory of the task.

The issue here is that `String` is coercible to both `File` and `Directory` and so calling it with a `String` is technically ambiguous; the ambiguity was introduced with the addition of the `Directory` type in 1.2.

An ambiguity like that makes it more difficult to do the type calculus on an expression in a generic way, especially for function calls that may have type parameters. While in _this_ specific case, we know that `basename("foo")` has type `String` for its expression because all of the overloads that are ambiguous return `String`, that may not be the case for some other function where the overloads differ on return type. In both cases I would want my type calculator to report on the ambiguity in binding the function arguments.

Additionally, depending on how an engine is implemented, it might not be possible to know what the "original" type was following a coercion, so it might prove difficult to fulfill the requirement of "If the argument is a String, it is assumed to be a local file path relative to the current working directory of the task." unless the engine knows the call was dispatched to the `String` overload of the function.

I propose we add the following overload to `basename` and amend its parameter description to `String|File|Directory`:

```
String basename(String, [String])
```

## The `size` function

The [`size`](https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#size)

```wdl
Float size(File|File?, [String])
Float size(Directory|Directory?, [String])
Float size(X|X?, [String])
```

For the same reason as above, a call with `String` for the first argument would be ambiguous between all the first two overloads only; it would not bind to the third overload due to the type constraint on `X` (`any compound type that contains File or File? nested at any depth`), as `String` does not satisfy that constraint.

Thus, I propose adding this additional overload:

```wdl
Float size(String|String?, [String])
```

We should also include the same note as that for `baseline` regarding the `String` overload, i.e. "If the argument is a String, it is assumed to be a local file path relative to the current working directory of the task."

claymcleod:
Added explicit `String` overloads for `basename` and `size` to remove the overload ambiguity introduced when WDL 1.2 added `Directory`. Expanded the `size` example to cover required and optional string paths. Closes #683.

This targeted `wdl-1.2`, the earliest affected minor version. The change must be forward-ported to `wdl-1.3` and `wdl-1.4` after it merges.

For OpenWDL team members:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ Keep the changelog pleasant to read in the text editor:
 version 1.2.2
 ---------------------------
 
++ Added explicit `String` overloads to `basename` and `size` to prevent ambiguous coercion to `File` or `Directory` ([#683](https://github.com/openwdl/wdl/issues/683)).
+
 + Clarified that `after` is a reserved keyword ([#766](https://github.com/openwdl/wdl/pull/766)).
 
 + Clarified that `env` is a reserved keyword ([#764](https://github.com/openwdl/wdl/pull/764)).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ Keep the changelog pleasant to read in the text editor:
 version 1.2.2
 ---------------------------
 
-+ Added explicit `String` overloads to `basename` and `size` to prevent ambiguous coercion to `File` or `Directory` ([#683](https://github.com/openwdl/wdl/issues/683)).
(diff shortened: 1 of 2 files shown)


peterhuene (approved):

claymcleod:
This vote has passed with 2 for (Peter Huene, Clay McLeod) and 0 against.

PR #791, opened 2026-07-16
feat: define numeric literal forms

### LINKED ISSUE #774: Clarify supported numeric literal forms
WDL should explicitly define which numeric literal forms are valid in WDL source. The `1.3` spec defines the `Int` and `Float` value domains, and it shows decimal examples such as `0` and `27.3`, but it does not give a grammar for numeric literals. That leaves implementations to infer whether forms such as exponent notation, hexadecimal integers, or octal integers are part of the language.

For `1.4`, we should decide whether each of these numeric literal forms is supported WDL syntax, unsupported syntax, or deliberately implementation-defined:

- decimal integers, e.g., `0`, `42`
- hexadecimal integers with `0x` or `0X`, e.g., `0xCAFE`
- octal integers with `0o` or `0O`, e.g., `0o755`
- legacy leading-zero octal integers, e.g., `0755`
- decimal floats with a decimal point, e.g., `1.0`, `1.`, `.5`
- floats with exponent notation, e.g., `1e3`, `1E10`, `1.25e-3`

The decision should cover both source literals and related parsing surfaces where users encounter numbers, including JSON inputs and `read_int`/`read_float`, if those surfaces need different rules.

The issue should also add conformance tests for the accepted and rejected forms. At minimum, tests should exercise the accepted examples above if they remain valid, and should include negative cases for forms we reject, such as malformed prefixes, invalid octal digits, incomplete exponents, and non-finite values if they are not already covered elsewhere.

claymcleod:
Defined decimal, hexadecimal, octal, and binary `Int` forms and the accepted decimal `Float` forms, while rejecting legacy leading-zero octal syntax. Applied the same forms to `read_int` and `read_float`, kept JSON parsing governed by the JSON number grammar, and added positive and negative conformance cases. Closes #774.

This targeted `wdl-1.1`, the earliest maintained affected minor version. The change must be forward-ported to `wdl-1.2`, `wdl-1.3`, and `wdl-1.4` after it merges.

For OpenWDL team members:

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -552,8 +552,9 @@ An `Int` literal has one of the following forms:
 * Decimal: `0`, or a digit from `1` through `9` followed by zero or more ASCII decimal digits (`0` through `9`).
 * Hexadecimal: `0x` or `0X` followed by one or more hexadecimal digits (`0` through `9`, `a` through `f`, or `A` through `F`).
 * Octal: `0o` or `0O` followed by one or more octal digits (`0` through `7`).
+* Binary: `0b` or `0B` followed by one or more binary digits (`0` or `1`).
 
-Leading zeros are not allowed in a decimal literal, so a legacy octal form such as `0755` is invalid. WDL does not support binary or other base prefixes in integer literals.
+Leading zeros are not allowed in a decimal literal, so a legacy octal form such as `0755` is invalid. WDL does not support other base prefixes in integer literals.
 
 ##### Floating-Point Literals
 
@@ -580,6 +581,8 @@ A decimal numeric literal without a decimal point or exponent is an `Int` litera
       Int uppercase_hexadecimal = 0XCAFE
       Int lowercase_octal = 0o755
       Int uppercase_octal = 0O755
+      Int lowercase_binary = 0b101010
+      Int uppercase_binary = 0B101010
       Float decimal_point = 1.0
       Float trailing_point = 1.
       Float leading_point = .5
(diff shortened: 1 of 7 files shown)


shields/sprocket.json
```diff
@@ -1,6 +1,6 @@
 {
   "schemaVersion": 1,
   "label": "Sprocket / WDL 1.1",
-  "message": "146/146 passed",
-  "color": "brightgreen"
+  "message": "152/153 passed",
```
peterhuene: I'd actually expect more tests to fail for `sprocket` given that it (now definitely incorrectly) supports both `0` and `0x` prefixes.

SPEC.md
```diff
@@ -543,6 +545,205 @@ The following primitive types exist in WDL:
       * Remote files must be treated as read-only.
       * A remote file is only required to be vaild at the time that the execution engine needs to localize it.
 
+##### Integer Literals
+
+An `Int` literal consists of one or more ASCII decimal digits (`0` through `9`). It is interpreted in base ten, including when it has leading zeros; for example, `0755` has the value `755`. WDL does not support hexadecimal, octal, binary, or other base prefixes in integer literals.
```
peterhuene: I'm not sure where I got the idea that the grammar supported base prefixes, to be honest; looking back at the `wdl-parsers` repo it's definitely _not_ in the kinda-official antlr files.

So I can't explain why I added support for it in Sprocket.

My first instinct was to push back at limiting it to base 10 representation as it felt limiting.

It's probably not common to use in bioinformatics, but WDL is, in theory, general purpose.

Should it have a mechanism to represent base 8 and base 16 (and base 2)?

PR #792, opened 2026-07-16
docs: clarify defaulted input type semantics in WDL 1.1

### LINKED ISSUE #761: Clarify "implicitly optional" phrasing for inputs with defaults
The ["Optional inputs with defaults"](https://github.com/openwdl/wdl/blob/wdl-1.3/SPEC.md#optional-inputs-with-defaults) section currently says:

> Inputs with default initializers are implicitly optional: callers may omit the input or supply None whether or not its declared type carries the optional quantifier ?

This came up during review of stjude-rust-labs/sprocket#724, where @a-frantz [pointed out](https://github.com/stjude-rust-labs/sprocket/pull/724#pullrequestreview-3943550728) that the phrasing is misleading. The input itself isn't optional—it always resolves to a concrete value (either caller-supplied or the default). What's optional is the caller's obligation to provide it. Supplying `None` just means "use the default," not that the value becomes `None`. As Ari put it, the declaration's type doesn't change; it's just an alternative syntax for falling back to the default.

IMO the paragraph should be reworded to make this distinction clearer, since "implicitly optional" reads like the type itself becomes optional.

Separately, this behavior was formalized in v1.2 (cf. #462), but cromwell and miniwdl already implement it regardless of WDL version. It would be worth back-porting this clarification to v1.0 and v1.1 so the spec matches existing engine behavior rather than leaving it undefined in earlier versions.

claymcleod:
Clarified that inputs with default initializers retained their declared types when callers omitted them or supplied `None`, and backported the WDL 1.2 behavior table to WDL 1.1. The changelog entry was added under the unreleased WDL 1.1.4 section. Closes #761.

### Checklist
- [ ] Valid examples WDLs were added or updated in SPEC.md (see the [guide](https://github.com/openwdl/wdl-tests/blob/main/docs/MarkdownTests.md) on writing markdown tests)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,10 @@ version development
 version 1.1.4
 ---------------------------
 
++ **Backported from WDL 1.2.** Clarified that inputs with default initializers
+  retain their declared types when callers omit them or supply `None`
+  ([#761](https://github.com/openwdl/wdl/issues/761)).
+
 + Clarified that `after` is a reserved keyword ([#766](https://github.com/openwdl/wdl/pull/766)).
 
 version 1.1.3
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,7 +47,7 @@ version 1.1.4
 
 + **Backported from WDL 1.2.** Clarified that inputs with default initializers
   retain their declared types when callers omit them or supply `None`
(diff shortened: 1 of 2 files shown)


peterhuene (approved):

claymcleod:
This vote has passed with a record of 2 for (Clay McLeod, Peter Huene) and 0 against.

PR #796, opened 2026-07-16
fix: preserve task output colocalization in WDL 1.1

### LINKED ISSUE #762: Extend same-parent colocalization guarantee to task outputs
It's common practice for tools to require companion files to be co-located (e.g., a data file and its index). The spec already requires that input files from the same parent directory be localized together, which handles the case where a user passes these files in directly. But if an upstream task _produces_ those files as outputs, there's no guarantee the engine will keep them together when delocalizing—so the downstream task can't rely on the same-parent rule.

The spec currently [recommends working around this](https://github.com/openwdl/wdl/blob/wdl-1.3/SPEC.md#task-input-localization) by symlinking files into a shared directory. But this is insufficient—some tools resolve symlinks to their root paths, which defeats the purpose of the colocalization entirely.

If we're going to enforce colocalization for inputs, we should probably do the same for outputs. IMO this isn't a backwards incompatible change—it's a constraint on engines, not on workflows, and it's the same category of spec tightening as #646.

claymcleod:
Required execution engines to keep task `File` outputs in the same storage directory when they shared a parent directory in the execution environment, closing the gap between output persistence and the existing input localization guarantee described in #762. This also addressed the companion-file portability concern in #289 without introducing a new file-bundle type. The change lands in the next unreleased WDL 1.1 patch; WDL 1.0 remains unchanged.

This replaces #794, which GitHub closed when its head branch moved from WDL 1.0 to WDL 1.1. It references rather than closes #762 because equivalent wording still needs to be forward-ported to `wdl-1.2`, `wdl-1.3`, `wdl-1.4`, and `wdl-2.0`.

### Checklist


claymcleod:
This vote has passed with a vote of 1 for (Clay McLeod) to 0 against.

Issue #771, opened 2026-05-10
Add `read_table` stdlib function for typed tabular ingestion


claymcleod:
## Motivation

WDL workflows that consume sample sheets currently fall back to the deprecated `Object` type via `read_objects` or hand-roll parsing in task command blocks. No idiomatic typed path exists for reading tabular data into structured WDL values.

Closing this gap strengthens one of WDL's core advantages: its strong type system. Tabular inputs like sample sheets are ubiquitous in bioinformatics (see the [appendix](#sample-sheet-prevalence-in-adjacent-ecosystems) for a survey of how nf-core and Snakemake pipelines rely on them), and the language should be able to ingest them directly into typed structs rather than forcing users to weaken their types or push parsing logic into Bash.

## Proposal

Add a new stdlib function `read_table` that reads a delimited file and coerces each row to an instance of a caller-specified struct type `T`. The function is generic over `T`: the file's header row maps column names to the fields of `T`, and each subsequent row produces one `T` value with per-field type coercion.

### Signature (strawman)

```
Array[T] read_table(File)
Array[T] read_table(File, String delimiter)
Array[T] read_table(File, String delimiter, Boolean strict)
```

- `T` is a struct type. The type is inferred from the call-site assignment (e.g., `Array[InputRow] rows = read_table(tsv)`).
- `delimiter` defaults to `"\t"` (tab), matching the existing `read_tsv` convention. Users may pass `","` for CSV or any other single-character delimiter.
- `strict` defaults to `true`. When `true`, the engine raises an error if the file contains columns whose headers do not correspond to any field of `T`. When `false`, extra columns are silently discarded; engines are encouraged to emit a warning but are not required to.

### Header matching

The first line of the file is always a header row. Each header value must match exactly one field name of `T` (case-sensitive). Extra headers that do not correspond to any field of `T` are handled according to the `strict` parameter: in strict mode (the default), they cause an error; in non-strict mode, they are discarded. Missing headers for required (non-optional) fields of `T` are always an error regardless of the `strict` setting. Missing headers for optional fields of `T` default to `None` for every row.

### Per-field coercion rules

Each cell is coerced according to the type of the corresponding struct field:

- **`Int`**: The cell is parsed as a strict decimal integer. An optional leading `+` or `-` is permitted, followed by one or more digits. Leading and trailing whitespace is trimmed before parsing. Values containing a decimal point (e.g., `"3.0"`), scientific notation, hexadecimal, octal, or an empty string are rejected.
[...]

Issue #373, opened 2020-06-10
Automatic coercions cosidered harmful


orodeh:
# Automatic coercions considered harmful
Authors: O. Rodeh with J. Didion (DNAnexus Inc.)

This issue describes a line of thought regarding type checking for WDL programs. It is based on experience acquired while working with DNAnexus customers on their WDL pipelines. The author also wrote a WDL type checker, evaluator, and cloud executor.

# Coercions

WDL makes wide ranging use of coercions to convert one type to another. This is, in general, a well accepted idea used in a wide variety of programming languages. At DNAnexus we have run into several instances where WDL workflows errored out because static type checking was too lenient. This resulted in long running expensive customer pipelines failing. The issue is twofold: the turn around time is long, and the cost can run to thousands of dollars.

For example, under the WDL type system, a `File` is auto-coercible into `File?`. However, this is not the case in the DNAnexus simpler type-system. If you have a task that takes an `Array[File?]` and call it with `Array[File]` this will fail. For example:

```wdl
version 1.0

task foo {
  input {
    Array[File?] bams
  }
  command {}
}


workflow bar {
  input {
    Array[File] my_files
  }

  call long_expensive_calculation {}

  call foo { input : bams = my_files }

  output {}
}
```

Workflow `bar` will fail when it calls `foo`, after it has spent a long time running an expensive calculation. This is frustrating for the users. We could prevent this, as well as many other errors, by performing stricter type checking. Below is an eclectic list of automatic coercions that appears in [GATK pipelines](https://github.com/gatk-workflows). The GATK pipelines are large real world workflows that our customers expect to execute flawlessly on our platform. The code is a simplified version of the real program lines, stripping a way the context to focus on the main point.

Converting between `T` and `T?` occurs automatically
```
  # convert an optional file-array to a file-array
  Array[File] files = ["a", "b"]
  Array[File]? files2 = files

  # Convert an optional int to an int
  Int a = 3
  Int? b = a

  # An array with a mixture of Int and Int?
  Int c = select_first([1, b])
```

Converting between strings, integer, and floats "just works".
```
  # Convert an integer to a string
  String s = 3
  Int z = 1.3

  # converting Array[String] to Array[Int]
  Int d = "3"
  Array[Int] numbers = ["1", "2"]
```

The branches of a conditional may not resolve to the same type. Here, one side is `Array[String]`, the other is `Array[String?]`.
```                                                                                                                                
[...]

PR #797, opened 2026-07-16
feat: restore `String` coercions

claymcleod:
Restored globally valid coercions from `String` to `Int`, `Float`, and `Boolean`, defined their shared parsing rules in the type-coercion section, and made `read_int`, `read_float`, and `read_boolean` use those rules. Invalid values now pass static type analysis and raise errors during dynamic evaluation, which enabled string-valued tabular readers to populate typed structs without introducing a separate reader. Related to #771. If accepted, this change should be backported to `wdl-1.1` and subsequent versions.

The concern in #373 was valid: permissive automatic coercions can defer failures until after expensive work. This change deliberately favored target-directed conversion over having the type system assume that users' strings cannot represent primitive values. Fallibility was already part of WDL through conversions such as `String` to `File` or `Directory`, the existing `Float` to `Int` conversion, and `Union` values whose compatibility with the expected type is resolved dynamically; this made that tradeoff explicit rather than treating fallibility as a reason to reject these conversions categorically.

For OpenWDL team members:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ Keep the changelog pleasant to read in the text editor:
 version 1.4.0
 ---------------------------
 
-+ Added globally valid coercions from `String` to `Int`, `Float`, and `Boolean`.
++ Added globally valid coercions from `String` to `Int`, `Float`, and `Boolean` ([#797](https://github.com/openwdl/wdl/pull/797)).
 
 + Clarified that `after` is a reserved keyword ([#766](https://github.com/openwdl/wdl/pull/766)).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ Keep the changelog pleasant to read in the text editor:
 version 1.4.0
 ---------------------------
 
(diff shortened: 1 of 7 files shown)


peterhuene (approved):
The proposed changes look good to me with respect to correctness.

Regarding the feature overall, is the primary motivation for this being #771 or are there other use cases?

If the original concern was that implicit casts are bad (which is certainly valid), have we considered in the past using stdlib conversion functions? While not backportable to previous versions, it does make their use explicitly opt-in and can perhaps take additional optional parameters such as a radix for the `String` -> `Int` conversion, etc.

claymcleod:
#771 is _a_ driving use case, but I would say more generally that it's just a coercion that is accepted in the community. The justification is essentially (a) we already have fallible conversions, so allowing one here is not a regression and (b) since so much of the WDL world already assumes these coercions, it's probably better overall to semantically define them rather than omit them.

peterhuene:
I find the argument that there's a lot of WDL <= 1.0 out there that relies on the implicit coercions to be compelling enough to accept this PR and to backport the coercions.

PR #799, opened 2026-08-20
fix: correct malformed JSON in WDL 1.2 examples

adamnovak:
There's a missing comma in the JSON for the `relative_paths_context.wdl` test, so it can't be parsed correctly.

I also fixed file_sizes_task.wdl's JSON to parse, because it was missing a comma and using () for pairs. I changed the pairs to arrays, but I'm not sure they're not meant to be objects with "left" and "right" keys.

- [X] You have considered whether the `README.md` or other documentation needs updating to account for these changes.
- N/A You have updated the `CHANGELOG.md` describing the change and linking back to your pull request.
- [X] You have read and agree to the [`CONTRIBUTING.md`](https://github.com/openwdl/governance/blob/main/CONTRIBUTING.md) document.
- [X] You have considered adding or updating relevant example WDL tests to the specification.

For OpenWDL team members:

- [ ] Assign the appropriate individual to this PR.
- [ ] Triage the PR and add appropriate labels.

### DIFF
--- a/SPEC.md
+++ b/SPEC.md
@@ -951,7 +951,7 @@ Example output:
 
 ```json
 {
-  "relative_paths_context.result": "hello.txt"
+  "relative_paths_context.result": "hello.txt",
   "relative_paths_context.content": "hello"
 }
 ```
--- a/SPEC.md
+++ b/SPEC.md
@@ -8266,7 +8266,7 @@ Example output:
   "file_size.nested": {
     "a": (10, "out.txt"),
     "b": (50, null)
-  }
+  },
   "file_sizes.nested_bytes": 22.0
 }
 ```
--- a/SPEC.md
+++ b/SPEC.md
@@ -8264,8 +8264,8 @@ Example output:
   "file_sizes.created_file_bytes": 22.0,
   "file_sizes.multi_file_kb": 0.022,
   "file_size.nested": {
-    "a": (10, "out.txt"),
-    "b": (50, null)
+    "a": [10, "out.txt"],
+    "b": [50, null]
   },
   "file_sizes.nested_bytes": 22.0
 }
--- a/SPEC.md
+++ b/SPEC.md
@@ -951,7 +951,6 @@ Example output:
 
 ```json
 {
-  "relative_paths_context.result": "hello.txt",
   "relative_paths_context.content": "hello"
 }
 ```
@@ -960,7 +959,7 @@ Test config:
 
 ```json
 {
-  "exclude_outputs": ["result"]
+  "exclude_outputs": ["relative_paths_context.result"]
 }
 ```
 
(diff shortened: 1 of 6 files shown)


SPEC.md
```diff
@@ -951,7 +951,7 @@ Example output:
 
 ```json
 {
-  "relative_paths_context.result": "hello.txt"
+  "relative_paths_context.result": "hello.txt",
```
claymcleod: Remove this `result` expectation and change the test config to `"exclude_outputs": ["relative_paths_context.result"]`. The task produces `output.txt`, not `hello.txt`, and the extracted output is a flat object keyed by `relative_paths_context.result`; Spectool compares exclusions against that literal key, so `"result"` does not remove it.

SPEC.md
```diff
@@ -8264,9 +8264,9 @@ Example output:
   "file_sizes.created_file_bytes": 22.0,
   "file_sizes.multi_file_kb": 0.022,
   "file_size.nested": {
-    "a": (10, "out.txt"),
-    "b": (50, null)
-  }
+    "a": [10, "out.txt"],
+    "b": [50, null]
+  },
```
claymcleod: Inline this map into `size(...)`, remove `nested` from the output section, and delete its expected JSON entry. The declared value remains `Map[String, Pair[Int, File?]]`, and this specification says direct Pair serialization is an error; writing arrays here does not convert the WDL value, while the documented Pair-to-Array conversion applies only to homogeneous `Pair[X, X]`.

claymcleod (changes requested):

claymcleod:
Note to self: this fix also needs to be forward-ported to the `wdl-1.3` and `wdl-1.4` branches.

claymcleod:
(if you don't have time to address it, just let me know and I'll take it over)

SPEC.md
adamnovak: I think in general there might want to be more explanation in the spec about the `Pair`-related considerations for workflow authors. It's easy to write a subworkflow that can never be run alone, or a workflow that has default input values that can't be overridden by the user, or a workflow that gets all the way to the end and then sometimes producers a result that can't be output.

The spec says that an implementation can support output formats other than JSON, and that serializing a `Pair` to JSON is an error. Even though the example for conversion shows the converted value as an output, I'm not sure it's clear enough that this isn't just talking about how the workflow author is responsible for never passing `Pair`s to the JSON serialization WDL functions, but also how the end user is responsible for never calling a workflow or task directly in a way that would cause it to output a non-null `Pair` when they have their implementation set to produce JSON output. And it would make sense to caution workflow authors against writing such a workflow at all, somewhere other than in the section on serialization.

Is an implementation required to allow a workflow or task where `Pair` appears in task inputs or outputs to type-check? Is it required to allow a `Pair` to be passed into or out of a task at runtime, when the value will not appear in the final top-level workflow output? Is it required to allow serializing an empty `Array[Pair[Int, Int]]` or a null `Pair[Int, Int]?`?

The implementations I'm familiar with have invented nonstandard ways to serialize pairs, because they want to be able to do things like record and cache the outputs of a task. So they then end up arguably not actually supporting the WDL-specified JSON format, and only supporting an extension of it where pairs are given some representation.

Issue #692, opened 2024-11-25
request: `exists(File | Directory)` STDLIB function


a-frantz:
I want to do something like this:
```wdl
    output {
        Array[String] labels = if exists("labels.txt") then read_lines("labels.txt") else []
    }
```
but am currently writing an ugly workaround to achieve the same logic.

rhpvorderman:
```
command <<<
my_commad

if [ ! -f "labels.txt" ] 
then touch labels.txt  # Labels.txt will be empty.
fi

>>>
output
Array[String] labels = read_lines("labels.txt")
```

Alternatively if the existance of the file is dependent on the inputs you could use those rather than checking for the existence of the file.

patmagee:
I can also see a separate use case for this which the spec (or maybe engines) handles poorly currently. If I run a workflow with a `File` type (note not File?) most engines will not assert that the file actually exists until it goes to try and use that file.

so you could have `exists('...')` used in the workflow body itself.

claymcleod:
To me, it seems like there is at least one, if not two, bona fide use case(s) here where one would ideally be able to avoid the overhead of a task to use the existing standard library functions. This is also relatively simple and, to my eye, seems to be a lower bar for existing that some of the already implemented standard library functions. Given these things, I'm inclined to include it in WDL 1.4. Let me know what folks thoughts are on that.

claymcleod:
It occurs to me that, in 1.2.1 and 1.3, we clarified that `File` and `Directory`s _must_ exist at evaluation time. So this technically possible today by trying to create a `File?` or `Directory?` and checking if the value is `defined()`. That being said, it doesn't feel natural to think that way, so perhaps this just needs to take a `String` argument that points to the path.

patmagee:
I feel like `exists(..)` is just a syntactic wrapper around:

```
exists(testval){
  File? foo = testval
  return defined(foo)
}
```

claymcleod:
I think the conditional-output use case is sufficient for 1.4, but the useful signature is `Boolean exists(String)`, not `exists(File|Directory)`: a `File` or `Directory` value has already passed existence validation. The function should use the normal path-resolution rules for its evaluation context, return `false` only when the path is absent, return `true` for either a file or directory, and preserve errors for inaccessible or unresolvable paths. Would anyone object to narrowing the proposal to those semantics?

Issue #474, opened 2021-08-17
Enable glob for input files


kirkgrubbs1:
I have a program that wants me to input a fairly large list of files. So far I've been doing this but literally listing them out in array format (Ex: Array[File] files = ["a.fna", "b.fna", "c.fna", "d.fna", etc...]). I was thinking this would be much easier if I could use something like the output statement Ex: Array[File] files = glob("*.fna"). 

I would also appreciate any suggestions on how to avoid literally listing out all elements for arrays of files. I'm new, so I'm guessing I'm missing something.

jdidion:
Are you hard-coding the file array in your WDL, or passing it in as an input? I suggest doing the latter, and you can write a script that creates the input JSON for you.

Another option is to use the `Directory` type, but that requires using the `development` version of WDL, which isn't universally supported yet.

kirkgrubbs1:
Writing a script to create the input JSON is the route I'm currently going about, but I just thought it was a bit cumbersome.

I did see the `Directory` type, but haven't gotten to updating to the `development` version yet.

kirkgrubbs1:
I guess my main issue with this is that it seemingly makes the language inconsistent. A function for gathering outputs from the disk would seem to logically apply to inputs as well. Such that you can gather inputs from the disk as well. 

I don't know, maybe its just me. I also understand that it is rare that a program requires gathering different sets of thousands of files each run.

patmagee:
You have to remember that globbing on disk during a task is a very different operation then globbing inputs from an arbitrary file system which may or may not be on disk. WDL was created first and Formost to be a distributed executtion language for running workflowa at scale in cloud environments. While this is also very easily transferable to a local hpc or computer, the language semantics largely reflect the cloud focus of its design.

There are situations where globbing input files would simply not work. For example if files are coming from arbitrary http urls, what would a glob actually mean? Most http endpoints don't know what globbing is, a d unless it's an ftp server it may not actually be discoverable. (Ie DRS apis from ga4gh dobt have a list functionality). If it's coming from a cloud bucket the globing semantics can be different depending on which cloud you are on. If you are running on a proprietary system (ie dnanexus) what does globbing mean when interacting with their file system.

While globbing inputs may seem like it's a good idea, imo there are too many situations where this wouldn't work to really make this a feature.

As @jdidion pointed out, the `Directory` type was introduced in an effort to solve some of these issues. There are certainly challenges with how the directory type is implemented as is, and it may not be stable moving forward.

[...]

Issue #319, opened 2019-06-24
ability to get Directory file as directory + "/filename"


antonkulaga:
I opened the issue in Cromwell as I could not get a file inside the Directory ( https://github.com/broadinstitute/cromwell/issues/5041 ) , @geoffjentry pointed out that according to the spec it is the way it should work. For me this type of behavior is total nonsence: why do I need Directory type if I cannot get its children as a file? That is why I suggest to fix WDL specs

patmagee:
To me a directory is a static bundle of files that have some arbitrary structure. because of the limitations of alot of different environments which do not have the concept of directories (ie cloud) its often necessary then to represent a directory as something like a `tar` ball. that is copied wholly into a task. I do not think its really possible to easily "inspect" the contents of the directory in any sort of effiecient way to retrieve a child `File` object without first copying that directory to a task. Nor in fact do I think that should be the behaviour. Having the directory as a non subsetable object within the WDL spec to me makes sense, the same way that not being able to randomly access bytes of a file in WDL makes sense. Since WDL is mostly control logic these type of "heavy lifting" operatioon should be done mostly in the the context of a task where you are free to do whatever you would like

@antonkulaga what sort of changes are you requesting. Also just as a reminder, issues do not necessarily translate into changes. If you would like to see a change come to the spec the best way is to create a PR first.

rhpvorderman:
@antonkulage, Wouldn't a 
```WDL
Map[String, File] example_map
```
Fit your purposes better?
Or
```WDL
struct mydir {
     File blah
     File another_blah
}
mydir.blah  #access membership.
```

I prefer data structures that explicitly list which files are in there. Structs and Arrays fill this niche already. So I do not see why the directory type should be altered.

DavyCats:
@patmagee Wouldn't you still have to unpack such a tarball in order for a task to be run? If so, you would know it's contents.

I don't really see why `Directory + String` would be any different from `File + String`, they simply yield a path which either exists or doesn't.

patmagee:
@DavyCats you would unpack it, but my guess is the unpacking is done wihtin the task itself. Ie the executor would have command, something like the followinh:
```
cp $original_path $task_path.tar && tar -x $task_path.tar
```

DavyCats:
It seems to me that the question concerns task outputs.
eg.
```wdl
task blah {
  input {
    Directory dir
  }
  command {
    echo something > ~{dir}/someFile
  }
  output {
    File out = dir + "/someFile"
[...]

PR #800, opened 2026-08-26
feat: add Directory listing stdlib functions (reapply of #772)

markjschreiber:
## Summary

This PR reapplies the changes originally introduced in #772, which were merged on 2026-06-30 and subsequently reverted on 2026-07-16 (revert commit `7b6015a2`).

It restores the following standard library functions for WDL 1.4:

| Function | Signature | Description |
|----------|-----------|-------------|
| `list_directory` | `Array[File] list_directory(Directory, [String])` | List files in top level of directory, optionally filtered by glob |
| `list_directory_recursive` | `Array[File] list_directory_recursive(Directory, [String])` | List all files recursively, optionally filtered by glob |
| `list_subdirectories` | `Array[Directory] list_subdirectories(Directory)` | List immediate child directories |

These enable scatter-gather patterns over the contents of a `Directory`, which is currently not possible (see #474, #319).

## Changes
- `SPEC.md`: adds specification and examples for the three functions (+292 lines)
- `CHANGELOG.md`: changelog entry (+2 lines)

Relates to: #772 (original), #474, #319, #692

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ version 1.4.0
 version 1.3.0
 ---------------------------
 
++ Added `list_directory`, `list_directory_recursive`, and `list_subdirectories` standard library functions for listing the contents of a `Directory` as `Array[File]` or `Array[Directory]`, enabling scatter-gather patterns over directory contents.
+
 + Clarified that relative paths in `File` and `Directory` declarations are resolved relative to the WDL document's parent directory outside the `output` section, and relative to the task's execution directory inside the `output` section. Also clarified that optional files evaluate to `None` in both contexts if the path does not exist.
   ([#735](https://github.com/openwdl/wdl/pull/735))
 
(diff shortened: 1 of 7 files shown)


claymcleod:
What if we replaced the three functions with one:

```wdl
Pair[Array[File], Array[Directory]] list(
  Directory,
  [Boolean, [Boolean, [String]]]
)
```

The pair’s `left` value contains files, and its `right` value contains directories. The first optional Boolean controls recursive traversal and defaults to `false`. The second controls whether symlinks are included and defaults to `true`. A symlinked directory is returned but never traversed. The optional string filters files by basename.

Using the general name `list` also gives us a consistent extension point for other values that can be listed. WDL polymorphic function choices may have different argument and return types, as `read_tsv` already demonstrates. We could therefore add other `list` overloads later and define listing semantics appropriate to each input type, provided their argument signatures remain unambiguous.

markjschreiber:
> What if we replaced the three functions with one:
> 
> ```wdl
> Pair[Array[File], Array[Directory]] list(
>   Directory,
>   [Boolean, [Boolean, [String]]]
> )
> ```
> 
> The pair’s `left` value contains files, and its `right` value contains directories. The first optional Boolean controls recursive traversal and defaults to `false`. The second controls whether symlinks are included and defaults to `true`. A symlinked directory is returned but never traversed. The optional string filters files by basename.
> 
> Using the general name `list` also gives us a consistent extension point for other values that can be listed. WDL polymorphic function choices may have different argument and return types, as `read_tsv` already demonstrates. We could therefore add other `list` overloads later and define listing semantics appropriate to each input type, provided their argument signatures remain unambiguous.

That could work

claymcleod:
Alright, let's roll with that then, since that's a more compact representation of the idea. This idea has passed with 2 votes for (Mark Schreiber, Clay McLeod) and 0 votes against.

